PR #13, opened 2015-03-03
Add OCA as author of OCA addons

gurneyalex:
In order to get visibility on https://www.odoo.com/apps the OCA board has
decided to add the OCA as author of all the addons maintained as part of the
association.


legalsylvain:
see #15

PR #15, opened 2015-03-03
Add OCA as author of OCA addons

gurneyalex:
In order to get visibility on https://www.odoo.com/apps the OCA board has
decided to add the OCA as author of all the addons maintained as part of the
association.

### DIFF
--- a/pos_order_pricelist_change/__openerp__.py
+++ b/pos_order_pricelist_change/__openerp__.py
@@ -41,7 +41,7 @@ Copyright, Authors and Licence:
     * Copyright: 2014, GRAP: Groupement Régional Alimentaire de Proximité;
     * Author:
         * Sylvain LE GAL (https://twitter.com/legalsylvain);""",
-    'author': 'GRAP',
+    'author': "GRAP,Odoo Community Association (OCA)",
     'website': 'http://www.grap.coop',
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1:

legalsylvain:
:question: Excuse but I don't see why OCA is the 'author' of all the modules.
Can you send a link with discussion about that ?
Regards.

gurneyalex:
@legalsylvain this was a decision of the board. The goal is to let the OCA gain visibility, and one way to achieve this is through https://www.odoo.com/apps/contributors/modules/

We asked the editor to add a field to let people find OCA addons easily and we got a negative answer, so we decided to add the OCA as author in the addons metadata (of course without removing the original author, nor changing the website field). There will be other PR to update the description field of the addons to include the OCA template.

legalsylvain:
not very clear for me but if we don't have other choice...

gurneyalex:
@legalsylvain the goal is not to alienate our contributors, and I realize we should have communicated before doing the move. I apologize for not doing so.

legalsylvain:
No problem, I understand the problem, and this appears to be a "non-perfect-but-ok solution'
But yes, more communication from OCA Board (main decision) should be great.
Anyway, thanks for your work.
Regards.

PR #20, opened 2015-03-31
[ADD] Add 'pos_product_category' module

sylvainc:


pos_product_category/point_of_sale_view.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="product_template_form_view" model="ir.ui.view">
+            <field name="name">product.template.form.inherit</field>
+            <field name="model">product.template</field>
+            <field name="inherit_id" ref="product.product_template_form_view"/>
+            <field name="priority" eval="100"/>
+            <field name="arch" type="xml">
+                <field name="pos_categ_id" position="replace"/>
```
sebastienbeau: Can you make the field invisible instead of replacing it?
It will avoid problem if an other module base his field position on this field

legalsylvain:
Hi Sylvain,
Thanks for your contribution. 
- I think that the name of the module and the description are not very clear. I propose something like 'pos_remove_pos_category' for the name, and in the description, I think that you should more describe the behaviour of your module during the installation.
  https://github.com/OCA/pos/pull/20/files#diff-aa0e2b0776e1fb62d4d9a705427dcaf7R61
  delete all data in product_template. So you have to mention it very clearly in the module description.

:question: 
- What about if the user uninstall the module ? If you do so, pos_categ_id will be linked again to pos_category model, and so the database will be corrupted. I don't know if it's possible to run a script during the uninstallation of a module.

sylvainc:
Hi,

@legalsylvain : Good catch! I have reworked the tricky install/uninstall parts of the module, it is now much more "safe". I have also renamed the module pos_remove_pos_category.

@sebastienbeau : I have made the field invisible as you have proposed instead of discard it.

Thank you both for the review!

legalsylvain:
Hi, 
thanks for all the changes. Now OK for me. :+1: 
(code review, no test.)

legalsylvain:
@sebastienbeau is this PR ok for you ?

sebastienbeau:
it's ok for me :+1:

sylvainc:
Thanks!

PR #23, opened 2015-04-30
pos_payment_terminal: Add POT file, FR translation and demo data

alexis-via:


brother-bernard:
:+1: Thank you for french users !

legalsylvain:
:+1: Thanks.

PR #17, opened 2015-03-10
[ADD] new module pos_keep_draft_orders to let orders in 'draft' state

legalsylvain:
This module allow users to let a PoS Order in a 'draft' state when closing the session.
Please see full description in OpenERP file.

Remarks are welcome.

Code Coverage of this module : 100%.

Kind regards.

PR #24, opened 2015-05-06
[ADD] module pos_no_bank_statement

andrius-preimantas:


pos_no_bank_statement/model/point_of_sale.py
```diff
@@ -0,0 +1,21 @@
...
+# This file is part of OpenERP. The COPYRIGHT file at the top level of
+# this module contains the full copyright notices and license terms.
+
+from openerp import models, api
+
+
+class POSOrder(models.Model):
+    _inherit = "pos.order"
+
+    @api.v7
```
pedrobaeza: Why not use `@api.cr_uid_id_context`?

andrius-preimantas: Because original method in _point_of_sale_ is defined using old API. I've had some problems with old-new API compatibilities so I try to write with same API which original method is written. But if you think there would be no problems when decorated with this decorator then I'll rewrite it to new api.

Cheers

pedrobaeza: That decorator is used for that cases: you use cr, uid, context, and the third argument is a single id, loading in self the corresponding record, in contrast to `@api.multi`, that uses cr, uid, ids, context arguments to load a list of ids as the recordset.

andrius-preimantas: Right, I'm not very experienced with new API. Could you please show how this method should look like in new API so that it would work despite the original method is written in old API?

Cheers

alexis-via:
@andrius-preimantas There is a typo in the string of the field no_bank_statement on account.journal.

pos_no_bank_statement/model/point_of_sale.py
pedrobaeza:
```
@api.cr_uid_id_context
def add_payment(self, data):
    journal_id = data.get('journal')
    if journal_id:
        journal = self.env['account.journal'].browse(journal_id)
        if journal.no_bank_statement:
            return None
```

andrius-preimantas: @pedrobaeza, thanks but I can not make it work. With your code I get `TypeError: add_payment() takes exactly 2 arguments (6 given)` error. Are you sure Odoo knows how to map inherited object written in v7 to v8?

Cheers
Andrius

andrius-preimantas:
@alexis-via, Thank you, just fixed it!

pos_no_bank_statement/model/point_of_sale.py
pedrobaeza: Try this then:

```
@api.model
def add_payment(self, order_id, data):
    journal_id = data.get('journal')
    if journal_id:
        journal = self.env['account.journal'].browse(journal_id)
        if journal.no_bank_statement:
            return None
```

andrius-preimantas:
@pedrobaeza, thanks, this version works! I have more stuff to learn :-)

Cheers
Andrius

andrius-preimantas:
The module appeared to be a bit problematic with Orders created from backend. So closing this PR until I have better logic for the module.

Cheers

legalsylvain:
Hi. I don't know if you'll  realize again a PR, but in that case, could you consider to split your PR into 2 PR. (because there are two unrelated modules).
Kind regards.

andrius-preimantas:
@legalsylvain , Yes I've pushed another module to this branch without realizing that the branch was included in this PR. That's why I closed this PR.

Now I want one module to be merged to OCA modules which is this https://github.com/OCA/pos/pull/27

Cheers

legalsylvain:
OK. About your second module "pos_no_bank_statement", I'm not sure to understand the use case, but I think that maybe it's not the good way to process. (I guess you wanna to manage unpaid sale without invoices). If yes, maybe you could take a look on my PR : #17 . It's a V7 module that allow possibility to let a pos order in a draft mode. 
Kind regards.

PR #25, opened 2015-05-13
[FIX] correcting some chars

mncoon:


legalsylvain:
Hi @copmn, 
I'm not expert in German, but I don't understand why do you want to remove ummlaut accent in the words ?
Odoo / trunk has ummlaut in de.po translation. 
cf : https://github.com/odoo/odoo/blob/8.0/addons/account/i18n/de.po#L302

CC: @OSguard .

Kind regards.

mncoon:
Hi @legalsylvain, @OSguard 

this is because the display (Bixolon) is not printing ä, ö, ü. Instead it prints a, o, u.
At the moment I don't know why this is so. 

Annyways it is very bad if someone in Germany / Switzerland / Austria want's to use it and it prints wrong characters. 

That's the reason why I changed this characters only in this translations which are printed in the display.

Best regards.

legalsylvain:
Oh I see. In a strict way, it is a Bixolon problem. But OK, this kind of hardware doesn't like exotic chars. 
Thanks for your fix.

:+1:

alexis-via:
@copmn It may be possible to display special chars with the Bixolon LCD ; cf les dev guide, but I never tried.
For the moment, the accents are removed in pos/hw_customer_display/controllers/main.py line 94 via a call to unidecode()

mncoon:
@alexis-via 
Thank you for your addition on this. If I can spend a little time on this, I will do some testings.

alexis-via:
@copmn 
The datasheet of Bixolon BCD-1000/BCD-1100 says:
CHARACTER SET
Alphanumeric Character : 95
Extended Characters : 128 X 16 pages
International Character : 32

In the "Command Manual" version 1.00, it says:
21. Select an International Code set
<ESC> R n
But I didn't find any additionnal information, you I don't know which value I should give to "n". The answer seems to be in this document "Code pages manual": http://www.bixolon.com/upload/download/bcd_code_pages_manual_english_rev_2_00.pdf
Maybe the number is the "code pages" manual is the number you should give to "n" in the command <ESC> R n.

alexis-via:
mmm, it seems the answer is at the last page of the "Code pages" manual, I guess the value of "n" should be the first column of the table on the last page.

mncoon:
@alexis-via It wuld be very interesting to do some tests with it, but at the moment my POS-Hardware is fully packed to a box and I need to set up first.

Few weeks ago I implemented the http://www.glancetron.com/ProCPD8035.html successful to the hw_customer_display. There are only few code-changes and I planed to implement this to the hw_customer_display and make it selectable btw. the Bixolon and the Gnacetron in the pos_customer_display settings.

Hope I have time to do this in the next weeks.

mzclick:
@mncoon I'm right now implementing the Glancetron 8035 to hw_customer_display. The control commands in /hw_customer_display/controllers/main.py are updated according this device, but in any case when any text is sent to the display, I get the following error:

/opt/odoo/odoo/openerp/addons/hw_customer_display/controllers/main.py:94: 

RuntimeWarning: Argument <type 'str'> is not an unicode object. Passing an encoded string will likely have unexpected results.
  lines_ascii.append(unidecode(line))

Could you give me some light here? Is it related to the collate of the DB? I'm from Spain and we use some special chars like "á, é, í, ó, ú"

Thanks in advance.

PR #26, opened 2015-05-17
pos_customer_display: Add button "Display Total to Customer" + port to new API

alexis-via:
The problem with "pos_customer_display" is the following: currently, the only way to know that we have finished to scan the products for the order is when the user clicks on the payment method. So, with the current code of pos_customer_display, we print the TOTAL on the LCD when the user clicks on the payment method. BUT, the customer wants to know the total amount in order to choose the right payment method (for example, the customer could pay by cash or credit card, but he needs to know the total amount to check if he has enough cash). So I added a button "Display Total to Customer", so that we can display the TOTAL on the LCD when we have finished to scan the products.

This development has been inspired by in-field experience and feedback of real-life users :)

### DIFF
--- a/pos_customer_display/__openerp__.py
+++ b/pos_customer_display/__openerp__.py
@@ -48,11 +48,13 @@ project of the Odoo Community Association http://odoo-community.org/.
 You are invited to become a member and/or get involved in the
 Association !
     """,
-    'author': "Aurélien DUMAINE,Odoo Community Association (OCA)",
+    'author': "Aurélien DUMAINE,Akretion,Odoo Community Association (OCA)",
     'license': 'AGPL-3',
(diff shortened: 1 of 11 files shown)


pedrobaeza:
:+1:

legalsylvain:
LGTM. code review. No test.

alexis-via:
I have put this new code in production at the Randol Abbay at the beginning of the afternoon. As the shop is open today, I'll have immediate feedback in case of issues :)

legalsylvain:
Waiting for your feeback.
Any ways, the general rule is to wait for 5 days before merging.

alexis-via:
It worked perfectly well all the afternoon at the Randol Abbey. So let's merge in 5 days if there aren't any negative review.

PR #2, opened 2014-11-15
[ADD] 7.0 'pos_second_header' new module to have an extra header in POS

legalsylvain:
Add 'pos_second_header' in 7.0 serie.

This PR has been previously approved on launchpad:
- @ovnicraft :+1: 2014-07-30
- @pedrobaeza :+1:  2014-09-23

Ref: https://code.launchpad.net/~sylvain-legal/openerp-pos/7.0_pos_second_header/+merge/228820

PR #27, opened 2015-05-19
Pos autoreconcile

andrius-preimantas:


legalsylvain:
Hi. Thanks for your module !
Could you add a test for that module ?

pos_trade_receivable_autoreconcile/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# This file is part of OpenERP. The COPYRIGHT file at the top level of
+# this module contains the full copyright notices and license terms.
+
```
legalsylvain: Could you add a more classical header in each python file.
I know that licence is redundant in many python file but this is the general rules in OCA. And it allows to indicate differents authors if a file is written by many people.
sample : https://github.com/OCA/server-tools/blob/8.0/admin_technical_features/__openerp__.py

legalsylvain:
Hi. i'm not in favour to use SUPERUSER_ID to reconcile entries. 
Closing session means writing entries, so the user has to have the right to write account move and reconcile.
@alexis-via : What do you think about that point ? (as you know both POS and account modules)

Kind regards.

andrius-preimantas:
@legalsylvain , based on standard access rights (see https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/security/ir.model.access.csv) nobody in POS groups has rights to create account.move.reconcile. I could either give this access or make reconciliation as a superuser. Of course there might be setups where even SUPERUSER can not create reconciliation, but I don't know if it's possible.

pos_trade_receivable_autoreconcile/model/point_of_sale.py
```diff
@@ -0,0 +1,53 @@
...
+                                  move_id=None, context=None):
+        to_ret = super(POSOrder, self)._create_account_move_line(
+            cr, uid, ids, session=session, move_id=move_id, context=context)
+
+        account_def = self.pool.get('ir.property').get(
+            cr, uid, 'property_account_receivable', 'res.partner')
+
+        grouped_data = {}
+
+        for order in self.browse(cr, SUPERUSER_ID, ids, context=context):
```
sbidoul: SUPERUSER_ID is not needed here?

legalsylvain:
@andrius-preimantas : Oh I didn't know that.

> SUPERUSER can not create reconciliation:

this case can not happen because there is not access check if uid = 1. (almost in V7, but I guess in V8, there is the same behaviour).
But I'm really not in favor to use SUPERUSER_ID.

> give access to POS group:

Yes, I think it's a better solution than the first one, adding just this information in the module's description. The good solution because with this module, you allow POS user to reconcile entries.

If solution 2 is not OK, a third solution could be to simply add in pos_config a reconcile_user_id (res.users), and use it instead of current uid. (and instead of SUPERUSER_ID).

What do you think ?

pos_trade_receivable_autoreconcile/model/point_of_sale.py
```diff
@@ -0,0 +1,53 @@
...
+                if (line.partner_id.id == key[0] and
+                        line.account_id.id == key[1] and
+                        (line.debit > 0) == key[2]):
+                    grouped_data[key].append(line.id)
+                    break
+        for key, value in grouped_data.iteritems():
+            if not value:
+                continue
+            self.pool.get('account.move.line').reconcile_partial(
+                cr, SUPERUSER_ID, value)
```
sbidoul: why reconcile_partial and not reconcile?

andrius-preimantas:
@legalsylvain, I vote for No. 2. If nobody minds I'll grant POS user reconciliation record creation rights and mention that in README as suggested.

pos_trade_receivable_autoreconcile/model/point_of_sale.py
sbidoul: The POS user group can create move lines but not reconciles. Perhaps it's better to add an ACL on account.move.reconcile to let POS users create them instead of using SUPERUSER_ID. 

I've no clear advice on this topic. Actually I wish Odoo would have a mode where ACL's are bypassed when the code says it is safe to do so, so the write_uid is preserved and we don't need to give too much rights to the user. But that's another story.

sbidoul:
Thanks for this contrib. It's very much needed.

Anyone knows if there is a way to test it on OCA's runbot?

pos_trade_receivable_autoreconcile/model/point_of_sale.py
andrius-preimantas: Because I'm not sure if those entries can always be fully reconciled. After all `reconcile_partial` also fully reconciles entries if it is possible.

legalsylvain:
hi @andrius-preimantas, any news about adding ACL rights ?
Regards.

andrius-preimantas:
@legalsylvain , sorry, a bit overwhelmed with work.

Will do that ASAP.

Cheers

legalsylvain:
@andrius-preimantas : could you fix flake8 issue ?
@sbidoul : I guess it is possible to test it on runbot.

sbidoul:
I did a few tests on runbot and it works fine. I tested with and without customer and tested with negative quantities. All reconciled correctly.

Can you update the README.rst according to the latest template in 
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
And also look after the file headers (https://github.com/OCA/pos/pull/27#discussion_r30697119).

rafaelbn:
I tested in runbot fine. Waiting for update in README.rst and https://github.com/OCA/pos/pull/27#discussion_r30697119 as @sbidoul said. Thanks

Diferences with (yellow) and without this module (blue)

andrius-preimantas:
Changes made according to your comments :-) Thanks!

legalsylvain:
Hi @andrius-preimantas. 
Around this feature, I did a module in V7 here : https://github.com/grap/odoo-addons-misc/blob/7.0/pos_invoicing/__openerp__.py
This module : 
1. reconcile pos entries; (what your module do) ; 
2. forbid payment for paid invoices until session is closed and reconcile invoiced PoS ; 
3. other unrelated feature (allow to invoice unpaid)

What about to include the feature #2 in your module ? (I'm OK to do the PR, and to add test).

Use case #A fixed : 
- Create a PoS order ;
- Pay and invoice the PoS order ; 
- Close the session ;
  --> there are unreconciled payment, because the invoice created is in draft mode ; 

Other Use case fixed : 
- Create a PoS order ;
- Pay and invoice the PoS order ; 
- Open the invoice : it is possible to register a payment ;
- Close the session ;
  -> the real payment is not reconciled ;

andrius-preimantas:
Hi @legalsylvain ,

Actually I did see your module before creating this one, however my use case was a bit different from what your module do.

If we add feature #2 in this module than I believe we need to rename the module, because it implies that only Trade Receivable is reconciled, not Invoices. Maybe yet another module pos_invoice_autoreconcile is more appropriate approach.

Regarding use case #A:
You say that Invoice is in draft mode but actually when you make an Invoice in POS it is automatically validated (hence: https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/point_of_sale.py#L651).

The problem still remains that Payment is only booked when session is closed, so we can not automatically reconcile Invoice when it's issued. Also you would have to forbid paying an Invoice which was created from POS (new flag on account.invoice maybe?)

Cheers
Andrius

legalsylvain:
Thanks for your answer,

> If we add feature #2 in this module than I believe we need to rename the module, because it implies that only Trade Receivable is reconciled, not Invoices. Maybe yet another module pos_invoice_autoreconcile is more appropriate approach.

What about to have an only one module pos_autoreconcile that reconcile "regular trade" AND invoices ? It is what we need. (and what it's expected) ? 
I don't know... what do you think ? 
i could understand to have two module, but I don't see the case when we need only one of both.
CC @rafaelbn, @sbidoul 

> You say that Invoice is in draft mode but actually when you make an Invoice in POS it is automatically validated (hence: https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/point_of_sale.py#L651).

In fact, only by UI, by back-office the invoice is in draft mode. (I justed tested on runbot). (I didn't know that by UI, it was OK, thanks, but the problem should to be fixed for invoice created in back-office : in real life, the customer frequently ask an invoice after the payment, so the 'invoice' button in Front UI is no longer available). But if it works on Front Office, and not in back-office, I'll push a PR directly in Odoo / OCA.

BTW, could you pull -- rebase your PR, the Front UI is currently broken because of old base, on runbot, for your PR.

> The problem still remains that Payment is only booked when session is closed, so we can not automatically reconcile Invoice when it's issued. Also you would have to forbid paying an Invoice which was created from POS (new flag on account.invoice maybe?)

yes : exactly what my module does in V7 : https://github.com/grap/odoo-addons-misc/blob/7.0/pos_invoicing/model/account_invoice.py#L37

andrius-preimantas:
@legalsylvain 

Yeah, one module for invoice/trade receivable reconciliation sounds like a good idea. Would you be willing to contribute to that? I can create pos_autoreconcile module based on this one so you could fork it, make reconciliation on Invoices and create PR so I can merge our code and push to OCA/pos, what do you think?

Cheers
Andrius

legalsylvain:
hi @andrius-preimantas, 
Yes sounds good !
I'll have time to work on that the next week. If you can prepare the PR, it's perfect.

In the meantime, I set a "WIP" label on that PR, if you're OK.

Regards.

andrius-preimantas:
Hi @legalsylvain ,

I renamed module to be pos_autoreconcile. You can fork now and make a PR with your changes.

Thanks,
Andrius

legalsylvain:
thanks !

adrienpeiffer:
@andrius-preimantas, currently, isn't possible to test this one on runbot. An exception is raised when we start the session. Could you check why ?

legalsylvain:
I guess a pull --rebase could solve the problem.

andrius-preimantas:
I rebased this branch to start from the most recent version of 8.0 branch. @legalsylvain be warned that I forced those updates so you might need to update your local branch.

@adrienpeiffer, an error should be gone now.

Cheers
Andrius

adrienpeiffer:
Hi guys !
I don't really understand the problem to do that on invoices.
A loop like this might work?

```
    for key, value in grouped_data.iteritems():
            for line in order.invoice_id.move_id.line_id:
                if (line.partner_id.id == key[0] and
                        line.account_id.id == key[1] and
                        (line.debit > 0) == key[2] and
                        line.state == 'valid'):
                    grouped_data[key].append(line.id)
                    break
```

rafaelbn:
@andrius-preimantas @legalsylvain which is the status of this PR?

cc @jcarlosmontoya @JGarcia-Panach it could be very nice your review here! Thanks

jcarlosmontoya:
This module works fine (y)

jcarlosmontoya:
:+1:

rafaelbn:
@legalsylvain why this PR is WIP? I think is ready to be merged

JGarcia-Panach:
:+1:

rafaelbn:
If you are interested this PR is related with https://github.com/OCA/pos/pull/110

legalsylvain:
Hi, 
Sorry I had no time to work on this extra feature we talked. (flag to forbid payment against pos invoiced).
what about to accept this PR, and next improve the module in a second step ?
regard.

rafaelbn:
Hi @legalsylvain , OK but do you approve this PR for merging?

@jcarlosmontoya @JGarcia-Panach take care about https://github.com/OCA/pos/pull/27#issuecomment-142851648 comment as @legalsylvain said.

JGarcia-Panach:
@rafaelbn, I agree with @legalsylvain, we can merge this module and after we will propose #110 as an improvement.

We have been working with pos_autoreconcile with two clients for two months in production and everything is ok, so is a clear :+1:  for me

legalsylvain:
thanks for your contribution & the reviews.

libreec:
Hi, is this module working for v9 or only for v8?.

andrius-preimantas:
Hi.

Module not tested on v9.

PR #29, opened 2015-06-02
[ADD] module pos_require_customer

JosDeGraeve:
This module allows you to require a customer for each pos order.  In the
pos configuration, you can choose to require the customer for pos
orders.

If a customer is not selected, the pos ui will display an error message.
In the backend the customer field is required when needed.


pos_require_customer/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+
+This module allows you to require a customer for each pos order.  In the
+pos session configuration, you can choose to require the customer for pos
+orders.
+
+If a customer is not selected, the pos ui will display an error message.
+In the backend the customer field is required when needed.
+
+    """,
+    'author': 'Apertoso NV',
```
legalsylvain: Could you add OCA as coauthor ?
Sample : https://github.com/OCA/pos/blob/8.0/hw_customer_display/__openerp__.py#L78

legalsylvain:
Hi. thanks for sharing this module. 
Code review : OK for me. 
Just on point, could you add a constraint _constraint, on pos.order based on session_id, partner_id, to create the same check if the user create pos order by back end GUI ?
 Regards.

legalsylvain:
>  In the backend the customer field is required when needed.

In fact I don't see this feature implemented.

JosDeGraeve:
Hi,

Tnx for the review.  The advertised functionality is present now :)

I actually implemented it on a version running with a client, but merged an older version.  Sorry for the confusion.

I 'll further try to add some tests.

pos_require_customer/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+##############################################################################
+
+{
+    'name': 'Point of Sale Require Customer',
+    'version': '1.0.1',
+    'category': 'Point Of Sale',
+    'sequence': 6,
+    'summary': 'Point of Sale Require Customer',
+    'description': """
+Require customer for pos
```
pedrobaeza, line 28: Please extract this to a README.rst file and use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

pos_require_customer/__init__.py
```diff
@@ -0,0 +1 @@
+from . import pos
```
pedrobaeza: Put encoding and move the models file to a folder `models`

pedrobaeza:
I wasn't able to try on runbot. I got a too much recursion error.

@legalsylvain, do you know about this error?

JosDeGraeve:
Imo, i got i all right now.

pedrobaeza:
@JosDeGraeve, it still continues with the same error starting the POS:

seleccion_001

legalsylvain:
Hi, @JosDeGraeve. The problem @pedrobaeza mentions seems to be out of the scope of your PR. The branch V7 / all seems to be broken in Runbot.
It is urgent to add test on web client readable by Travis / Runbot.

legalsylvain:
@JosDeGraeve : could you rebase or merge the last upstream, please. It could fix @pedrobaeza remarks I guess.

legalsylvain:
:+1: Code review and test. Thanks a lot !

pedrobaeza:
Can you please name it _pos_customer_mandatory_ for avoiding verbs in the module name?

StefanRijnhart:
:+1:. @pedrobaeza can we be lenient about the name? It has been a long review cycle already.

pedrobaeza:
That last change is very easy. I can do it at merge time if the author can't do it, but I prefer to keep the best possible names now that the module is not yet on the repo.

StefanRijnhart:
OK, but then I think that pos_customer_required is more appropriate given the prevalence of the 'required' terminology over 'mandatory' in Odoo.

pedrobaeza:
OK, I agree with the name. Should I proceed with the merge?

legalsylvain:
Ow ! 
I guess I'll have to re-rename my V9 PR ... ;-) (https://github.com/OCA/pos/pull/74)

StefanRijnhart:
@pedrobaeza that is fine by me. Make sure you search+replace all occurrences of pos_require_customer in the module (but I don't really have to tell you that, do I).

pedrobaeza:
Renamed and merged!

PR #33, opened 2015-06-22
[ADD] POS with tax addon

antespi:
Include tax in price on POS.

**Note**: This addons takes care of tax configuration: price_include, showing prices
with tax included even price_include == True or price_include == False

**Motivation**: This helps to shop owner because he can set product prices with taxes
or without taxes, but prices are always shown in POS with taxes

Similar to website_sale_with_tax (https://github.com/OCA/website/pull/47) in website repo


pos_with_tax/README.rst
```diff
@@ -0,0 +1,38 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: AGPLv3 License
+
+Point of sale with tax
+======================
+
+Include tax in price on POS.
+
+This addons takes care of tax configuration: price_include, showing prices
```
pedrobaeza: Put it more at user level: `This addon shows the correct price indepently of the tax configuration (with or without the "Tax included in price" check).

pos_with_tax/README.rst
```diff
@@ -0,0 +1,38 @@
...
+with tax included even price_include == True or price_include == False
+
+This helps to shop owner because he can set product prices with taxes
```
pedrobaeza: s/with taxes or without taxes/with or without taxes

pos_with_tax/__init__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Python source code encoding : https://www.python.org/dev/peps/pep-0263/
+##############################################################################
```
pedrobaeza: Remove header (except coding part)

pos_with_tax/README.rst
```diff
@@ -0,0 +1,38 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: AGPLv3 License
+
+Point of sale with tax
```
pedrobaeza: Product prices in POS with taxes

pos_with_tax/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Point of sale with tax',
```
pedrobaeza: Product prices in POS with taxes

pedrobaeza:
The method you are using overwrites any other modification, so it's incompatible with other modules in the repository like https://github.com/OCA/pos/tree/8.0/pos_pricelist. Can you find another method that gets the price from super and add the tax seeing if it's needed?

antespi:
@pedrobaeza which method are you referring?
- .include()?
- .extend()?

Or method in templates? I've just changed inherit strategy in XML files, in order to be more compatible to other addons

pedrobaeza:
I mean the `instance.point_of_sale.Orderline` one

antespi:
@pedrobaeza we inherit `Orderline` model in the same way:
- pos_with_tax: https://github.com/OCA/pos/pull/33/files#diff-0ab33440268aa7a12c5aedac81d54a61R138
- pos_pricelist: https://github.com/OCA/pos/blob/8.0/pos_pricelist/static/src/js/models.js#L119

Which's the better way to do it, in order to be as more compatible as possible?

pedrobaeza:
But you are not calling super nor using possible parent values. I'm not a JS expert to tell you the exact code I'm afraid.

pos_with_tax/static/src/js/main.js
```diff
@@ -0,0 +1,148 @@
...
+
+    openerp.pos_with_tax = function (instance) {
+        var _t = instance.web._t,
+            _lt = instance.web._lt;
+        var QWeb = instance.web.qweb;
+
+        var round_pr = instance.web.round_precision;
+
+        instance.point_of_sale.ProductListWidget.include({
+            compute_all: function(taxes, price_unit) {
```
legalsylvain, line 41: About @pedrobaeza point :

You overwrite this function. https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/static/src/js/models.js#L860

And I can not see what you changed. Could you explain ? and set some comments to mentions the lines you changed ?

Kind regards.

antespi, line 41: @pedrobaeza, @legalsylvain, we inherit two objects:
- ProductListWidget: where `compute_all` and `get_product_all_prices` methods do not exist. In this case, OrderLine `get_all_prices` can't be reused because product selection is done inside method, so I re-implemented the method (this way go against DRY principle, but i didn't find any better way to do) with `product` as parameter. Maybe I can re-use `compute_all`, i'll check it
- Orderline: where I change `get_display_price` method. Here super call doesn't apply because I want to change returned value

antespi, line 41: @pedrobaeza, @legalsylvain Confirmed, `compute_all` method from `Orderline` model is not generic (but object dependent) because read `this.pos.currency.rounding` and `initialize` needs to be called before calling `compute_all`

antespi:
@pedrobaeza 

> But you are not calling super nor using possible parent values. I'm not a JS expert to tell you the exact code I'm afraid.

Most methods are not inherited but added to object, there is no parent in prototype. The methods overwritten (`get_display_price` for instance) are methods for configuring prices and this addons explicitly change this behavior.

antespi:
@pedrobaeza, @legalsylvain any news? We have this addon in production running well
Thanks for reviewing

rafaelbn:
:+1: Working fine in production environment!

AdilHoumadi:
Hello @antespi,

Thanks for this contribution :clap:. I tried your module and I have some remarks regarding the implementation:

1)  I see no benefit of this two methods `get_product_all_prices` and `compute_all` in ProductListWidget, it's just a copy past of what exists in  `Orderline (models.js standard POS)`.
You can achieve to the same goal like this way : 

``` javascript
instance.point_of_sale.ProductListWidget.include({
    get_all_prices: function (product) {
        var orderline = new openerp.point_of_sale.Orderline({}, {
            pos: this.pos,
            order: null,
            product: product,
            price: product.price,
        });
        return orderline.get_all_prices();
    },
    get_product_price_without_tax: function (product) {
        var res = this.get_all_prices(product);
        return res['priceWithoutTax'];
    },
    get_product_price_with_tax: function (product) {
        var res = this.get_all_prices(product);
        return res['priceWithTax'];
    }
});
```

2) Your implementation for `get_display_price` method shows price with taxes whatever the fiscal position of the customer. You should take care of this point and to do that you should compute price based on fiscal position of the customer which already done in `pos_pricelist`. It's clear that is not trivial as computation :worried:

``` javascript
instance.point_of_sale.Orderline = instance.point_of_sale.Orderline.extend({
    // Show order line prices with taxes included
    get_display_price: function () {
        return this.get_price_with_tax();
    }
});
```

What I suggest is to improve the existing `pos_pricelist`module instead creating another one.
You can add an option to the config of POS to decide to show price with or without taxes.

Thanks.

antespi:
This features are included, with better implementation, in this PR: https://github.com/oca/pos/pull/39

PR #34, opened 2015-06-27
[ADD] new module pos_check_session_state to avoid PoS users use closed sessions;

legalsylvain:
This module add a simple blocking pop up that is displayed if the session the users is working on is now closed. 
This feature avoid bugs if users forgive to close his window when he stops selling.

capture du 2015-06-27 14 56 03


pos_check_session_state/static/src/js/pos_check_session_state.js
```diff
@@ -0,0 +1,68 @@
...
+        template:'ErrorClosedSessionPopupWidget',
+
+        check_session_frequency: 10000,
+        session_name: '',
+
+        init: function(parent, options) {
+            var self = this;
+            this._super(parent,options);
+            this.intervalID = setInterval(function() {
+                var loaded = self.pos.fetch('pos.session', ['name','state'], [['id', '=', self.pos.get('pos_session').id]])
```
legalsylvain: This part raise an
pos - fetch error
 uggly recurring window if server is not available. (that is possible in PoS because it must works offline). 
Any idea to manage that case ? I tried with a try / catch without any success.

Kind regards.

legalsylvain: TODO : Try : event.preventDefault();

legalsylvain:
I close this PR as there is no activity on it.

PR #38, opened 2015-07-09
[ADD] Show taxes details in POS order addon

antespi:
This module track list of taxes in a POS order. Then shows one line per tax with its total amount in POS order PDF report.

**Notice** : As a POS order hasn't a mandatory partner, this addon computes taxes regardless partner's fiscal position (if any). User can create an invoice from the POS order if he wants to take care about partner's fiscal position.

**Notice** : Taxes are computed when POS Order is paid, because after this action product will not change.


guewen:
:+1:

pos_order_tax_detail/models/pos_order.py
```diff
@@ -0,0 +1,80 @@
...
+        return True
+
+    @api.model
+    def _install_tax_detail(self):
+        """Create tax details to pos.order's already paid, done or invoiced.
+        """
+        # Find orders with state : paid, done or invoiced
+        orders = self.search([('state', 'in', ('paid', 'done', 'invoiced'))])
+        # Compute tax detail
+        orders.compute_tax_detail()
```
legalsylvain: This will not work if an Order was realized with a product with VAT 1, and now the product has a VAT 2 defined.
Sample : VAT changes in France from 19.6% to 20% (in 2014).
Notice : I have no solution. Maybe you could : 
- mention the possible problem in the readme file;
- compute_tax_detail function could return False if pos_order.amount_tax != sum(pos_order_tax.amount).
- action_paid should fail if compute_tax_detail return False. (so it will be possible for user to fix the problem).
  (This last case is if Taxes change when the session is opened.)

antespi: @legalsylvain, thanks for your warning

Taxes are copied (name, base, amount) when order is paid, so taxes are the ones were configured in product when order was paid. Taxes can be changed in product, but this will not affect to order.

Original Odoo code does the same, calculates amount_tax and amount first in JS code (POS) and then recalculates taxes again when creating account_move when POS session is closing. So, you can check this in runbot and see that payments are incoherent with account moves created when POS session is closed and validated. 

It's not expected that user change taxes of a product when a POS session is in process, but you're right, I've added this notice in README. I've also implement an amount check to advice user when taxes have been change to one or more products during POS session.

antespi:
These features are included, with better implementation, in this PR: https://github.com/oca/pos/pull/39

PR #39, opened 2015-07-12
[8.0] Improve pos_pricelist

AdilHoumadi:
**[IMP]**: improve JS indentation

**[IMP]**: remove _super method from Backbone.Model. Never manipulate the native Backbone.Model entity like this way because it is considered as bad practice.

**[ADD]**: Add support for price with taxes, new option is introduced in the POS config to let the user show price with taxes in product widget.
The UI is updated when we change the customer in order to adapt the prices
The computation take in account the pricelist and the fiscal position of the customer
The idea is based on what is done in this PR : #33 by @antespi  

**[IMP]**: simplify JS for price computation (get_all_prices)

**[ADD]**: When we mouseover the price tag, a tooltip is shown to indicate the computation depending on the quantity like this output :
- 1x -> 100 €
- 3x -> 70 €
- 5x -> 50 €

ui-prices

**[IMP]** : update readme to indicate the new improvements


antespi:
@ah-taktik thanks for this improvement. I've tried it in runbot and I've got this client error, with debug actived:

```
Odoo Client Error
Uncaught RangeError: Maximum call stack size exceeded
http://3110750-39-7537c2.runbot2.odoo-community.org/pos_pricelist/static/src/js/models.js:33
```

I've fixed with this ugly workaround:

``` javascript
    module.PosModelParent = module.PosModel;
    module.PosModel = module.PosModel.extend({
        /**
         * @param session
         * @param attributes
         */
        initialize: function (session, attributes) {
            // this.constructor.__super__.initialize.apply(this, arguments);
            module.PosModelParent.prototype.initialize.apply(this, arguments);
            this.pricelist_engine = new module.PricelistEngine({
                'pos': this,
                'db': this.db,
                'pos_widget': this.pos_widget
            });
            arrange_elements(this);
        },
```

In other hand, if you set pos.config.display_price_with_taxes = true, then price is showing without taxes in order line. I suggest you to extend `get_display_price` to show prices properly:

``` javascript
        get_display_price: function(){
            if (this.pos.config.display_price_with_taxes){
                return this.get_price_with_tax();
            }
            return this.constructor.__super__.get_display_price.apply(this, arguments);
        },
```

Finally, I've propused another PR (https://github.com/OCA/pos/pull/38, pos_order_tax_detail), in order to track taxes associated to a `pos.order`. This is important in Spain, for example, because when POS order is printed we need a detailed tax table with total base and amounts for each tax, like invoices. This addon doesn't care fiscal position and calculates taxes in server side, just when POS order goes to paid state. But there's an issue (https://github.com/OCA/pos/pull/38#discussion_r34341471) reported by @legalsylvain that will be fixed if JS code track taxes detailed totals and sync them to server like any other pos.order attribute. This will be the best implementation. Could you integrate this too in this improvement?

AdilHoumadi:
Hello @antespi, Thanks for your feedback :+1:  
I adapt the code as you mention, it works better now !

Thank you.

pos_pricelist/static/src/js/models.js
```diff
@@ -116,16 +130,23 @@ function pos_pricelist_models(instance, module) {
...
+    var OrderlineParent = module.Orderline;
     module.Orderline = module.Orderline.extend({
+        /**
+         * @param attr
+         * @param options
+         */
         initialize: function (attr, options) {
-            this._super('initialize', attr, options);
+            OrderlineParent.prototype.initialize.apply(this, arguments);
             this.manuel_price = false;
```
antespi: Is this a typo-error? Manuel is my  son's name :smile:  Should it be manu**a**l_price?

antespi:
@ah-taktik , thanks for your changes. Are you planning to implement taxes tracking? You should declare another Colletion in Order model, like `TaxlineColletion`, take care about base in `module.Orderline.compute_all` method and update collection for example in `module.Order.getTaxDetails` method with info returned by `line.get_tax_details()`

pos_pricelist/static/src/js/models.js
AdilHoumadi: Yeah, exactly :laughing:

Thanks.

AdilHoumadi:
Hello @antespi,
I am not sure if I understand your question, but each line has his own taxes which means that we don't need to track this changes, if you want to have taxes per line it's possible to override or create a new method to have this kind of output :

---

1x POS Product 1 : 121.00 €
--> VAT 21% - Services 21.00 €

1x POS Product 2 : 112.00 €
--> VAT 12% - Services 12.00 €

1x POS Product 3 : 127.00 €
--> VAT 21% - Services 21.00 €
--> VAT 6% - Services 6.00 €

Total: 360.00 €

---

If it's your request,  Yeah we can propose a new module to override this method and customize the ticket.

Thanks.

antespi:
Hi @ah-taktik, tax tracking in JS (POS offline client) side is ok. We don't need tax detail per line. But in Python (Odoo model) side, pos.order hasn't tax detail, only total tax amount, and when user print a ticket from backend then tax details doesn't appeas. It is quite different compared with ticket printed by POS client.

Some customers want to print tickets in PDF once per week to send them to theirs account department or subcontract account expert. In backend, user can select tickets for a week and print them together in one PDFs but account expert can't figure out what amount come from each tax.

Is this a common need? or this only happends in Spain?

pablocm-aserti:
Hi @antespi 
That lack of tax tracking is not a problen just for us in Spain, but for everyone.
Currently, since the resulting taxes after computing the fiscal position mappings are not stored in the pos.order.line the same way it is done in sale.order.line, it just uses the taxes found in the product itself, as seen [here](https://github.com/odoo/odoo/blob/aa3ce8b594eb827feefe18609daa3abdfd3a8720/addons/point_of_sale/point_of_sale.py#L1324).
It is not just a problem if you want to re-print the ticket. Although the module adds support for fiscal position mappings in the frontend, it does not in the back end and the resulting pos.order amounts are wrong.
Example (VAT 21% -> VAT21% included in the fiscal position):
captura de pantalla de 2015-07-21 12 06 26
captura de pantalla de 2015-07-21 12 52 17

I'll start working now in trying to add the taxes to the pos.order.line
If you are already on it, just let me know

antespi:
@pablocm-aserti We're not working on it right now. I've just overviewed a way to implement it. I think we have to do the same as Odoo standard do for order lines, reusing @ah-taktik code for calculating fiscal position and prices to apply. 

So, IMHO, there are two collection of taxes:
1. Taxes in Order line: A list of taxes applied only to the order line
   - You should declare another Colletion in `module.Orderline` model, for instance `TaxlineColletion`. An save them in `module.Orderline.get_all_prices` method.
2. Taxes in Order: A list of taxes with total base and amount for each one
   - You should declare another Colletion in `module.Order` model, for instance `TaxColletion`. Take care about bases in `module.Orderline.compute_all` method and update collection for example in `module.Order.getTaxDetails` method with info returned by `line.get_tax_details()`

For example, this Order:
1. Product 1: qty=10, price=50, taxes=20%, base=500, tax_amount=100, total=600
2. Product 2: qty=1, price=20, taxes=10%, base=20, tax_amount=2, total=22

Now, we will care about customer with a fiscal position that add a 5% discount tax to all product, so taxes per Orderline are:
1. Product 1:
   - base=500
   - tax20%: percent=0.20, amount=100
   - tax5%: percent=-0.05, amount=-25
   - total this line=575
2. Product 2:
   - base=20
   - tax10%: percent=0.10, amount=2
   - tax5%: percent=-0.05, amount=-1
   - total this line=21

An now, at Order level, we have these taxes:
- tax20%: base=500, percent=0.20, amount=100
- tax10%: base=20, percent=0.10, amount=2
- tax5%: base=520, percent=-0.05, amount=-26

If you add these two Collections, then when JS object is serialized and sent to Odoo server it will include tax information per `pos.order.line` and `pos.order`. So, you should extend those models in Odoo to save taxes. When user print `pos.order` there's no need to recalculate anything because it is already saved in `pos.order`, Odoo just print the tax info (per line, per tax, ...)

pablocm-aserti:
@antespi I've come out with a solution for the backend tax incongruences when they were mapped on the frontend that is very similar to what you propose. Comments are welcome here ah-taktik/pos#2 Thanks!

antespi:
@ah-taktik, @pablocm-aserti this improvement rocks!!

For printing ticket from backend, this could be an example:
image

Taxes are printed at the end of ticket, aggregating each type of tax. This can be calculated in python-side from each order line tax info.

antespi:
@ah-taktik , @pablocm-aserti I'm working on it. I'll do a pull request to your branch. I've done before, it's just copy-paste for me. Wait me some time

antespi:
@ah-taktik , @pablocm-aserti, here https://github.com/ah-taktik/pos/pull/3 you have my part.
Please, review and merge if validated.

AdilHoumadi:
@antespi, @pablocm-aserti, Thank you for you contribution, it's great !

antespi:
Sorry @ah-taktik, but i forgot to add security ACL rules to `__openerp__.py` file
Here you have the PR: https://github.com/ah-taktik/pos/pull/5

This will fix runbot error

antespi:
:+1:
I think this is a great sample of community collaboration :smile:

legalsylvain:
Hi @ah-taktik. Just a quick overview. Is it possible to move in a separate module taxes management ? It seem that your module include some taxes detail that could be better in (#38) (Or did I miss something ?)
Thanks for your clarification.

AdilHoumadi:
Hi @legalsylvain, 
I think this PR replace this two PR (https://github.com/OCA/pos/pull/33, https://github.com/OCA/pos/pull/38). For the 1st one you can check this comment : 
https://github.com/OCA/pos/pull/33#issuecomment-120605777 and for the 2nd check this comment : https://github.com/OCA/pos/pull/39#issuecomment-124488714

@antespi, I'd like you feedback here.
@legalsylvain, Anyway if you find that it's better to split this feature, I am ok with it!

Thanks in advance.

antespi:
@legalsylvain, those PRs (https://github.com/OCA/pos/pull/33, https://github.com/OCA/pos/pull/38) are included in this PR and with a better implementation, because this one take care about pricelist and fiscal position when a customer is selected. This one track properly taxes in pos.order.line and this haven't issues (https://github.com/OCA/pos/pull/38#discussion_r34341471) when a product change its taxes. 

I think that "few addons" criteria is better when possible, and features implemented in this addon are related. So, I prefer to have only one addon: pos_pricelist. Maybe now this name doesn't represent properly features implemented.

IMHO, this features should be implemented in ofiicial point_of_sale addon. They aren't optional in any case. All our customers want pricelist and full tax management in POS orders as they have in sale orders and account invoices. In Spain, a POS order (a ticket) is as legal as an invoice, and we have to take care about fiscal position and taxes detail.

rafaelbn:
:+1:

pedrobaeza:
:+1:

Merging...

AdilHoumadi:
@pedrobaeza, Thank you.

PR #44, opened 2015-08-17
[FIX] init calculation of taxes

JavierIniesta:
This PR fixes the init calculation of taxes when you install this addon.


JavierIniesta:
@ah-taktik If there was orders before installing this addon, the method _install_tax_detail was launched before adding the pos.order.tax. With this change, it will be executed after.

JavierIniesta:
product_taxes_rel is related with product templates not product products. So I added two joins to the SQL statement.

pedrobaeza:
Fields names doesn't help too much on discovering template/product duality...

:+1:

antespi:
:+1: Cool, tested on local environtment

rafaelbn:
Review on runbot, working in production enviroment already
:+1: 
@antespi please check it

PR #45, opened 2015-08-20
[ADD] Pos Gift Ticket Module

ismaelcj:
We propose this module that allows to print a gift ticket, which is the same ticket than normal but without prices and taxes.

To do this, we have had to modify the normal behavior of the _Print Ticket_ button. Both button now refresh the ticket in order to change the view.

Because of refresh the ticket, it changes its date each time.  So we modify the ticket view and change the date, now both take the order's `creationDate`.

### DIFF
--- a/pos_gift_ticket/__openerp__.py
+++ b/pos_gift_ticket/__openerp__.py
@@ -24,7 +24,7 @@
 {
     "name": "Gift Ticket",
     "version": "0.1",
-    "author": "FactorLibre",
+    "author": "FactorLibre, Odoo Community Association (OCA)",
     "website": "http://www.factorlibre.com",
     "license": "AGPL-3",
     "category": "Point Of Sale",
--- a/pos_gift_ticket/__openerp__.py
+++ b/pos_gift_ticket/__openerp__.py
@@ -1,0 +1,39 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright :
+#        (c) 2015 FactorLibre S.L. (http://www.factorlibre.com)
+#                 Ismael Calvo <ismael.calvo@factorlibre.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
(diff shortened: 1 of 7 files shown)


rafaelbn:
Hi @ismaelcj , I have tested module. I don't see any utility on this module. Why is it useful? Which is it general purpose?

rafaelbn:
Hi Ismael, sorry I misunderstood functionality.
Tested in runbot! :+1:

pos_gift_ticket/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Gift Ticket",
+    "version": "0.1",
+    "author": "FactorLibre",
```
legalsylvain: Could you add OCA ?
Sample : https://github.com/OCA/pos/blob/8.0/pos_pricelist/__openerp__.py#L24
(It's a convention, to make easier search).

pos_gift_ticket/static/src/js/main.js
```diff
@@ -0,0 +1,48 @@
...
+        },
+        print: function() {
+            this.refresh()
+            this._super()
+        },
+        print_gift: function() {
+            this.refresh_gift_ticket()
+            this.pos.get('selectedOrder')._printed = true;
+            setTimeout(function() {
+                    window.print();
```
legalsylvain, line 42: This will not work with a PosBox setting ? Or did I missed something ?

ismaelcj:
Hi @legalsylvain. This feature adds a different ticket, which will coexist with the normal. Does the normal ticket has a PosBox setting?

legalsylvain:
Hi, 
Thanks for OCA reference;

about printing, AFAIK, Odoo PoS allow printing by two ways: 
- 'regular way' : by a window.print():
  https://github.com/OCA/OCB/blob/8.0/addons/point_of_sale/static/src/js/screens.js#L950
  -> what it is printed is your page. -> OK for your module.
- 'posbox way': via the proxy system (cf, optional setting in your pos.config):
  https://github.com/OCA/OCB/blob/8.0/addons/point_of_sale/static/src/js/screens.js#L1341
  -> what is printed is a json generated here :  
  https://github.com/OCA/OCB/blob/8.0/addons/point_of_sale/static/src/js/models.js#L1212
  and interprated here : 
  https://github.com/OCA/OCB/blob/8.0/addons/hw_escpos/controllers/main.py#L210

yes, all the PoS printing settings are duplicated in Odoo ... What else ? ;-)

I guess you don't use the posbox way, because I suppose your solution will not work using a proxy.

Can you take a look on that point ? For me, it's not mandatory to change your code to make it compatible with a posbox, but could you write a clear text in the "Limits" section of the description of the module to mention that your module will not work with the 'print by proxy' setting enabled ? 

Kind regards.

pedrobaeza:
The section is know as "Known issues / Roadmap"

ismaelcj:
Ok, now I get it. I can't test it with a PosBox, so I added a _Known issues / Roadmap_ section.

legalsylvain:
:+1: . Thanks.

pedrobaeza:
Shouldn't this be renamed to _pos_gift_receipt_?

PR #56, opened 2015-09-29
[IMP] Show unit price when qty is not 1

antespi:
When qty is not 1 (for example, 2, 3 or 0.56) POS ticket template (PosTicket) is not indicating unit price (but XmlReceipt template, used when proxy configured, implement this feature).
This is specially useful in this addon because unit price would not be product list_price because price list are applied even taking care customer.

So, with this improvement, if qty is equal to 1, unit price is not shown (as now occurs) but if qty is not equal to 1, then unit price is shown.

Please, @ah-taktik and @pablocm-aserti review these changes and do your comments, thanks.

### DIFF
--- a/pos_pricelist/static/src/js/widgets.js
+++ b/pos_pricelist/static/src/js/widgets.js
@@ -17,6 +17,8 @@
  ******************************************************************************/
 function pos_pricelist_widgets(instance, module) {
 
+    var round_di = instance.web.round_decimals;
+
     module.OrderWidget = module.OrderWidget.extend({
         set_value: function (val) {
             this._super(val);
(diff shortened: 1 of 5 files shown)


pos_pricelist/static/src/js/widgets.js
```diff
@@ -63,5 +65,24 @@ function pos_pricelist_widgets(instance, module) {
             this.pos.pricelist_engine.update_products_ui(customer);
         }
     });
+
+    module.PosBaseWidget.include({
+        format_pr: function(amount, precision) {
+            console.log('format_pr: amount = ' + amount)
```
legalsylvain: please remove console.log.

pos_pricelist/static/src/js/widgets.js
```diff
@@ -63,5 +65,24 @@ function pos_pricelist_widgets(instance, module) {
...
+            // Do not call _super because no addon or XML is using this method
```
legalsylvain, line 71: not surto understand this comment. is format_pr a function defined in odoo ?

pos_pricelist/static/src/js/widgets.js
antespi: Sorry, in a moment. I'm waiting to finish runbot to see if any other fix is required.

pos_pricelist/static/src/js/widgets.js
antespi, line 71: Yes it is defined here: https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/static/src/js/widget_base.js#L47 but it not used by anyone. Looks like a deprecated code. For this improvement is cool because I need to show price with currency configuration (decimals) but without currency symbol.

JavierIniesta:
Tested in runbot :+1:

janetrp:
really useful :+1:

rafaelbn:
Tested in runbot :+1:

rafaelbn:
@legalsylvain could make last review? I think it could be merged and change is not critical.
cc @pedrobaeza

antespi:
Version increment, squash and rebase done.

pedrobaeza:
:+1:

PR #65, opened 2015-11-30
[ADD] Add pos_analytic_config module.

adrienpeiffer:
Use analytic account defined on POS configuration for invoices and accounting entries generated form POS orders.


pos_analytic_config/README.rst
```diff
@@ -0,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+===================
+POS Analytic Config
+===================
+
+This module was written to allow to use analytic account defined on POS configuration
```
pedrobaeza: Use the simpler form "This module allows to use"

pos_analytic_config/README.rst
```diff
@@ -0,0 +1,50 @@
...
+for invoices and accounting entries generated form POS orders.
```
pedrobaeza: s/form/from

pos_analytic_config/README.rst~
```diff
@@ -0,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
legalsylvain: could you remove this file ? (with ~)

legalsylvain:
:+1: except minor remark. thanks for this module. (and test file). 

Just a question, how does analytic works by default for PoS. (I don't use analytic). Which account is used ? products setting / other / nothing ?

regards.

pedrobaeza:
The name can be only _pos_analytic_, because there's no other module that handles analytic in POS, and we are putting all the transversal modules that handles analytic stuff to OCA/account-analytic, like the one for MRP that I'm doing or the one for product from Alexis.

adrienpeiffer:
Thanks !
@legalsylvain without development, there is no analytic account.
https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/point_of_sale.py#L1084

pos_analytic_config/README.rst
```diff
@@ -0,0 +1,50 @@
...
+------------
+
+* Adrien Peiffer <adrien.peiffer@acsone.eu>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
```
elicoidal: http (x2)

pos_analytic_config/__openerp__.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
elicoidal, line 1: Please use simple headers following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

pos_analytic_config/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+##############################################################################
+{
+    'name': "POS Analytic Config",
+
+    'summary': """Use analytic account defined on
+                  POS configuration for POS orders""",
+    'author': 'ACSONE SA/NV,'
+              'Odoo Community Association (OCA)',
+    'website': "http://acsone.eu",
+    'category': 'Point Of Sale',
```
elicoidal: add 'accounting'?

legalsylvain:
@pedrobaeza : Not sure to be agree with your PoV. 
Another possible implementation could be to reuse product analytic config. So a pos_analytic_product is possible. (based on account_analytic_default) Don't you think ?
Set all pos sales in one single analytic account is a choice.

Regards.

pedrobaeza:
Yeah, and that module can also be hosted on account-analytic.

legalsylvain:
And yes, if MRP and product are done in another repository, this PR should be done against

Cedric-Pigeon:
:+1: (Code review)

pos_analytic_config/__openerp__.py
adrienpeiffer, line 1: Done. Thanks !

adrienpeiffer:
The discussion will continue on account-analytic.Thanks to all !

PR #70, opened 2015-12-15
[ADD] Add pos_payment_entries_globalization module

adrienpeiffer:
This module allows to globalize payment entries created by the POS.

In some cases, all banking transactions are received in a single bank statement line.
With this module, it's possible to reconcile this one line with the payment globalization line.


libreec:
Hi, this module sounds interesting, are you still using it?, is it available for v9 and v10?...

legalsylvain:
Hi @adrienpeiffer. Interesting module indeed. Thanks for sharing ! Just one question.
why did you defined the field account pos_payment_globalization_account and not just used pos_payment_globalization_journal.default_debit_account_id / default_credit_account_id ?

Otherwise it works like a charm. :+1: 

@libreec : Did you tested this module ? If yes do you put a :+1:  ? 

thanks.

rafaelbn (approved):
Thanks!

PR #74, opened 2016-01-20
[9.0] [MIG+IMP] pos_customer_required

legalsylvain:
Hi, 
- ported the V8.0 PR #29 (pos_require_customer) from @JosDeGraeve in V9 API;
- added some features;

With this module,
- you can prevent user to start selling (front / back office) if customer is not defined;
- OR you can prevent user to make payment (front + back end) if customer is not defined;

Settings are available on each pos config.
fromage_blanc

complete description on readme.rst file.

Thanks for your review.

Kind regards.


pedrobaeza:
Can you please name it _pos_customer_mandatory_ for avoiding verbs in the modules name?

pos_require_customer/models/pos_config.py
```diff
@@ -0,0 +1,32 @@
...
+        ('order', 'Required before starting the order'),
+    ]
+
+    require_customer = fields.Selection(
+        selection=_REQUIRE_CUSTOMER_KEYS,
+        string='Require customer',
+        help="Require customer for orders in this point of sale:\n"
+        "* 'no': customer is optional;"
+        "* 'payment': customer is required before paying;"
+        "* 'order': customer is required before starting the order;")
```
JosDeGraeve: It might be better to have the "user friendly" name of a selection item in the help text, like 'Optional' instead of 'no' ?

JosDeGraeve:
I never got around to writing js test to verify the client side functionality.  That's why the module was still in PR.

Have you planned to include js tests ?

legalsylvain:
@JosDeGraeve : Well, I'd like to include test, but I didnt learnt how to write such tests for the time being. That's a lot of time I have to work on this topic. I don't think it's was a blocking point for your PR. (2 :+1:  was missing).
But yes it's an important point on the "OCA / web" and "OCA / pos" roadmap.

@pedrobaeza, @JosDeGraeve : Thanks for your review. I change the description, the module name, according your remarks.

legalsylvain:
Hi @pedrobaeza, @JosDeGraeve,
- changed the name of the module, according the name change of the 8.0 serie;
- fixed the test;
- added a migration script from V8.0 serie;

thanks for your review.

CC  @StefanRijnhart, as you reviewed the V8.0 serie...

@gurneyalex : I used openupgradelib to make migration. Do I have to do something to make runbot works, or do you have to add a lib manually runbot ?

kind regards.

pedrobaeza:
The module name should be _pos_customer_required_

gurneyalex:
@legalsylvain I just installed the mssing dependency and forced rebuild. Let's see how it goes

legalsylvain:
@pedrobaeza : I renamed the module as you proposed.

@gurneyalex : Build is now ok. Thanks !

Regards.

pos_customer_required/migrations/9.0.2.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+def migrate_require_customer(cr):
+    for old, new in BOOLEAN_TO_SELECTION.iteritems():
+        cr.execute("""
+            UPDATE pos_config
+            SET require_customer='%s'
+            WHERE %s = '%s'
+        """ % (new, openupgrade.get_legacy_name('require_customer'), old))
+
+
+def migrate(cr, installed_version):
```
pedrobaeza: Decorate with `@openupgrade.migrate()`

pos_customer_required/README.rst
```diff
@@ -0,0 +1,88 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+PoS - Customer Mandatory
```
pedrobaeza: Line above

pos_customer_required/README.rst
```diff
@@ -0,0 +1,88 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
pedrobaeza: Target link

pos_customer_required/README.rst
```diff
@@ -0,0 +1,88 @@
...
+* select the point of sales you want configure
+* search for the "Require Customer" and choose between the following values:
+    * 'Optional'; (this module has no effect on this PoS config)
+    * 'Required before paying';
+    * 'Required before starting the order';
+
+Usage
+=====
+
+For further information, please visit:
```
pedrobaeza: Remove this and put try me on runbot button

pos_customer_required/demo/pos_config.yml
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Please use header like in https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

pos_customer_required/migrations/9.0.2.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+BOOLEAN_TO_SELECTION = {
+    'false': 'no',
+    'true': 'order',
+}
+
+
+def migrate_require_customer(cr):
+    for old, new in BOOLEAN_TO_SELECTION.iteritems():
```
pedrobaeza: You can use openupgradelib `map_values` method: https://github.com/OCA/openupgradelib/blob/4bb94fec0e3571cbc530e09b9c5b7903ece35f00/openupgradelib/openupgrade.py#L801

pos_customer_required/migrations/9.0.2.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+column_renames = {
+    'pos_config': [
+        ('require_customer', None),
+    ]
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    if not version:
```
pedrobaeza: This is not needed with the decorator

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
   - printf '[options]\n\nrunning_env = dev' > ${HOME}/.openerp_serverrc
-  - pip install unidecode pyserial pycountry
+  - pip install unidecode pyserial pycountry openupgradelib
```
pedrobaeza, line 36: Migration scripts are not called in Travis, so no need to add openupgradelib.

pos_customer_required/models/pos_config.py
```diff
@@ -0,0 +1,26 @@
...
+        ('no', 'Optional'),
+        ('payment', 'Required before paying'),
+        ('order', 'Required before starting the order'),
+    ]
+
+    require_customer = fields.Selection(
+        selection=_REQUIRE_CUSTOMER_KEYS,
+        string='Require Customer',
+        help="Require customer for orders in this point of sale:\n"
+        "* 'Optional' (customer is optional);"
```
pedrobaeza: You have missed `\n` at the end of these lines.

pos_customer_required/pos_customer_required.pot
```diff
@@ -0,0 +1,62 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove the pot template

pedrobaeza:
In the icon, instead of "Agrolait", I would just put "Customer".

legalsylvain:
@pedrobaeza, @JosDeGraeve : I totally forgot this PR and your comments. 
Now it is good. 

thanks.

pedrobaeza:
Can you check Travis?

.travis.yml
legalsylvain: Travis failed because I remove openupgradelib ref. reverted.

legalsylvain:
@pedrobaeza. Done. It's now green.

StefanRijnhart:
:+1: thanks

pos_customer_required/README.rst
```diff
@@ -0,0 +1,92 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/184/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/pos/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/{project_repo}/issues/new?body=module:%20pos_customer_required%0Aversion:%209.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Put here repo, or better switch to the new text

pos_customer_required/__openerp__.py
```diff
@@ -0,0 +1,30 @@
...
+        'static/src/xml/templates.xml',
+        'views/pos_config_view.xml',
+        'views/pos_order_view.xml',
+    ],
+    'demo': [
+        'demo/pos_config.yml',
+    ],
+    'installable': True,
+    'external_dependencies': {
+        'python': ['openupgradelib'],
```
pedrobaeza: This is why you need openupgradelib in Travis config. I don't usually put this dependency as it's only needed when updating module, not when installing. I think it's better to remove it.

pos_customer_required/__openerp__.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2004-Today Apertoso NV (<http://www.apertoso.be>)
+# Copyright (C) 2016-Today: La Louve (<http://www.lalouve.net/>)
+# @author: Jos DE GRAEVE (<Jos.DeGraeve@apertoso.be>)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html
+
+{
+    'name': 'Point of Sale Require Customer',
+    'version': '9.0.2.0.0',
```
pedrobaeza: Version should be 9.0.1.0.0 (it's different from 8.0.1.0.0, so no problem with that).

pos_customer_required/__openerp__.py
pedrobaeza: And more as the migration scripts will only be called in a migration from 8.0 to 9.0 via OpenUpgrade.

pos_customer_required/migrations/9.0.2.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: The folder name is incorrect

pos_customer_required/__openerp__.py
legalsylvain: Do you propose to try except the import ?

pedrobaeza: No, you just need to remove this dependency. If someone needs to upgrade, then he/she will face the error, but only on that case.

pedrobaeza:
There are still comments unattended

pedrobaeza:
Please don't name GH aliases on commits, or we will be pinged each time the branch is forked, pushed and so on.

legalsylvain:
Ow. OK. I didn't know. without @ it's OK ?

pedrobaeza:
Yeah, that can serve

legalsylvain:
Commit ammended. regards.

pedrobaeza (approved):

.travis.yml
pedrobaeza: This is finally not needed, but it doesn't hurt also.

PR #79, opened 2016-02-15
[FIX] Round unit price

StefanRijnhart:
Fixes a rounding issue with prices from the pricelist. Reproduce by applying a 10% discount pricelist on an article of 0,75 in a database with product price rounding set to two decimals. Sell in the POS at quantity 4 and you'll find a unit price of 0,68 but a total of 2,70 (4x 0,675). Customer pays 2,70 but the POS order in the backend keeps an open balance of 2 ct.

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -148,7 +148,7 @@ function pos_pricelist_models(instance, module) {
                     db, product, partner, qty
                 );
                 if (price !== false) {
-                    this.price = price;
+                    this.price = round_di(parseFloat(price) || 0, this.pos.dp['Product Price']);
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
@AdilHoumadi do you remember if it is a bad idea to just call set_unit_price() in initialize() as well? It would call the trigger though.

StefanRijnhart:
No response from @AdilHoumadi, so I'm leaving it like this.

AdilHoumadi:
@StefanRijnhart, Sorry for the delay, I don't remember exactly. Your improvement LGTM.
It's better to pass via the parent instead of triggering the change like this way. Any way the parent will do the same.

Thanks !

StefanRijnhart:
OK, thanks for the response!

antespi:
:+1: Thanks

pedrobaeza:
:+1:

woepwoep:
Stefan when i set decimal_precision to .001 it works in sale_order but not in pos_pricelist, can you confirm?
tested in odoo v9.
thx Ron

PR #86, opened 2016-03-14
[FIX]:Pos FrontEnd onchange product tax_ids unassigned taxes

percevaq:
In the POS BackEnd, taxes are not allocated in the order lines.

### DIFF
--- a/pos_pricelist/models/point_of_sale.py
+++ b/pos_pricelist/models/point_of_sale.py
@@ -70,6 +70,17 @@ class PosOrderLine(models.Model):
     price_subtotal = fields.Float(compute="_amount_line_all", store=True)
     price_subtotal_incl = fields.Float(compute="_amount_line_all", store=True)
 
+    @api.multi
+    def onchange_product_id(self, pricelist, product_id, qty=0,
+                            partner_id=False):
+        result = super(PosOrderLine, self).onchange_product_id(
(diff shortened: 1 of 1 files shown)


legalsylvain:
Hi @percevaq, did you sent your CLA ? I'd like to review and merge your PR.

Kind regards.

percevaq:
Hi legalsylvain, it's already sent and confirmed.

Regards.

legalsylvain:
Hi again. Thanks for your CLA and this usefull fix. I just wonder if this fix should be better in OCB. It doesn't seem to be related to pos_pricelist. What do you think ?

legalsylvain:
Hi again. Thanks for your CLA and this usefull fix. I just wonder if this fix should be better in OCB. It doesn't seem to be related to pos_pricelist. What do you think ?

CC : @StefanRijnhart, @pedrobaeza, @antespi, @rafaelbn

kind regards.

pedrobaeza:
I agree with @legalsylvain that this onchange method is not related to the module _pos_pricelist_.

legalsylvain:
@pedrobaeza thanks for your PoV.

@percevaq, could you make a PR again this branch ? https://github.com/OCA/OCB/tree/8.0

It will be in old API however.

https://github.com/OCA/OCB/blob/8.0/addons/point_of_sale/point_of_sale.py#L1366

kind regards.

PR #69, opened 2015-12-15
[CHG] Update gitignore according with https://github.com/OCA/maintainer-tools/blob/master/.gitignore

adrienpeiffer:

PR #88, opened 2016-03-15
[FIX] Rounding problem when partner is selected afterwards

StefanRijnhart:
Followup of #69. The bug also hid out in another part of the code where the prices of existing lines are recalculated when selecting a partner on an ongoing order.


pedrobaeza:
:+1:

lmignon:
:+1: @StefanRijnhart I suppose that the followup is of #79

legalsylvain:
Thanks. (Runbot error is not related.)

PR #94, opened 2016-04-06
Module to integrate MEV/SRM to Odoo POS (Quebec)

philchamp:
Create a module for the Odoo's Restaurant POS so it can work with the SRM/MEV billing in Quebec.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,10 @@ addons:
       - python-lxml  # because pip installation is slow
 
 env:
-  - VERSION="8.0" LINT_CHECK="1"
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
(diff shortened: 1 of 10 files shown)


pedrobaeza:
This should goes in the l10n-canada repository, as is something specific of this country.

philchamp:
You're right. I'll move it. Thanks

On Wed, Apr 6, 2016 at 11:29 AM -0700, "Pedro M. Baeza" notifications@github.com wrote:

This should goes in the l10n-canada repository, as is something specific of this country.

---

legalsylvain:
Hi @philchamp,
Indeed, the title seems to mention something relative to l10n-canada localization, so could you propose your PR in this repository.
Once accepted, we can add a link from the github home page of OCA / pos repository to mention that a pos module exist in canada localization.
thanks.

PR #98, opened 2016-04-13
[IMP] pos_pricelist, not load supplierinfo for not available products

ismaelcj:
Avoid loading the supplier info from products not available in pos. When you have many products, it takes too long to load this information.


legalsylvain:
:+1: Thanks for this improvement.

pedrobaeza:
:+1:

Issue #90, opened 2016-03-18
[8.0] If start with using product template in price-list rule it is not working


krupesh-weboffice:
Hello,
When I am using this module for pricelist it is not working. It is only working with the If we use only products or categories it works perfect , once we use product template in any of the rule all products after that rule( same or higher sequence) will have same price which is completely messing up the pricing . 

Module: pos_pricelist

@StefanRijnhart 
@antespi 
@pablocm-aserti 
@AdilHoumadi 
@JavierIniesta 
@pablocm-aserti

legalsylvain:
Do you use pos_product_template too ? or only pos_pricelist ?
regards.

krupesh-weboffice:
@legalsylvain  
No, I am only using pos_pricelist.

rafaelbn:
There is a PR for solving this issue in https://github.com/OCA/pos/pull/100

krupesh-weboffice:
@rafaelbn thanks for your reply. I will have a look to it and also I fixed that bug earlier by my self.

again thanks

PR #100, opened 2016-05-06
[Fix][8.0] Fix the way pricelists are selected when a template_id is set as rule of pricelist

llacroix:
Added a method that help matching which pricelist can be used.

It can be inherited which allow other modules to extend the way the
pricelists are chosen.

Also a rule in the while loop could be removed as it duplicated
one of the tests when chosing the available pricelists.

Fix the issue here:
https://github.com/OCA/pos/issues/90

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -389,7 +389,16 @@ function pos_pricelist_models(instance, module) {
          * @param version
          * @returns {boolean}
          */
-        rule_match: function (item, product, qty, version) {
+        rule_match: function (db, item, product, qty, version) {
+          // get categories
+          var categ_ids = [];
+          if (product.categ_id) {
+              categ_ids.push(product.categ_id[0]);
+              categ_ids = categ_ids.concat(
+                  db.product_category_ancestors[product.categ_id[0]]
+              );
+          }
+
           var cond = true
           var template_false = item.product_tmpl_id === false
           var product_false = item.product_id === false
@@ -452,20 +461,12 @@ function pos_pricelist_models(instance, module) {
                 return false;
             }
 
-            // get categories
-            var categ_ids = [];
-            if (product.categ_id) {
-                categ_ids.push(product.categ_id[0]);
-                categ_ids = categ_ids.concat(
-                    db.product_category_ancestors[product.categ_id[0]]
-                );
-            }
 
             // find items
             var items = [], i, len;
             for (i = 0, len = db.pricelist_item_sorted.length; i < len; i++) {
                 var item = db.pricelist_item_sorted[i];
-                if (this.rule_match(item, product, qty, version)) {
+                if (this.rule_match(db, item, product, qty, version)) {
                   items.push(item)
                 }
             }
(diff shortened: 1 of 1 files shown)


pos_pricelist/static/src/js/models.js
```diff
@@ -381,6 +381,53 @@ function pos_pricelist_models(instance, module) {
...
+          cond = cond && (
+            (product_false && template_false) ||
+            (product_match && (template_false || template_match)) ||
+            (template_match && product_false)
+          )
+
+          // Check category
+          cond = cond && (
+              item.categ_id === false ||
+              categ_ids.indexOf(item.categ_id[0]) !== -1
```
StefanRijnhart, line 424: Where is categ_ids defined?

pos_pricelist/static/src/js/models.js
```diff
@@ -418,12 +465,8 @@ function pos_pricelist_models(instance, module) {
...
             for (i = 0, len = db.pricelist_item_sorted.length; i < len; i++) {
                 var item = db.pricelist_item_sorted[i];
-                if ((item.product_id === false
-                    || item.product_id[0] === product.id) &&
-                    (item.categ_id === false
-                    || categ_ids.indexOf(item.categ_id[0]) !== -1) &&
-                    (item.price_version_id[0] === version.id)) {
-                    items.push(item);
+                if (this.rule_match(item, product, qty, version)) {
+                  items.push(item)
```
StefanRijnhart: If we only work with valid rules from here, can we perhaps break when we find the first valid rule?

StefanRijnhart:
Thanks for the fix! I added some small discussion points to the code.

llacroix:
I fear it might not be that simple. I remember it was working with pos_pricelist alone, but apparently with pos_template there are still some problems :(

pos_pricelist/static/src/js/models.js
llacroix, line 424: Ah this seems to be missing when I moved some code to that method. I'll fix that.

llacroix:
Ah non forget my previous comment, I just forgot to move some code in the function I added. The problem I had is something completely different.

llacroix:
Any news about this patch?

rafaelbn:
Hi @ismaelcj  @jcarlosmontoya @AdilHoumadi  , have you experimented this issue with pos_pricelist? Maybe you could take a look here. Thanks!

rafaelbn:
Hi @krupesh-weboffice, this PR resolves your issue https://github.com/OCA/pos/issues/90, could you test it please? Thanks!

rafaelbn:
Hi @llacroix , there have been some changes in travis. Please could you rebase this PR in order to check it again? Thanks

llacroix:
I guess I can do that. A bit later today.

legalsylvain:
@llacroix : could you rebase ?
@krupesh-weboffice : could you test this patch ? 

thanks.

llacroix:
Rebased

pedrobaeza:
Is this OK? Or should it be closed?

StefanRijnhart (approved):
Code review, no test.

pedrobaeza:
Merging this after a while with one verified reviewer and green status.

PR #103, opened 2016-05-17
[FIX] pos_pricelist: support attribute prices on product variants

StefanRijnhart:
This change adds attribute value prices when a product variant price is calculated using the POS pricelist mechanism, analogous to https://github.com/odoo/odoo/blob/8.0/addons/product/product.py#L633

_How to reproduce_
Create a product with an attribute with two values. Assign prices to each of the values. The prices of the resulting product variants will be composed of the product price plus the attribute value prices. With pos_pricelist installed, without this change, the variants will be listed with the product prices without the attribute value prices.


pedrobaeza:
:+1:

omar7r:
👍

woepwoep:
I checked and it works fine. Thx @StefanRijnhart

PR #111, opened 2016-07-28
[FIX] pos_pricelist, format_pr error when precision undefined

ismaelcj:
In the piece of code:

``` javascript
if (precision && (typeof this.pos.dp[precision]) !== undefined) {
    decimals = this.pos.dp[precision];
}
```

If `this.pos.dp[precision]` is undefined, the typeof will be "undefined" (string), so will never be `undefined`.

I don't know if there is any reason to use the `typeof` here.


pedrobaeza:
:+1:

legalsylvain:
:+1: thanks.

rafaelbn:
:+1:

PR #115, opened 2016-09-04
9.0 pos access right

legalsylvain:
This module allow you to reduce the PoS rights for some users. For example, you can set that a user can not use discount option, or changing unit prices.
6 options are available : 

new_groups

If you change PoS user during the session (by clicking on the name of the user), the right will be updated.

A message will be displayed, if the user try to use a unauthorized feature : 
demo_error

Works on production since 2 months. Review welcome.


mcassuto:
+1 (I have no rights, but cool feature !)

pos_access_right/models/pos_config.py
```diff
@@ -0,0 +1,88 @@
...
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models, api
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    group_pos_negative_qty = fields.Many2one(
```
simahawk: aren't these fields supposed to have `_id` suffix?

legalsylvain: You're right, I'll fix that.

simahawk: if you want to keep field names shorter, you can drop `_pos_`... we know we are talking about "pos" things here ;)

legalsylvain:
@phuctran1105 : I fixed the bug you mentioned. Could you test ?

@simahawk : I removed "_pos_" prefix and added "_id" suffix, according your remark.

Thanks for your review.

rafaelbn:
Please @legalsylvain could you rebase?

mslokhat:
Hello
my 2 cents (i already asked Yelizariev to add this in his module) :
add a setting to prevent the staff to make a payment. it can be useful for staff taking only orders and only a specific cashier can take the money
Rgds

legalsylvain:
@rafaelbn : rebase done.
@mslokhat : yes this feature could be integrated into this module, but I have not time for that.

regards.

rafaelbn:
Hi @legalsylvain , thanks. Tested in runbot 👍 

But please if possible add a section in user setting for this group. If not when mor modules installed it's not clear. Not blocking.

rafaelbn (approved):

legalsylvain:
Hi @rafaelbn : indeed, it will be more easy to configure with a given "point of sale" group. But this group should be in the core module, and not in this one.
Alternatively, I can create a "Point Of Sale Access Right" Group if you want. 

What do you think ?

legalsylvain:
Well, let's go merging, and group refactoring could be done in a second time.

rafaelbn:
👍

PR #127, opened 2016-10-25
[ADD] New module: pos_save_client_error_fix

zamberjo:
The PoS displays a generic error message when an error occurs while creating users.

This module improves the error handling providing a useful message to the end user.


pos_save_client_error_fix/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+POS Save Client Error Message FIX
+==============
+
+This module extends the functionality of POS and displays the error message returned by the server rather than "Your Internet connection is probably down.".
+
```
rafaelbn, line 10: I've tested in runbot and I get the message "You Internet connection is probable down" so I don't know what to expect with this module.

Could you please explain in the "Usage" seccion how to reproduce the functionality of the module please?

rafaelbn:
Thanks @zamberjo for this contribution. Please, rebase and after this review travis error to get all green.

In my opinion we should remove "fix" from the module's name. This module is not a fix is an improvement as I understand. The readme is not very clear but we are talking about "creating user" so we mean customers, for me a better approach of the name is 'pos_customer_save_error. Maybe here @pedrobaeza could give an opinion

rafaelbn (changes requested):
I am not able to test this module because I don't understand what I must test (improve README please)

pos_save_client_error_fix/static/src/js/pos_save_client_error_fix.js
```diff
@@ -0,0 +1,46 @@
+/* © 2016 Diagram Software S.L.
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+(function ($) {
+    "use strict";
+    openerp.pos_save_client_error_fix = function (instance, module) {
```
legalsylvain, line 6: I think that the header is unnecessary. (function($)) and jQuery at the end.

See more lighter header here : 
https://github.com/OCA/pos/blob/8.0/pos_customer_required/static/src/js/pos_customer_required.js#L1

pos_save_client_error_fix/static/src/js/pos_save_client_error_fix.js
```diff
@@ -0,0 +1,46 @@
...
+                if (this.uploaded_picture) {
+                    fields.image = this.uploaded_picture;
+                }
+
+                fields.id           = partner.id || false;
+                fields.country_id   = fields.country_id || false;
+                fields.ean13        = fields.ean13 ? this.pos.barcode_reader.sanitize_ean(fields.ean13) : false;
+
+                new instance.web.Model('res.partner').call('create_from_ui',[fields]).then(function(partner_id){
+                    self.saved_client_details(partner_id);
```
legalsylvain, line 35: Is it not more correct to call this._super(partner_id) like here : 
https://github.com/OCA/pos/blob/8.0/pos_customer_required/static/src/js/pos_customer_required.js#L22

(just a question, I'm not sure).

pos_save_client_error_fix/README.rst
legalsylvain: @rafaelbn, how did you tested the module ? If you loose the connection, this message is normal. The module is changing message if connection is OK, but customer creation failed, as far as I understand.

rafaelbn:
Closing as old PR without answer in the reviews

PR #129, opened 2016-11-05
WIP - [9.0] port barcodes generate

legalsylvain:
Give possibility to generate barcode.
Still WIP, comments are welcome but it's not ready to merge.

* [ ] description outdated
* [ ] python tests to write
* [ ] squach commits

PR #130, opened 2016-11-05
9.0 port barcodes generate

legalsylvain:
Give possibility to generate barcode.
Still WIP, comments are welcome but it's not ready to merge.

* [x] Make Travis Green
* [x] description outdated
* [x] python tests to write
* [x] squach commits
* [ ] renaming module

(second commit #129)


pedrobaeza:
Why is this module here and not in stock-logistics-barcode?

legalsylvain:
@pedrobaeza : allways the first to review ! Thanks !

Well, because it depends on Point of Sale, due to "bad design of Odoo".
This module allows user to generate custom barcode. For that purpose, it uses barcode.rule and barcode.rule are available only if point_of_sale module is installed. My PoV is that it should not be but this is Odoo Core.

I can move it to on stock-logistic-barcode but it will depends on pos, or it will not work.

All it's written here : https://github.com/OCA/pos/pull/130/files#diff-a63d726f23537f3c0a35c821d660101aR101

What do you think ?

pedrobaeza:
OK to let it here then. Thanks for the explanation. Please call it then _pos_barcode_generator_ to highlight this dependency.

legalsylvain:
Well for the name :
- Barcodes and not barcode because original module (odoo) is named barcodes, so for consistency reason...
- I'd like to avoid to set pos_ because it should not be the name. (due to design problem said before). So in a future version, I hope that barcode field on res.partner will be displayed in barcodes view and not in point_of_sale view, and entry menus will be available without point_of_sale module installed. To avoid to have migration scripts to write in a future, I think pos_ prefix should be avoided

Seeing that 2 points, what about 'barcodes_generator' ?

kind regards.

pedrobaeza:
Then if you look the generality, I prefer the other repository to be the host for the module, although the dependency. Hopefully this dependency changes in a future version and relocate module is not so comfortable.

legalsylvain:
I close this PR. 
See : https://github.com/OCA/stock-logistics-barcode/pull/49

Thanks for your review.

PR #135, opened 2016-12-09
[FIX] pos_product_template: fix bug that maintains the limitation of …

jonerikceberio:
…the shown products by variants, and therefore templates with lots of variants are not shown

### DIFF
--- a/pos_product_template/static/src/js/ppt.js
+++ b/pos_product_template/static/src/js/ppt.js
@@ -478,6 +478,28 @@ Overload: point_of_sale.PosDB
                 this.product_attribute_value_by_id[product_attribute_values[i].id] = product_attribute_values[i];
             }
         },
+    
+        /*
+        Overwrite  get_product_by_category so templates with lots of variants do not hide results
+        */
+        get_product_by_category: function(category_id){
+            var product_ids  = this.product_by_category_id[category_id];
+            var list = [];
+            if (product_ids) {
+                var templates = [];
+                var i = 0;
+                var product;
+                while (templates.length <= this.limit && i<5000 && i<product_ids.length) {
+                    product = this.product_by_id[product_ids[i]];
+                    if (templates.indexOf(product.product_tmpl_id)==-1) {
+                        templates.push(product.product_tmpl_id)
+                    }
+                    list.push(product);
+                    i++;
+                }
+            }
+            return list;
+        },
     });
 
 
--- a/pos_product_template/static/src/js/ppt.js
+++ b/pos_product_template/static/src/js/ppt.js
@@ -159,7 +159,7 @@ Define : pos_product_template.SelectVariantPopupWidget
             // Render Attributes
             var attribute_ids  = this.pos.db.attribute_by_template_id(template.id);
             var attribute_list = [];
-            for (var j = 0, len = attribute_ids.length; j < len; j++) {
+            for (var j = 0, len2 = attribute_ids.length; j < len2; j++) {
                 attribute_list.push(this.pos.db.get_product_attribute_by_id(attribute_ids[j]));
             }
             this.attribute_list_widget.set_attribute_list(attribute_list, template);
--- a/pos_product_template/static/src/js/ppt.js
+++ b/pos_product_template/static/src/js/ppt.js
@@ -153,14 +153,14 @@ Define : pos_product_template.SelectVariantPopupWidget
             for (var i = 0, len = variant_ids.length; i < len; i++) {
                 variant_list.push(this.pos.db.get_product_by_id(variant_ids[i]));
             }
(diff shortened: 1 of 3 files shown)


rafaelbn:
Hi @jonerikceberio , thanks for this fix. Could you please explain how to test this fix? Also please, could you rebase in order to re-check what's going on with travis and have runbot available? Thanks again

jonerikceberio:
Solved some of the travis problems, not all as the original module is not mine. 
How to test the fix: 
- Create a product.template with lots of variants, e.g. 50. 
- Duplicate several times. 
- Open POS
Without the fix: You would only see a number of product.templates that sum 70 product.product (in the example, 2)
With the fix: you see all the templates

libreec:
Hi, what do you need to merge this PR?, I think that it should be ported to v9 too.

Thanks a lot.

dansalsero:
Hi, I thinks this PR should be merged, I already tested it and it's working.

Daniel.

pedrobaeza:
Travis is red, so not possible to merge until this is solved.

damendieta:
Hi, travis fails for two really simple to fix problems, there is any way I can get write access to fix them?

legalsylvain:
You can make a PR on that PR. (on Digital5-Odoo/pos  /8.0_fix_product_template_limit )

thanks.

blaggacao:
How I solved it, if anyone's interested (in db model): https://gist.github.com/blaggacao/f180ea5db4f20e563973f6e372eab67e

While getting rid of the `is_primary_variant` splice...

pos_product_template/static/src/js/ppt.js
```diff
@@ -478,6 +478,28 @@ Overload: point_of_sale.PosDB
...
+        Overwrite  get_product_by_category so templates with lots of variants do not hide results
+        */
+        get_product_by_category: function(category_id){
+            var product_ids  = this.product_by_category_id[category_id];
+            var list = [];
+            if (product_ids) {
+                var templates = [];
+                var i = 0;
+                var product;
+                while (templates.length <= this.limit && i<5000 && i<product_ids.length) {
```
blaggacao, line 492: Why 5000? - The next awkward UX behaviour is just luring around the corner..

pos_product_template/static/src/js/ppt.js
```diff
@@ -478,6 +478,28 @@ Overload: point_of_sale.PosDB
...
+                    product = this.product_by_id[product_ids[i]];
+                    if (templates.indexOf(product.product_tmpl_id)==-1) {
+                        templates.push(product.product_tmpl_id);
+                    }
+                    list.push(product);
+                    i++;
+                }
+            }
+            return list;
+        },
```
blaggacao, line 502: search `search_product_in_category` needs also a fix. (at least in v10 this is the case)

blaggacao (changes requested):
Some functional imperfections in my point of view as noted.
Also the splicing method based on primary-variant must be refactored, else it can happen that a search would have a vaild hit but never returns anything, because primary variant happens not to be in the search..

vcloudbcn:
I I managed to eliminate the limit using the following module pos_product_limit_mac5
  odoo v8

hparfr:
Hi everyone.

The PoS limits search and display (with the same limit). So if you have a lot of products, some of them are never shown on searches.
It's just more visible with this module. But it's always the case and it should be resolved out of the context of pos_product_template context.

jonerikceberio:
@hparfr agreed that the Pos limits the searchs. But the problem is that the "object" that limits the search is always product.product (variants). With pos_product_template the limit is STILL the number of product.product in each search, NOT the number of product.template. Product.template is the object shown, so it should be the one governing the limit.
Therefore it IS a problem of pos_product_template: the module changed the object shown but did not change the criteria of the limit.

hparfr:
Try this on a instance without pos_product_template: 
edit odoo_addons/point_of_sale/static/src/js/db.js Change limit to 4.
You will see how this limit make the pos un-usuable.

I will fix this issue in another module.

hparfr:
Here :  #268

rafaelbn:
Hi  @hparfr could you test this PR and give a review?

How to test here: https://github.com/OCA/pos/pull/135#issuecomment-269358853

I'm rebuilding for testing :smiley:

hparfr:
Hi @rafaelbn,

IMO, the clean solution is to backport pos_fix_limit to v8 and remove **get_product_by_category** from this PR.
I don't use this in v8. So if v8 users are satisfied with this fix; I will not ask any change to this PR.

rafaelbn:
@jonerikceberio what do you think about https://github.com/OCA/pos/pull/135#issuecomment-384921159

I sill cannot test it as runbot never ends! :cry:

PR #138, opened 2016-12-09
[ADD] pos_fix_tax_included: correction of the tax included calculatio…

jonerikceberio:
…ns, so it mirrors the one in sale orders


pedrobaeza:
Why not include this in pos_pricelist?

rafaelbn:
Hi @jonerikceberio , thanks for this fix, as @pedrobaeza commented the correct way is to fix it in pos_pricelist and not in a separate module.

Could you please review this FIX of pos_pricelist https://github.com/OCA/pos/pull/100 ? @llacroix and you could review eachother. Thanks!

rafaelbn:
Please review travis and runbot in order to test it functionally (I will), thanks!

jonerikceberio:
The only problem I saw in including it in pos_pricelist is that changes the way that the price is calculated por POS users. Imagine a company that uses odoo for POS selling, and doesn't use sale orders: if this modification is included in pos_pricelist suddenly the prices will change, and they won't understand why.
In my opinion, for including it in pos_pricelist, it should be understood as a BUG, and I'm sure a lot of companies wouldn't agree on this. That's why I made an independent module.
In any case, if the community understands that this is the way to go I absolutely agree, so there is a Single Source of Truth in prices. 
I will review the checks!

rafaelbn:
Thanks for explanation @jonerikceberio , I would like to know the opinion of @legalsylvain, @AdilHoumadi   (original author) and @pedrobaeza before taking a decision. In my opinion pos_pricelist must take in care all possibilities (tax excluded and included).

jonerikceberio:
I concur, but I think the problem is different: the pos_pricelist module does nothing wrong in itself and it can use tax included prices. The problem is how it takes them into account, because the criteria to apply them is different than the sale module. That does not exactly mean that it is wrong. 
So for me the decission is about the community: Obviously we want a unique way of processing the taxes (SSoT), and in that sense the pos_pricelist way of considering the taxes should be understood as a bug and therefore should be changed there. 
Those are my two (overthought) cents!

pedrobaeza:
For me the solution is to include it in pos_pricelist, and put an option at pos.config level to decide which behavior to use. By default, the behavior should be the same, but you can configure to have the one you include.

jonerikceberio:
Solved the linting.
@pedrobaeza : how do we ask the user in the pos.config? "Included taxes included in POS?" I find it difficult to express the problem, and even worse to explain why it should be different in sale.order and pos.order... It is going to be a very long help in the config :(

legalsylvain:
Hi @rafaelbn.  Sorry I didn't saw your comment.
i have not a clear PoV. If it's not a huge work for @jonerikceberio, I'd prefer the approach of @pedrobaeza. But if it's not possible, Better to have this module in OCA umbrella than nothing.

We could plan to merge this feature into pos_pricelist (adding config) in a more recent serie. (9.0 or 10.0).

regards.

jonerikceberio:
Ok, back in business. Merged the module to pos_pricelist as @pedrobaeza suggested, and added an option to pos.config to let the user choose the behaviour. But the pos_pricelist has lots of linting issues...

rafaelbn:
Hi @jonerikceberio 

> Ok, back in business. Merged the module to pos_pricelist as @pedrobaeza suggested, and added an option to pos.config to let the user choose the behaviour. But the pos_pricelist has lots of linting issues...

So I'm closing this PR, OK? Let me know please

jonerikceberio:
I don't understand why it is being closed: there is a clear inconsistency between how sale orders and pos orders understand included taxes. Even worse, it is opaque to the user, i.e. the user will never understand why a sale order amount and the pos order amount with the same product do NOT match, and he cannot fix it. @pedrobaeza asked to be optional and put it in pos_pricelist, and it has been done.
So IMO it should not be closed.

pedrobaeza:
You have to rebase and fix conflicts then.

rafaelbn:
Hi @jonerikceberio , sorry I closed because I red wrong you message. I reopen and we can review you PR is you resolve conflicts and all bots are :green_heart: :smile: Thank you!

PR #140, opened 2016-12-13
[FIX][10.0] travis: Update travis file to new standard

lasley:
Hopefully this fixed runbot issue from #139

Issue #139, opened 2016-12-13
Runbot is failing in /OCA/pos


rafaelbn:
Hi @legalsylvain @gurneyalex ,

Do you know why we cannot rebuild in https://runbot.odoo-community.org/runbot/repo/github-com-oca-pos-184 to test modules? What does it mean "Skipped" state?

Thanks

![2016-12-13_15-17-42](https://cloud.githubusercontent.com/assets/8736623/21143606/a08fe68e-c147-11e6-95a8-0218cc5334f7.png)

gurneyalex:
Hello,

This is likely a travis file configuration issue. Can you check if there
is a line in .travis.yml env with TESTS=1. The runbot will use the first
of these to get a configuration.

@dlasley can probably help (I'm working on site with a customer this
week with way too little time for the OCA... sorry about this...)


-- 
Alexandre Fayolle
Chef de Projet
Tel : +33 4 58 48 20 30

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac Cedex
http://www.camptocamp.com

rafaelbn:
@lasley could you help?

lasley:
@gurneyalex - Money > OCA

@rafaelbn - I need runbot logs to tell for sure, but I'll give this a shot with my standard fixes & hope for the best. PR incoming

lasley:
Alright the build is running, so the Travis file was indeed the issue. #140 for v10 & I'll back port once merged. This can be closed & we can just track in PR

rafaelbn:
Thanks @gurneyalex @lasley !!! Let's see if runbot works after this PR 😄

pedrobaeza:
Closing this as the fix has been proposed.

PR #141, opened 2016-12-13
[FIX][9.0] travis: Update travis file to new standard

lasley:
* Add new matrix
* Add transifex support
* Switch coveralls to travis_after_tests_success
* Move pip requirements to requirements.txt

Related #140  #139


pedrobaeza:
Fast-tracking as well.

PR #145, opened 2016-12-14
[ADD][8.0][pos_product] New module to be able to configure the default numpad button in POS depending on a product

wpichler:
Does add new module to be able to configure the default numpad button in POS depending on a product.
For example:
 If you sell apples - you can set the default button to be Quantity
 If you sell a product, where you often change the price - then the default button will be Price on the NumPad

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ pos
 Point of sale
 
 [//]: # (addons)
+
 Available addons
 ----------------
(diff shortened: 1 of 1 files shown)


rafaelbn:
Hi @wpichler , thanks for this contribution:

Please follow instructions in https://github.com/OCA/maintainer-tools and module template

PR #153, opened 2017-02-26
[FIX] add python file

cvinh:


pedrobaeza (changes requested):
Move the file to the models directory and use relative import: `from . import...`

cvinh:
I moved the product.py file in v10, in v8, I just fixed the problem

pedrobaeza:
Is this working now with only this? My doubt is why the JS doesn't give error before if the field doesn't exist (as the file was not loaded).

cvinh:
I did the tests, Yes indeed, it works now
It did not show error because the field has_image did not exist
By the way, there was error in odoo logs

pedrobaeza (approved):

pedrobaeza:
OK, thanks

@JGarcia-Panach please review

pedrobaeza:
Oops, Travis fails.

cvinh:
yes I see, I think th v8 never worked because the python.py was never launched... for v10, I changed a few things. I am going to backport them

pedrobaeza:
Cool, I was not able to review them as I'm going to holidays today, but count my approval trusting in your changes.

cvinh:
PR updated.
hummm. I did not test the v8 neither... I think we need 2 more approvals ;-)

JGarcia-Panach:
Tested on runbot and local enviroment, with the same result.

image

The function expects a singleton.

pos_default_empty_image/product.py
```diff
@@ -9,12 +9,8 @@ class ProductTemplate(models.Model):
...
-            record.has_image = bool(record.image)
+    def _get_has_image(self):
+        self.ensure_one()
+        self.has_image = self.image is not False
 
-    has_image = fields.Boolean(
-        compute='_has_image',
-        store=True,
-        readonly=True)
+    has_image = fields.Boolean(compute='_get_has_image', string='Has Image')
```
legalsylvain, line 16: * please keep has_image stored for performance purpore
* rename _get_has_image by _compute_has_image
* keep for record in self, to avoid singleton error. (well, the original function was working correctly)

legalsylvain (changes requested):
hi @invitu,
some comment inline.

Thanks for your work.

legalsylvain:
Closing this PR. superseded by https://github.com/OCA/pos/pull/165.

PR #157, opened 2017-03-02
[ADD] add missing pos_access_right for 10.0

cvinh:
the module is not migrated yet, just add it because it's missing in 10.0 branch
(Installable = False)


legalsylvain:
Hi @invitu . This is not the good way to recover work from old branch. 
You have to follow the method described here : 
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

This way, you'll keep original author, and be sure that all the commit has been taken into account.

Sample : https://github.com/OCA/pos/pull/160

PR #165, opened 2017-04-17
[8.0] pos_default_empty_image fix import

legalsylvain:
Trivial fix. 
supersed : https://github.com/OCA/pos/pull/153

PR #166, opened 2017-04-18
[FIX] import and error in product code

legalsylvain:
supersed : https://github.com/OCA/pos/pull/165.

(the old PR was in the bad repository)

CC : @pedrobaeza, @invitu

### DIFF
--- a/pos_default_empty_image/__init__.py
+++ b/pos_default_empty_image/__init__.py
@@ -1,0 +1,1 @@
+from . import product
--- a/pos_default_empty_image/product.py
+++ b/pos_default_empty_image/product.py
@@ -5,16 +5,13 @@
 from openerp import models, fields, api
 
 
-class ProductTemplate(models.Model):
-    _inherit = ['product.template']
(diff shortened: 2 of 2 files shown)


NL66278 (approved):
I made similar (but more extensive) changes in this PR: https://github.com/OCA/pos/pull/167/files. In my PR I also standardized the directory structure. If you agree with my changes, I can rebase my PR on yours and add it to your PR. Or you can close this PR in favour of mine. If you do not agree, I approve this PR anyway, so we can get ahead with this important fix.

legalsylvain:
@NL66278 you let the field on template, but I think it should be set on product. Don't you think ? 
Otherwise, no problem to close my PR, and keep yours.

regards.

NL66278:
@legalsylvain Looking into it, it probably is better to set the field on the product. I changed my PR accordingly.

legalsylvain:
supersed by https://github.com/OCA/pos/pull/167.
closing.

PR #167, opened 2017-04-28
[8.0][FIX] Make sure module pos_default_empty_image correctly and completely loaded.

NL66278:
model was not really loaded, therefore no image would be shown at all.

also fixed directories according to OCA standards.


legalsylvain (approved):
Thanks A lot ! @OCA/pos-maintainers : please review this bugfix. 
regards.

rafaelbn (approved):

PR #168, opened 2017-05-02
[8.0][ADD] New module to only load specific customers in POS.

NL66278:
Module helps in fast loading of POS. Also protects against spreading privacy sensitive customer data.


pos_anonymous_customer/models/res_partner.py
```diff
@@ -0,0 +1,12 @@
...
+# © 2017 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    available_in_pos = fields.Boolean(
+        string='Available for POS',
```
legalsylvain: How do you see default value : 
1> currently : default : none
2> could be default = is_customer (to initialize correctly existing partners).
3> or maybe =True, if customer is created via the UI : not sure if it is easy to catch.

In anyways, an api@depend, setting available_in_pos to False if is_customer is set to False could be great ? Don't you think ?

legalsylvain (commented):
Hi @NL66278. Good improvment indeed ! 
code review. LGTM.
2 remarks :  

1) i think that the name pos_anonymous_customer, doesn't feat with the feature. it should be better pos_restricted_customer_list, or something else... customers data are not anonymized, if I understood this module. (for the time being).
2) in a multi PoS / multi company context, It should be great to add a new setting field on pos.config. (selection). with the following option. ' customer_load' : 
- 'no_one' : disable customer loading. (usefull for self service pos, without customer feature)
- 'selection' : the feature you implemented. (load only checked customers). (default value).
- 'all' : for super cashier, (after sale PoS), etc.

what do you think ? 

If it's not in the scope of your PR, but you think it could be great, just add a comment in you README file in the roadmap section.

regards.

pos_anonymous_customer/models/res_partner.py
NL66278: @legalsylvain The default is quite deliberately to not load the customer when not specifically set. We have a project with > 132.000 customers (they also have a webshop). Loading the customer data takes minutes at the physical shop they have, and is never used.

If a partner is no longer a customer, he/she will not be loaded in the POS anyway, as in the end I did not replace the default domain - [['customer','=',true]] - but appended the check for available_in_pos.

I will rename the module according to your suggestion.

Also the ideas for a more flexible approach are good, but indeed not (yet?) in the scope of what we need right now.

legalsylvain: > The default is quite deliberately to not load the customer when not specifically set. 

I understalease, just add a comment to keep that future feature in the road mapnd, you're right, this is the expected behaviour. if not desired, a mass edit on res.partner, will easily realise the work.

About the more flexible approach. I'll add this feature, when i'll use this module. let's go accepting this PR as a first good step.

regards.

legalsylvain (approved):

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
```diff
@@ -0,0 +1,25 @@
...
+			// check if not already done by someone else
+            for(var i = 0; i < m.domain.length; i++) {
+                var domain_tuple = m.domain[i];
+                if (domain_tuple[0] === 'available_in_pos') {
+                    return true; // domain already added
+                }
+            }
+            m.domain.push(['available_in_pos','=',true]);
+			return true; // no need to continue
+		});
```
hparfr, line 22: Your intent is unclear because of the indentation or {} or the style (for loop in function) ?

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
```diff
@@ -0,0 +1,25 @@
...
+
+	// we can't extend it because self.pos not ready yet
+	var _initializePosModel_ = module.PosModel.prototype.initialize;
+	module.PosModel.prototype.initialize = function(session, attributes){
+		// override domain for res.partner to limit customers loaded
+		this.models.some(function (m, idx) {
+			if (m.model !== 'res.partner')
+				return false;
+			// check if not already done by someone else
+            for(var i = 0; i < m.domain.length; i++) {
```
hparfr, line 14: take care of var hoisting

hparfr:
Prevent the loading of all the customers will boost the loading time of the POS drastically.
👍

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
NL66278: @hparfr You are seeing the indentation (which of course makes no technical difference in javascript, unlike python) differently from me. This is due to tabs in the file instead of spaces. I will correct this.

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
NL66278: @hparfr Done. And made some more changes to make the code pass tests on jslint.com.

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
hparfr: Indeed, it's better if the indentation is consitent in the file.

What I mean by style is that you have a `for` (procedural) in a `some` (functionnal).
You can use a `forEach` or `some`: 

```js
		this.models.some(function (m) {
			if (m.model !== 'res.partner')
				return false;

			var already_added = m.domain.some(function (dom) {
				return dom[0] == 'available_in_pos';
			})
			if (!already_added)
				m.domain.push(['available_in_pos', '=', true]);
		});
```
(not tested)

hparfr: I don't remember if the case of more than one entry of _res.partner_ in _this.models_  is possible:  
```js
this.models.filter(function (m) {
	return m.model == 'res.partner';
}).forEach(function (m) {
	var already_added = m.domain.some(function (dom) {
		return dom[0] == 'available_in_pos';
	})
	if (!already_added)
		m.domain.push(['available_in_pos', '=', true]);
});
```

rafaelbn (approved):
Thanks @NL66278

PR #170, opened 2017-05-05
[ADD] module pos_empty_home

legalsylvain:
New simple module that hide products at the start of the PoS, or when no category are selected and no search by name is pending. It reduces display time and avoid user to make errors, selecting undesired products.

Home Page of the PoS, once the module is installed.

pos_empty_home
 
review welcome.


rafaelbn (approved):
Tested in runbot 👍

JGarcia-Panach:
Tested in runbot, works really fine 👍

PR #172, opened 2017-05-16
[ADD] start module pos_order_no_offline, the aim is to have a blocking synchronisation in order to be sure to synchronise the order without any posibility to destroy it

sebastienbeau:
Regarding the new finance law in France we need to be sure that a pos order can not be destroy

Right now it's easy
-  run the point of sale in a private navigation
- cut the network connection
- sale some product (with ticket)
- close the private navigation
- reconnect the network and reopen the browser in a no private navigation

As no order have been send to the Odoo server you data have been drop.

We think about implementing an hash on server side and removing offline mode for sending the order (no ticket without sending the order to odoo). We can also add the first number of the hash on the ticket

For now I just play with the pos in order to test if we can remove the offline mode.


pos_order_no_offline/README.rst
```diff
@@ -0,0 +1,38 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+POS Order No offline mode
+===========================
+
+This module remove the possibility to generate a pos ticket without connexion.
+This mean that no pos order can be drop
```
bealdav, line 8: s/drop/dropped

legalsylvain:
Hi @sebastienbeau. Thanks a lot for this PR. Indeed the offline mode can be saw as a possibility to cheat with the Sapin law. In another hand, cutting offline mode will be a problem for a lot of shops, specially the big ones. I can not figure out to sell a PoS system to a customer if I say, "your cashiers can not work if you have an internet trouble, or your cashier should know how to give Wifi access to the PoS with a phone". It's possible in a little shop, really not in a supermarket.
So, maybe a more flexible approach should be having a setting on pos.config with the following options : 
- 'block' : the feature you describe. It is not possible to print a ticket without internet connection.
- 'warning' : the same feature but if the connexion is lost, a warning is printed on the ticket, saying : 'Internet Trouble, the ticket has been edited without uniq number....blabla..."

What do you think ? I can implement the second feature. 

Note : it's possible to protect users to use this feature, adding a setting in a config file like 'pos_warning_mode' allowed. (False by default).

sebastienbeau:
HI Sylvain I agree that sharing the wifi of the phone is not a solution for big shop (it only work for little shop).

For big shop I see too possibility:
- having at least 2 Internet connexion with an automatic fail-over or better an overthebox https://www.ovhtelecom.fr/overthebox (this solution is amazing it share the bandwith beetween all your box and do not fail if one connexion is down)
- having a local odoo (with external replication for backup)

Regarding the law I do not think having a warning is enough... and if we really want something secure I do not see other solution then cutting the offline mode.

But I think it's acceptable in France Internet connexion are good. A good proof is how many company are already internet/telephony depend ? The best example is the telephony, how many company use asterisk (or other solution) and it's critical for there business?

Moreover the POS do not really work in offline mode (for now) there is no possibility to open a session without internet and no possibility to print an invoice too.

But we have too be aware that having 2 connexion have an extra-cost and so only the big one can paid for it. But this is always the same thing risk vs cost. If the cost of the failure is important you need to invest in solution, paying for 2 or 3 connexion is for me totally acceptable. If you are a really small shop and the cost of failure is not important you can always use your smart phone as a backup.

What do you think?

legalsylvain:
Hi @sebastienbeau.

Thanks for your response. 

> and no possibility to print an invoice too.

We are talking about degraded mode. It's "normal" that in degraded mode, some features are anavailable.

I'm fully OK that having a double connexion reduces problems, but it can not be sufficient in some cases. (server problem, failover system problem, etc...) you'll have in many times some PoF in your infrastructure.
I just defend the PoV that once launched at the beginning of the day, the PoS *should* work.
So my proposal is to have an optional option ;-) that can be enabled. It's quite easy to add this feature in that module because the module will block printing and this feature too, and your module will add extra info on the receipt and this feature too.
If this feature is only available if it is enabled in a config file, is it ok for you ?

legalsylvain:
In accordance with @sebastienbeau .
this PR is superseded by https://github.com/OCA/l10n-france/pull/108.

PR #173, opened 2017-05-19
[10.0][MIG] pos autoreconcile mgs

GillesTephaneMeyomesse:


rafaelbn:
Hi @GillesTephaneMeyomesse , thanks for this contribution. As you can see coverage has decreased -22.06% please make test to at least maintain the same coverage modules had. It has reduce to much. thanks!

pos_autoreconcile/models/point_of_sale.py
```diff
@@ -0,0 +1,50 @@
...
+    def _create_account_move_line(self, session=None, move=None):
+        to_ret = super(PosOrder, self)._create_account_move_line(
+            session=session, move=move)
+
+        account_def = self.env['ir.property'].get(
+            'property_account_receivable', 'res.partner')
+
+        grouped_data = {}
+
+        for order in session.order_ids:
```
benwillig: ```session``` might be equal to None so there can be an exception here.

pos_autoreconcile/models/point_of_sale.py
```diff
@@ -0,0 +1,50 @@
...
+            current_company = order.sale_journal.company_id
+            order_account = (
+                order.partner_id and
+                order.partner_id.property_account_receivable_id.id and
+                order.partner_id.property_account_receivable_id.id or
+                account_def and account_def.id or
+                current_company.account_receivable.id
```
benwillig: ```res.company``` model has no field named ```account_receivable```. And I think that the sale journal is not mandatory on a ```pos.order```.

benwillig (changes requested):

legalsylvain (changes requested):
Hi @GillesTephaneMeyomesse. I think you have recovered undesired commits from old branch. you should filter by the module name.
kind regards

GillesTephaneMeyomesse:
@legalsylvain. You have right, too fast commit sorry...
That is fixed thanks

legalsylvain (commented):
ok thanks ! 
(I have no time to review this PR for the time being, but I'm not blocking this PR anymore. I can not remove my objection, except if I accept the PR)

Cedric-Pigeon:
I think this PR is useless. Feature seems to be covered by core in V10.0

sbidoul:
It's implemented here in Odoo 10 https://github.com/odoo/odoo/blob/6fa705399bc05d4b0dce5e7181c40a4a5ab8b1c3/addons/point_of_sale/models/pos_order.py#L338

legalsylvain:
closing, due to last remarks.

PR #174, opened 2017-05-21
[FIX - 10.0 hw_telium_payment_terminal] setup parameters are wrong in README

cvinh:
config parameters are telium_terminal_device_* instead of payment_terminal_device_*


cvinh:
cc @alexis-via @legalsylvain

legalsylvain:
ping @alexis-via.

alexis-via:
hw_telium_payment_terminal is only maintained in v8 because POSbox uses Odoo v8.

PR #175, opened 2017-05-21
[FIX 10.0] hw_* modules are made to run on posbox... 

cvinh:
Manifest files still are `__openerp__.py` in posbox


legalsylvain (changes requested):
I don't understand what you propose. In v10, all modules have manifest and not openerp file, including hw_ files.
cf : https://github.com/OCA/OCB/tree/10.0/addons/hw_scanner

thanks for your explanation.

cvinh:
@legalsylvain the hw modules run on posbox that is still in v8

cvinh:
Steps to reproduce : 
- install posbox from https://nightly.odoo.com/trunk/posbox/
- install hw_* modules on the posbox
`__manifest__.py` file is not taken into account

cvinh:
If you consider that hw_* modules shoud be taken from 8.0 branch, then this PR is to be rejected... and that point may be added to the documentation

legalsylvain:
Well, sorry, I don't use posbox.

I don't understand why posbox is build only on V8 version. It should be build for each serie. Don't you think ? As said, in v10 all modules of Odoo Core have manifest file, including hw_ https://github.com/OCA/OCB/tree/10.0/addons/hw_scanner. 

> If you consider that hw_* modules shoud be taken from 8.0 branch, then this PR is to be rejected... and that point may be added to the documentation

Indeed, 8.0 poxbox should be installed with OCA / pos / 8.0 and 10.0 posbox should be installed with OCA / pos / 10.0 . I see it quite logical, but we can add a line in the documentation if you think it can be usefull.

Solution 1, try / use pywebdriver project. (https://github.com/akretion/pywebdriver) / sudo apt-get install pywebdriver. I think that this project could be under OCA umbrella in a few times.

Solution 2, install a posbox based on odoo 10 + OCA / pos.


@alexis-via, you I think you use posbox, do you have a PoV ? 

Regards.

alexis-via:
I'm very surprised by what @invitu says that Odoo's POSbox image uses odoo v8... but I have never used Odoo's official POSbox image, so I don't know.
I used a POSbox with my own Odoo setup on it, so I didn't have that kind of problem. And now I use pywebdriver, so it's another story...

StefanRijnhart:
Indeed, if you run the official posbox image that is based on Odoo 8.0 you need to use the modules from the 8.0 branch of this project. I think this can be closed.

cvinh:
Yes I agree, please close my PR... maybe you can add that in the documentation
BTW, I did not check the last posbox 1.6 image if it's v8 or v10

PR #176, opened 2017-05-23
[10.0][MIG] pos payment entries globalization

GillesTephaneMeyomesse:


pos_payment_entries_globalization/models/pos_session.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @api.multi
+    def _generate_globalization_entries(self):
+        """Generate globalization moves"""
+        self.ensure_one()
+        grouped_move_lines = self._get_move_lines_for_globalization()
+        to_reconcile = []
+        for key, lines in grouped_move_lines.iteritems():
+            global_account_id, global_journal_id = key
+            move = self._create_globalization_move(global_journal_id)
```
sbidoul: nice to have: _prepare instead of _create for easier extendability

pos_payment_entries_globalization/models/pos_session.py
```diff
@@ -0,0 +1,104 @@
...
+            counterpart_credit = 0.0
+            for line in lines:
+                counterpart_debit += line.debit
+                counterpart_credit += line.credit
+                new_line = self._create_reverse_line(line, move)
+                # Pair to reconcile : payment line and the reverse line
+                to_reconcile.append(line + new_line)
+            if counterpart_debit:
+                self._create_globalization_counterpart_line(
+                    counterpart_debit, 0.0, global_account_id, move)
```
sbidoul: nice to have: _prepare instead of _create for easier extendability

pos_payment_entries_globalization/models/pos_session.py
```diff
@@ -0,0 +1,104 @@
...
+            if counterpart_credit:
+                self._create_globalization_counterpart_line(
+                    0.0, counterpart_credit, global_account_id, move)
```
sbidoul: nice to have: _prepare instead of _create for easier extendability

pos_payment_entries_globalization/models/pos_session.py
```diff
@@ -91,6 +92,7 @@ def _generate_globalization_entries(self):
             if counterpart_credit:
                 self._create_globalization_counterpart_line(
                     0.0, counterpart_credit, global_account_id, move)
+        move._post_validate()
```
sbidoul: Note this is done unconditionally when posting the move, so maybe it's better to not do it here. So in case of problem, the draft move can be fixed by the accountant when posting it?

GillesTephaneMeyomesse:
@sbidoul I will fix the small changes on the next commit and check what you asked for. Thanks

pos_payment_entries_globalization/models/pos_session.py
GillesTephaneMeyomesse: According to my tests it is possible after a difference between the debit and the credit to correct posted move lines

rafaelbn:
Hi  @sbidoul  could you please re-review this PR from @GillesTephaneMeyomesse ? In my side is :+1:

rafaelbn (approved):
Functional review :+1:

rafaelbn:
Hello, just one question? Why do you create the journal item un-posted?

sbidoul:
@rafaelbn we use it in production, so I guess it's fine. @Cedric-Pigeon do you confirm?

Cedric-Pigeon:
@sbidoul @rafaelbn Yes we use it in production since september 2017.

chienandalu:
I'd say the question of @rafaelbn was more on the reason of the behaviour of the account moves registration, keeping those movements unposted pending manual validation.

Cedric-Pigeon:
Honestly I do not remember the reason why as it is like that since v8.0 when the addon has been developped

chienandalu:
I'd say it comes with the migration in this line, but I can't fully understand the comment: https://github.com/OCA/pos/pull/176/commits/1d655173d6893fa193fb6a44f43fabbd2cc4ba1e#diff-f97804b69d94fcbbaff341daad3aa702R70

rafaelbn:
@chienandalu please finish your review and merge!

setup/pos_payment_entries_globalization/odoo/__init__.py
```diff
@@ -0,0 +1 @@
+__import__('pkg_resources').declare_namespace(__name__)
```
chienandalu, line 1: Shouldn't this setup files go away and be added automatically after merge?

Cedric-Pigeon: It is as no influence, the generated one will the same. We add it in the PR to be able to use pip even if not merged.

chienandalu (approved):

chienandalu:
Thanks to all :+1:

PR #182, opened 2017-06-20
[FIX] re-introduce pos_remove_pos_category and avoid monkey patch and error on foreign key constraint

Cedric-Pigeon:
In version 10.0 monkey patches are not used anymore but replaced by an inheritance of BaseModel.
Previously this addon used a monkey patch to bypass a foreign key creation on column pos_categ_id of product_template model. These was triggered by an update of point_of_sale.

Unfortunately in V10.0, the module pos_remove_pos_category is not yet in the registry while the foreign key is created. So it is impossible to avoid it.
So, on an empty database, no problem but once you have one product linked to a product_category, it is impossible to update the point_of_sale module because it cannot find the pos_categ_id value in pos_category table.
This PR intend to fix this problem by switching the comodel on pos_categ_id and using a related as before but NOT STORED. A search method is added to allowing research on the field.

### DIFF
--- a/pos_remove_pos_category/README.rst
+++ b/pos_remove_pos_category/README.rst
@@ -1,0 +1,63 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+POS Remove POS Category
+=======================
+
+This module was written to replace POS categories by product categories
+in the point of sale.
+
+Important notes
+---------------
+- When the module is installed the link beetween products and POS categories
+  is **overwritten** by a link beetween products and product categories
+  (the link is the field pos_categ_id in the table product_template)
+- When the module is uninstalled the link beetween products and POS categories
+  is restored in an **empty** state (NULL values)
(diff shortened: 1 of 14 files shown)


Cedric-Pigeon:
@legalsylvain @sylvainc @mourad-ehm @simahawk

lmignon (approved):
(Code + Functional)

pedrobaeza:
@StefanRijnhart as you have discovered also the problem, can you check this PR (commit a37990f621101ab16ff004a88ad5b0c5ef53bff3) to see if the problem is solved?

StefanRijnhart (approved):

pedrobaeza:
@Cedric-Pigeon one last thing: please change module version (you can even put 10.0.2.0.0, it's a huge approach change).

Cedric-Pigeon:
@pedrobaeza Ok. Done!

simahawk (commented):
LGTM

PR #183, opened 2017-06-21
[ADD] new module pos_picking_load

legalsylvain:
Hi all.

This PR add a new module pos_picking_load that allow you to load pickings created by sale module, in the point of sale, to finish the workflow via the point of sale.

* You can see the whole description in the readme file with screenshots.
https://github.com/grap/pos/blob/8.0_ADD_pos_picking_load/pos_picking_load/README.rst

* In a few words: 

new button available
load_picking_01

List of pickings:
load_picking_02

Once the picking is loaded:
load_picking_03

Thanks for your review.


pos_picking_load/static/src/js/pos_picking_load.js
```diff
@@ -0,0 +1,329 @@
...
+                        message: _t('Connection error'),
+                        comment: _t('Can not execute this action because the POS is currently offline'),
+                    });
+                }
+                event.preventDefault();
+            });
+        },
+
+        on_click_picking: function(event){
+            this.load_picking(parseInt(event.target.parentNode.dataset.pickingId));
```
hparfr: add a radix:
`parseInt(event.target..., 10)`

pos_picking_load/static/src/js/pos_picking_load.js
```diff
@@ -0,0 +1,329 @@
...
+            contents.innerHTML = "";
+            for (var i = 0, len = pickings.length; i < len; i++){
+                var picking = pickings[i];
+                var pickingline_html = QWeb.render('LoadPickingLine',
+                    {widget: this, picking:pickings[i]});
+                var pickingline = document.createElement('tbody');
+                pickingline.innerHTML = pickingline_html;
+                pickingline = pickingline.childNodes[1];
+                pickingline.addEventListener('click', this.on_click_picking);
+                contents.appendChild(pickingline);
```
hparfr: it's better (for the perfs) if you add all nodes in one DOM call.
you may also add the event listener on the collection instead of each row

pos_picking_load/static/src/js/pos_picking_load.js
```diff
@@ -0,0 +1,329 @@
...
+
+        on_click_picking: function(event){
+            this.load_picking(parseInt(event.target.parentNode.dataset.pickingId));
+        },
+
+        render_list: function(pickings){
+            var self = this;
+            var contents = this.$el[0].querySelector('.picking-list-contents');
+            contents.innerHTML = "";
+            for (var i = 0, len = pickings.length; i < len; i++){
```
hparfr: (just for the sytle: you can use `.forEach()`  or `.map()` here instead of for loop.
And in javascript the scope delimited by functions and not {})

pos_picking_load/static/src/js/pos_picking_load.js
legalsylvain: Thanks for your review, I'll do the changes. About that point, could you point me some samples ?

hparfr: http://learn.jquery.com/performance/append-outside-loop/

StefanRijnhart:
Dear Sylvain, you are the POS maestro ;-) This module is a great example of how to extend the POS functionality.

I have no code remarks, and I tested the described functionality succesfully. My only gripe is that there is no connection between the original sale order and the new picking anymore. Did you look into linking the new picking to the sale order through the procurement group, or even reuse the existing stock moves if possible, to improve tracking of products? I can propose a change into your branch if you like.

Or in case you think this change can be considered after the merge of this PR, I will give my approval right away.

StefanRijnhart (commented):
Code review and test

StefanRijnhart (approved):

hparfr:
👍

legalsylvain:
Hi @StefanRijnhart. 
Thanks a lot for your compliment ! 

You're fully right about the lake of connection with the original sale order (forced to 'done') / stock picking (forced to 'canceled') / and stock moves (forced to 'canceled').

The main reasons are :

- At GRAP, I didn't implemented product tracking. (little shops, without lot management).
- I really don't know concepts like procurement group. I'm not a stock expert at all. My use is very basic.

But yes, your idea could be a great improvment. We could imagine to overload create_picking (https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/point_of_sale.py#L830)  function if the pos.order come from an existing picking, and altering the original picking (and moves), instead of creating a new one, and cancelling the previous one, and then setting the altered picking to the picking_id field of the pos.order.

As this PR has two :+1:, I propose to merge it, and to improve it in another PR.

Please ping me when it's done, so that I'll review it.

StefanRijnhart:
@legalsylvain perfect! I will work in it in the next two weeks or so.

PR #185, opened 2017-06-29
[ADD] pos_loyalty: loyalty programs for POS

geluboros:
**Loyalty Program**

This module allows you to define a loyalty program in the point of sale,
where the customers earn loyalty points and get rewards.

This module is a forward-port to v10 of the pos_loyalty module from Odoo's
saas-6 branch.
The functionality was moved to the Enterprise edition in later versions.

### DIFF
--- a/pos_loyalty/static/src/js/pos.js
+++ b/pos_loyalty/static/src/js/pos.js
@@ -276,7 +276,7 @@ odoo.define('pos_loyalty.loyalty_program', function (require){
                     return;
                 }
 
-                var line = this.add_product(product, {
+                this.add_product(product, {
                     price: 0,
                     quantity: 1,
                     merge: false,
@@ -300,7 +300,7 @@ odoo.define('pos_loyalty.loyalty_program', function (require){
                     return;
                 }
 
-                var line = this.add_product(product, {
+                this.add_product(product, {
                     price: -discount,
                     quantity: 1,
                     merge: false,
@@ -328,7 +328,7 @@ odoo.define('pos_loyalty.loyalty_program', function (require){
                     return;
                 }
 
-                var line = this.add_product(product, {
+                this.add_product(product, {
                     quantity: -spendable,
                     merge: false,
                     extras: { reward_id: reward.id },
@@ -384,10 +384,8 @@ odoo.define('pos_loyalty.loyalty_program', function (require){
                     'title': _t('No Rewards Available'),
                     'body':  _t('There are no rewards available for this customer as part of the loyalty program'),
                 });
-                return;
             } else if (rewards.length === 1 && this.pos.loyalty.rewards.length === 1) {
                 order.apply_reward(rewards[0]);
-                return;
             } else {
                 var list = [];
                 for (var i = 0; i < rewards.length; i++) {
--- a/pos_loyalty/static/src/js/pos.js
+++ b/pos_loyalty/static/src/js/pos.js
@@ -225,6 +225,21 @@ odoo.define('pos_loyalty.loyalty_program', function (require){
             }
         },
 
+        has_discount_reward: function(){
+            var res = false;
+            var lines = this.get_orderlines();
+
+            for (var i = 0; i < lines.length; i++) {
+                var line = lines[i];
+                var reward = line.get_reward();
+                if (reward && reward.type === 'discount') {
+                    res = true;
+                    break;
+                }
+            }
+            return res;
+        },
+
         /* The list of rewards that the current customer can get */
         get_available_rewards: function(){
             var client = this.get_client();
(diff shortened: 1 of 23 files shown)


legalsylvain:
Hi @gboros-rgbconsulting. thanks a lot to back-open-port this module. i'm very interested by this feature. 

I justed quickly tested on runbot. some remarks.

1. I don't understand discount feature. Use case with 2% discount / point cost 10 / pp_currency 1 : 
Order 1 : Sale 10€. new point 10.
order 2 : Sale 1000€. rewards : 2% or the current order = 20€.
I was expecting to have a discount based on the points of the customer (past orders). IE rewards = 2% of 10 points = 0,20 €.
do you know if I can configure PoS like that ? Otherwise, I'll propose a PR.

2. I guess it's a bug (or an Odoo feature ;-)) but it is possible to click many times on rewards, that give unexpected result. Don't you think ? What about to block user to click many times on rewards button for a pos.order.

image

legalsylvain:
Point 1. OK, it's not very intuitive, but to do what I want, I have to set a resale reward, with a product price of 0.02. (for 2%). 
sample with a customer with 75 points.
image

point 2. bug for discount product still a problem.

geluboros:
Hi @legalsylvain,
1. Actually, the resale reward is used to convert points into customer credit.

For example, if we set a point product with a price of 0.02 €, a customer with 100 points would have:
100 points * 0.02 € per point = 2 € credit
So if this reward is added to a sale order, a new line with -2 € will be added.
It's still a sort of discount, but it's not a percentage.

2. You are right, a discount reward added multiple times gives an unexpected result. I will review this.

geluboros:
I added a new commit that limits discount rewards to one reward per order.

legalsylvain (approved):
:+1: 
Functional test.

pos_loyalty/static/src/js/pos.js
```diff
@@ -0,0 +1,460 @@
...
+                            self.loyalty.rules_by_product_id[rule.product_id[0]].push(rule);
+                        }
+                    } else if (rule.type === 'category') {
+                        var category = self.db.get_category_by_id(rule.category_id[0]);
+                        if (!self.loyalty.rules_by_category_id[category.id]) {
+                            self.loyalty.rules_by_category_id[category.id] = [rule];
+                        } else if (rule.cumulative) {
+                            self.loyalty.rules_by_category_id[category.id].unshift(rule);
+                        } else {
+                            self.loyalty.rules_by_category_id[category.id].push(rule);
```
hparfr: lines 43:49 and 51:57 can be factorized

pos_loyalty/static/src/js/pos.js
```diff
@@ -0,0 +1,460 @@
...
+            if (!client) {
+                return;
+            } else if (reward.type === 'gift') {
+                var product = this.pos.db.get_product_by_id(reward.gift_product_id[0]);
+
+                if (!product) {
+                    return;
+                }
+
+                var line = this.add_product(product, {
```
hparfr: unused var

pos_loyalty/static/src/js/pos.js
```diff
@@ -0,0 +1,460 @@
...
+                    discount = discount_max;
+                }
+
+                var product = this.pos.db.get_product_by_id(reward.discount_product_id[0]);
+
+                if (!product) {
+                    return;
+                }
+
+                var line = this.add_product(product, {
```
hparfr: unused var

pos_loyalty/static/src/js/pos.js
```diff
@@ -0,0 +1,460 @@
...
+
+                if ( round_pr( spendable * product.price, crounding ) > order_total ) {
+                    spendable = round_pr( Math.floor(order_total / product.price), lrounding);
+                }
+
+                if ( spendable < 0.00001 ) {
+                    return;
+                }
+
+                var line = this.add_product(product, {
```
hparfr: unused var

pos_loyalty/static/src/js/pos.js
```diff
@@ -0,0 +1,460 @@
...
+                        item:  rewards[i],
+                    });
+                }
+                this.gui.show_popup('selection',{
+                    'title': _t('Please select a reward'),
+                    'list': list,
+                    'confirm': function(reward){
+                        order.apply_reward(reward);
+                    },
+                });
```
hparfr, line 398: not consistent: always (like :387, :390) or never add return before `}`

hparfr (approved):
👍 

Some minor comments on the JS code.

geluboros:
Hi @hparfr,
First of all thanks for the revision.

I removed the returns and unused variables.
About the factorization of lines 43:49 and 51:57, I am not sure how it could be done since rules are stored in different arrays > one for products and one for categories.

hparfr:
I did not test, but you may try that way : 

```js

function update_rules(rules, rule, id) {
	if (!rules[id])
		rules[id] = [rule];
	else if (rule.cumulative)
		rules[id].unshift(rule);
	else
		rules[id].push(rule);
}

rules.forEach(function (rule) {
	if (rule.type == 'product')
		update_rules(self.loyalty.rules_by_product_id, rule, rule.product_id[0])
	else if (rule.type == 'category')
		update_rules(self.loyalty.rules_by_category_id, rule, self.db.get_category_by_id(rule.category_id[0]));
})
```

Btw, why `var category = self.db.get_category_by_id(rule.category_id[0]); ` ?
Is `category.id != rule.category_id[0]` ?

geluboros:
Yes it is, i am not sure why a new variable was created.
This is how the code was in saas6, i adapted it to v10 but no refactorization was done.

geluboros:
Any prevision on merging this PR ? All the requested changes were done.

legalsylvain:
@hpar, is it ok for you ? regards.

hparfr:
👍

nabilelkadiri:
Hi,

There seem to be an issue with this module when you work offline on the POS, then try to sync orders once POS is online, when sync fails/retries, the calculated points are duplicated, customer gets to have the double of the earned points.

This has been tested on Odoo CE V12, self hosted.

Any fix please?

PR #186, opened 2017-07-11
[IMP] Ported pos_pricelist to v9. Based on work from x620. Added calc…

wpichler:
…ulcation based on standard_price. Fixed bug in simulate_price. Removed old addProduct function - not needed anymore


pedrobaeza:
You must keep commit attribution (you can squash all @x620 commits tough).

pos_pricelist/demo/pos_pricelist_demo.yml
```diff
@@ -39,12 +39,12 @@
     uom_id: product.product_uom_unit
     uom_po_id: product.product_uom_unit
     available_in_pos: True
-    seller_ids:
-      - delay: 1
-        name: base.res_partner_3
-        pricelist_ids:
-          - min_quantity : 2.0
-            price : 80
+#    seller_ids:
```
legalsylvain: please remove useless demo data.

pos_pricelist/models/point_of_sale.py
```diff
@@ -42,16 +42,18 @@ class PosOrderLine(models.Model):
     @api.multi
     def _compute_taxes(self):
         res = {
-            'total': 0,
+            # 'total': 0,
```
legalsylvain: same here.

pos_pricelist/models/pos_pricelist.py
```diff
@@ -24,5 +24,6 @@ class PosPriceListConfig(models.Model):
```
legalsylvain: could you rename into classical name. 
```class PosConfig```

File: pos_pricelist/models/pos_pricelist.py [hunk 4]
```diff
@@ -24,5 +24,6 @@ class PosPriceListConfig(models.Model):
 
     display_price_with_taxes = fields.Boolean(
         string='Price With Taxes',
-        help="Display Prices with taxes on POS"
+        help="Display Prices with taxes on POS",
+        related="iface_tax_included"
```
legalsylvain: should be better to detete display_price_with_taxes, and use iface_tax_included in the JS code, don't you think ?

pos_pricelist/static/src/js/models.js
```diff
@@ -15,29 +15,62 @@
...
+	var screens = require('point_of_sale.screens');
+	var PosDB = require('pos_pricelist.DB');
+	var core = require('web.core');
+	var utils = require('web.utils');
 
-    /**
+    var round_pr = utils.round_precision;
+    var round_di = utils.round_decimals;
+
+	PosDB = PosDB.extend({
```
legalsylvain, line 30: I don't understand this changes. you moved PosDB definition from db.js to models.js ?

legalsylvain (changes requested):
Hi @wpichler. thanks for the work. 
done a quick review for the time being. Not tested.
some remarks inline.

kind regards;
CC : @AdilHoumadi, @antespi, @StefanRijnhart, @pedrobaeza, @ismaelcj, @x620

pos_pricelist/static/src/js/models.js
wpichler: Yep - because else we would get a recursive require

pos_pricelist/demo/pos_pricelist_demo.yml
wpichler: ok

pos_pricelist/models/point_of_sale.py
wpichler: ok

pos_pricelist/models/pos_pricelist.py
wpichler: removed complete - because with the next changes it is not needed any more

pos_pricelist/models/pos_pricelist.py
wpichler: done

pos_pricelist/static/src/js/models.js
llacroix: You should use include instead of extend.

pedrobaeza:
Is this going to be continued? We have backported v11 functionality to v10. Maybe it's better to backport that one to v9 instead of continuing with this.

rafaelbn:
Hi @wpichler thank! for this work! Please you have duplicated this PR with https://github.com/OCA/pos/pull/267 could you close one and the continue in the other? Thank you!!! :smiley:

PR #191, opened 2017-07-20
[ADD] Make pos session available when selecting pickings

StefanRijnhart:
This way, in my customizations I can select pickings that come from the source location that is associated with the POS.

### DIFF
--- a/pos_picking_load/__openerp__.py
+++ b/pos_picking_load/__openerp__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'POS Picking Load',
-    'version': '8.0.1.0.0',
+    'version': '8.0.1.1.0',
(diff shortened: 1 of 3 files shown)


legalsylvain (approved):
Thanks. I propose fast merge for such trivial PR, If you're agree.

StefanRijnhart:
Yes please ;-)

pedrobaeza (changes requested):
Change module version number. You're changing signature of the methods, but I don't think this is going to be a huge problem.

StefanRijnhart:
@pedrobaeza version number updated. I don't think changing the signature is a big deal. Module was merged only nine days ago.

pedrobaeza (approved):

pedrobaeza:
Let's merge this for avoiding an "invalid" signature in stable.

PR #192, opened 2017-07-20
[IMP] API consistency

StefanRijnhart:
By wrapping the picking id in a list, the picking is instanciated on the server side when the api.multi decorator is applied to the method. The wrapping method that browses the picking is not needed anymore in that case.

### DIFF
--- a/pos_picking_load/models/stock_picking.py
+++ b/pos_picking_load/models/stock_picking.py
@@ -43,10 +43,11 @@ class StockPicking(models.Model):
         fields = self._prepare_fields_for_pos_list()
         return self.search_read(condition, fields, limit=10)
 
-    @api.model
-    def _prepare_pos_order(self, picking):
+    @api.multi
+    def load_picking_for_pos(self):
+        self.ensure_one()
         pickinglines = []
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Thanks again ! usefull refactoring.
Just a remark, what about to keep better a function that have 'prepare' in the name, as it's really a prepare function. (like ```def prepare_pos_order``` 
(not a blocking point)

hparfr (approved):

legalsylvain:
Hi @StefanRijnhart. this PR need rebase.

regards;

StefanRijnhart:
@legalsylvain thanks for noting, done!

legalsylvain:
thanks ! and what's your point of view about the name of the function ? (previous comment)

StefanRijnhart:
I don't see what is wrong with the current name load_picking_pos, as it is a method that allows the POS to load a picking.

legalsylvain:
Really not a problem. Just a matter of tastes and colors. sorry for the delay.
merging.
thanks for this improvment.

PR #195, opened 2017-07-21
[FIX] hw_telium_payment_terminal : pycountry now uses alpha_3 instead of letter

cvinh:
In pycountry >= 16.x ISO4217 now uses `alpha_3` instead of `letter`


Ousret:
You could use pyTeliumManager to improve compatibility and support of mobile payment device using Telium.

legalsylvain:
hw_telium_payment_terminal has been removed from 10.0 branch. (as all the hw_ modules) because posbox is delivered only with a 8.0 version for the time being.

so closing.

PR #199, opened 2017-08-10
[8.0][FIX] pos_pricelist : allow to use taxes type 'all'

legalsylvain:
Trivial fix. allow to use taxes type_tax_use == 'all' (and not only taxes with type_tax_use = 'sale'
CC : @rafaelbn, @StefanRijnhart, @ismaelcj, @pedrobaeza


StefanRijnhart (approved):

andhit-r (approved):
:+1:

Issue #200, opened 2017-08-17
[10.0]No more foreign keys in all tables when using this repoo


flotho:
Hi everyone,

This one seems really weird. First I suspected that Odoo CE has removed FK in fresh databases : https://github.com/odoo/odoo/issues/18880.
Yet starting with a fresh install with no specific addons in path make the things going well. Then repo by repo I added the dependencies and try to create new DB.
At last, this repo OCA/pos seeems to cause a creation with no Foreign keys.
I've updated the latest OCA/pos check all the depenecies, requirements etc, try many times to create with or without the dependencies. The result is the same, using OCA/pos cause no FK in the db!
Here are the logs, nothing special for me,any help would be appreciated.
[odoo.log.txt](https://github.com/OCA/pos/files/1232671/odoo.log.txt)

flotho:
Hi,
I made additionnal tests :
* OCA/pos : 10.0 ddfab70 OCA Transbot updated translations from Transifex
* odoo/odoo : 10.0 89cade3bffc [FIX] pos_restaurant: wrap product names on OrderChangeReceipt

Same result, all the FK disappearing from tables, Does anyone reproduce this ?

pedrobaeza:
Which module from OCA/pos? Check in runbot if it happens the same.

flotho:
I'm trying each module one by one to identify

flotho:
Well it looks like that pos_remove_category caused the problem. I'll recheck

flotho:
Well, I confirm that having the module pos_remove_category cause the problem. I made those tests :
* Creating db with adding module from oca/pos one by one, retsart odoo between each addition
** => everything work fine except when pos_remove_category is present in the addons path
* Creating db with all the module except pos_remove_category
** => Everything Ok
* creating db with only pos_remove_categ => make FK disappear

i'm suspecting the ir.module.module dependency so that I tried to comment the init part.

There's something I can't log/ identify more without your support. Thanks in advance.
regards

flotho:
@pedrobaeza it's not really easy for me to try on runbot because no error make the creation failed. Sadly there also nothing special in the logs. The db is created but the table structure haven't got anymore FK. I discovered that when a product.template has been deleted and its associated product.product still living its own ....
not easy to debug this one

legalsylvain:
Thanks @flotho for reporting this problem.

This line seems to create your bug. (not tested)
https://github.com/OCA/pos/blob/10.0/pos_remove_pos_category/models/product.py#L89

[...]

PR #201, opened 2017-08-21
[REM][8.0] remove pos_remove_pos_category Fix : #200

legalsylvain:
Fix #200.

PR #205, opened 2017-08-23
Revert "[REM][8.0] remove pos_remove_pos_category Fix : #200"

legalsylvain:
Reverts OCA/pos#201

See : https://github.com/OCA/pos/issues/200#issuecomment-324257572

CC : @StefanRijnhart, @pedrobaeza

### DIFF
--- a/pos_remove_pos_category/README.rst
+++ b/pos_remove_pos_category/README.rst
@@ -1,0 +1,62 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+POS Remove POS Category
+=======================
+
+This module was written to replace POS categories by product categories
+in the point of sale.
+
+Important notes
+---------------
+- When the module is installed the link beetween products and POS categories
+  is **overwritten** by a link beetween products and product categories
+  (the link is the field pos_categ_id in the table product_template)
+- When the module is uninstalled the link beetween products and POS categories
+  is restored in an **empty** state (NULL values)
+
+Installation
+============
+
+This module depends on the `point_of_sale` Odoo official module.
+
+Configuration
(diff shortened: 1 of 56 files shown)


hparfr (approved):

pedrobaeza:
Are you sure this is correctly working? Isn't the new approach better to be implemented also here?

bealdav (approved):

bealdav:
New approach is for v10 as i understood ?

StefanRijnhart:
This version is not affected by the bug, but @Cedric-Pigeon's solution is of course much more elegant. It can also be applied here.

bealdav:
Ok @StefanRijnhart for the precision.

We use this module with this code since almost 3 years  (150 000 sales by year) and we'll prepare migration to v10.

I'm happy to see @Cedric-Pigeon elegant solution. On our side we think we must invest in v10 in next weeks. v8 is an old version that must be kept working.

But for elegance, maybe we could think v10 or v11, or in another PR. Don't you think ?

pedrobaeza:
New approach will avoid other possible side effects, but if no one wants to implement it and this working fine, we revert it as is.

bealdav:
@legalsylvain 

error here is:

Installing Odoo https://github.com/odoo/odoo/archive/8.0.tar.gz
https://codeload.github.com/odoo/odoo/tar.gz/8.0:
2017-08-23 12:30:31 ERROR 500: Internal Server Error.

Maybe you can merge, as 2 approve and continuous-integration/travis-ci/pr green, or reload test

Thanks

pedrobaeza:
Rebuilding

StefanRijnhart (approved):
Fine by me to keep this 'as is'.

PR #211, opened 2017-09-08
[add] pos_default_payment_method

simahawk:


pos_default_payment_method/models/pos_config.py
```diff
@@ -0,0 +1,18 @@
...
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    default_payment_method_id = fields.Many2one(
+        comodel_name='account.journal',
+        string='Default payment method',
+        domain=[('journal_user', '=', True),
+                ('type', 'in', ['bank', 'cash'])],
+    )
```
legalsylvain, line 18: please set a constraint on journals pos config, to avoid bad settings.

pos_default_payment_method/static/src/js/pos_default_payment_method.js
```diff
@@ -0,0 +1,25 @@
...
+  // add `default_payment_method_id` to loaded pos config's fields
+  pos_models.load_fields("pos.config", "default_payment_method_id");
+
+  pos_screens.PaymentScreenWidget.include({
+    show: function(){
+      this._super();
+      // activate default payment method if any
+      if (this.pos.config.default_payment_method_id) {
+        this.click_paymentmethods(this.pos.config.default_payment_method_id[0]);
+      }
```
legalsylvain, line 21: Hi @simahawk. I don't understand the use case. The payment screen is displayed when you click on a journal. so why do you want to add a default payment. If I understand correctly, when user click on "Check" it will add check line and liquidity line (if liquidity is the the default payment method).

simahawk: the use case is pretty simple:
* you have several payment methods
* most of the time, depending on the POS station, you pay with the same method
* w/out this module you have to click on the payment method to enable it

It's just a way to speed up payment process when you already know how you want to pay but still allow to change the method on demand.

This is what you get w/out the module when you land on the payment page:
screenshot from 2017-09-08 13-08-02
with the module instead is already activated:
screenshot from 2017-09-08 13-08-56

pos_default_payment_method/models/pos_config.py
simahawk: https://github.com/OCA/pos/pull/211/commits/62a337ac2840aa6edee9021cb04aa61618f2a5e9 is this what you intended?

pos_default_payment_method/static/src/js/pos_default_payment_method.js
legalsylvain: Ow thanks ! I didn't seen that was for 10.0 version.

legalsylvain (approved):
:+1: Code review. No test.

- Thanks for the changes.
- Thanks for having improved description and added screenshots.

didierdonze (approved):
Functional test on runbot instance: ok

pos_default_payment_method/views/pos_config_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_pos_config_form" model="ir.ui.view">
+        <field name="name">pos.config.form.view default payment method</field>
+        <field name="model">pos.config</field>
+        <field name="inherit_id" ref="point_of_sale.view_pos_config_form"/>
+        <field name="arch" type="xml">
+            <field name="journal_ids" position="after">
+                <field name="default_payment_method_id" />
```
benwillig: Maybe put this field inside a ```group``` tag. Actually we don't see any label so it's really confusing.

benwillig (changes requested):
Functionnal test and code review :+1:

benwillig (approved):

simahawk:
@legalsylvain I missed this merge :)  Thanks \o/
Just one little thing: next time squash (or ask me/whoever to do it) to drop `fixup!` commits ;)

PR #214, opened 2017-09-29
[10.0] Pay invoice on PoS session

etobella:
This module allows to assign customer and provider invoices in point of sale sessions in order to pay them. Payments are reconciled automatically with the invoice.

It is useful when providers are paid with cash from the PoS. Also when you try to receive payment from an existing customer invoice.

PR #217, opened 2017-10-09
[11.0]][mig][account_cash_invoice]

JordiBForgeFlow:
Migration to v11. 

Account Cash Invoice
====================

This modules allows to pay an existing Supplier Invoice / Customer Refund, or
to collect payment for an existing Customer Invoice from a Cash Statement.

Usage
=====

#.  Go to *Settings* and activate the developer mode.
#.  Go to *Settings / Users & Companies / Users* and set the flag
    'Show Full Accounting Features'.
#.  Go to *Invoicing / Dashboard* and create and/or open an existing
    Cash Statement from a Cash Journal.
#.  Press the button **Pay Invoice** to pay a Supplier Invoice or a Customer
    Refund. You will need to select the expected Journal
#.  Select **Collect Payment from Invoice** in to receive a payment of an
    existing Customer Invoice or a Supplier Refund.
#.  Press **Validate** on the statement. The payment will then be reconciled
    with the invoice.


JordiBForgeFlow:
cc @rgarnau @mreficent @etobella

etobella (approved):
LGTM

HviorForgeFlow:
LGTM :+1:

HviorForgeFlow (approved):

legalsylvain (changes requested):
Hi @jbeficent,

Some questions : 
In the readme, the name is PoS Pay invoice and in the manifest, it' is account cash invoice. 
I think the second one is the good one, don't you think ? 

Also I don't understand the description, as this module depend on account (and not Point of sale)
This modules allows to pay an existing Supplier Invoice / Customer Refund, or
to collect payment for an existing Customer Invoice, using the PoS
backend session.

General question : why this module is in pos repository, and not better in account-invoicing repository.

Thanks.

account_cash_invoice/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'name': 'Account cash invoice',
+    'version': '11.0.1.0.0',
+    'category': 'Account',
+    'author': "Creu Blanca,"
+              "Odoo Community Association (OCA)",
+    'website': 'https://github.com/OCA/pos',
+    'summary': 'Pay and receive invoices from bank statements',
+    "license": "LGPL-3",
+    'depends': [
+        "account",
```
rgarnau: I had to install account_invoicing in order to use this module. Maybe it should depend on this instead of account?

rgarnau (commented):
Code and functional review :+1: (Just one comment)

JordiBForgeFlow:
@legalsylvain You're right. I have changed the module description in the readme. As for the repo it should exist in, we decided to use POS because this module is very much used in the context of POS. We think that it is rarely going to be used as standalone.

JordiBForgeFlow:
@legalsylvain Having both modules in the same repo will make it easier for people to test, IMHO

account_cash_invoice/__manifest__.py
JordiBForgeFlow: I agree. I have made the change. Thanks

legalsylvain:
Hi @jbeficent, thanks for your answer. Well, the description is pretty clear it is an accounting module.
> This modules allows to pay an existing Supplier Invoice / Customer Refund, or
to collect payment for an existing Customer Invoice from a Cash Statement.

So nothing to do with point of sale. For me it should be in some repo like https://github.com/OCA/account-payment (or maybe account-invoicing). 

I understand that it is used by a pos module, but it has nothing to do with pos. Nobody will try to look into PoS repository if he's searching such feature. 

> @legalsylvain Having both modules in the same repo will make it easier for people to test, IMHO.

we can test via runbot : in the second PR it is possible to reference the PR in accounting repository, in the requirements file. 

:-1: about the repository. But :+1: about this must have feature !

rgarnau (approved):

legalsylvain:
I so close this PR for the time being, as it is pretty clear that is not the good place for it. Please keep me in touch when you do the PR against the good one. i'll review it quickly.

Sorry for the extra work. I was on Hollydays during #214 and so I didn't have the possibility to alert about that point. 

kind regards.

PR #220, opened 2017-10-21
[MIG] pos_pricelist: Migration to 10.0

spacemonkeythe:


js-landoo:
Hi, @spacemonkeythe 
We are interested in this migration. 
We can test and code. I see most of the work is done, but it seems some features are not finished or enough tested.
If you agree we can discuss about features to be ported and what's the best way to do so.

I'll wait for your news. Many thanks for your efforts.

yelizariev:
cc @gabbasovdinar

pos_pricelist/__manifest__.py
```diff
@@ -18,23 +18,21 @@
...
 {
     'name': 'POS Pricelist',
-    'version': '8.0.1.2.0',
+    'version': '10.0.1.0.0',
     'category': 'Point Of Sale',
     'sequence': 1,
     'author': "Adil Houmadi @Taktik,Odoo Community Association (OCA)",
     'summary': 'Pricelist for Point of sale',
     'depends': [
+        'sale',
```
GabbasovDinar, line 27: why do you add the 'sale' to depends?

spacemonkeythe:
Hi @js-landoo 
Thanks for your response. I'm also interested to continue contribution on this matter, but I was (and still am) a little bit in a tight schedule. What's your suggestion, how should we proceed?

Kind regards

js-landoo:
Hi.
First.
We needed a fast solution to be able to use customer pricelist in POS v10. After some tests to this module we concluded it doesn't work (at least in our use case: formula type pricelists and each POS with a different pricelist based on standard one). And we didn't see any option to have it working in two weeks. That's why we finally bought a module in apps marketplace.

Second. 
During this time we have seen odoo v11 has this feature in community version, so our interest to develop such module for v10, has fallen down. In the future we will move to v11, so requirement will be solved with standard odoo.

Best regards.

pedrobaeza:
I close this one as there's a working backport from v11 already done in #247

PR #222, opened 2017-10-23
[8.0] hw_telium_payment_terminal: Update to newer pycountry API

alexis-via:


legalsylvain:
Hi @alexis-via. Thanks for maintaining this module.
If I understand correctly, pycountry API changed. Don't you think it could be good for people that use that module to : 
- bump the version of this module
- add a comment in the readme like. this module requires pycountry >= 17.x.y ? (I don't know another place where to put this important information).

otherwise, minor travis growls.

regards.

cvinh:
Hello I did the same PR here : https://github.com/OCA/pos/pull/195

alexis-via:
@invitu I didn't see your PR, sorry.
For POSbox module (hw_* modules), as the POSbox image runs on v8, we should only maintain the modules in the v8 branch. As v8 is no longer maintained, Odoo SA will probably update the POSbox image to a newer version, but it's not done yet.
I use pywebdriver, so I don't use this module any more. I still use the test-script though...

bealdav (approved):

bealdav:
@invitu  @legalsylvain I think you can approve now, for a quick merge. Thanks

PR #225, opened 2017-11-13
[ADD] new module pos_to_weight_by_product_uom

quentinDupont:
### Motivation
At the moment, when we create a new product and choose the unity of measure **kg**, 
we have to tick the checkbox 'To Weigh With Scale' in the Point of Sale part.

We want to **speed the process** of creating a new product, avoiding filling manually this parameter.

### Brief presentation 

This module extends the unities of measure (UOM) by adding a field **To weigh** in UOM categories and UOM, linked with the product.template field "to_weight".


hparfr (approved):

legalsylvain (approved):
:+1: !

Issue #112, opened 2016-08-04
[8.0] use of related field in rule


cubells:
Is It possible to use a related field for a rule?

I've got an error:
`ValueError: Invalid field 'company_id' in leaf "<osv.ExtendedLeaf: ('company_id', '=', False) on account_fiscal_position_tax (ctx: )>"`
due to this line:
https://github.com/OCA/pos/blob/8.0/pos_pricelist/security/account_fiscal_position_security.xml#L9

Is it necessary store=True here?:
https://github.com/OCA/pos/blob/8.0/pos_pricelist/models/account_fiscal_position.py#L11

pedrobaeza:
I haven't experimented any problem with that (I was the original author of it in https://github.com/OCA/pos/pull/85)

cubells:
Error raised after update all code from server and try to do a sale order. I'll test again.

legalsylvain:
hi @cubells, is this bug confirmed ?

cubells:
@legalsylvain : Yes, It is.

Tested and retested. If related field is not defined as store=True, then it can't be searched.

And then, this rule:
https://github.com/OCA/pos/blob/8.0/pos_pricelist/security/account_fiscal_position_security.xml#L9

throw a 

`ValueError: Invalid field 'company_id' in leaf "<osv.ExtendedLeaf: ('company_id', '=', False) on account_fiscal_position_tax (ctx: )>"`

Solution is this put store=True in this line:
https://github.com/OCA/pos/blob/8.0/pos_pricelist/models/account_fiscal_position.py#L11

Thanks.

StefanRijnhart:
@cubells How did you test this? I could not reproduce it. See my comment on Sylvain's fix, #229

legalsylvain:
Hi @cubells. I just retested this bug and did not reproduced it.
(Installed Odoo (OCB) + pos_pricelist) and try to access to the model account_fiscal_position_tax with the demo user. 

Maybe do you use an old version of Odoo or there is a problem on your instance. 
If the bug persist, could you point more precisely. (for exemple, with a demo on an instance on runbot, or writing a test in pos_pricelist) ?

Closing in the meantime. Feel free to reopen if you have new elements.

regards.

PR #229, opened 2017-12-05
[FIX][8.0] pos_pricelist rule error

legalsylvain:
Set account.fiscal.position.tax company_id to store = True to avoid rule error.

FIX : https://github.com/OCA/pos/issues/112

CC : @pedrobaeza, @StefanRijnhart, @cubells

### DIFF
--- a/pos_pricelist/models/account_fiscal_position.py
+++ b/pos_pricelist/models/account_fiscal_position.py
@@ -9,4 +9,5 @@ class AccountFiscalPositionTax(models.Model):
     _inherit = "account.fiscal.position.tax"
 
     company_id = fields.Many2one(
-        related="position_id.company_id", string="Company")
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
I don't have this problem
```
>>> self.env['account.fiscal.position.tax'].search([('company_id', '=', False)])
account.fiscal.position.tax()
>>> self.pool['account.fiscal.position.tax'].search(self.env.cr, self.env.uid, [('company_id', '=', False)])
[]
```
So I'd like to see a test that reproduces the problem.

legalsylvain:
OK, I will investigate more. Sorry for the noice.

regards.

legalsylvain:
Just tested. Impossible to reproduce this bug. Close this PR. Sorry for the inconvenience.

PR #230, opened 2017-12-05
[9.0] Remove hw_customer_display and hw_telium_payment_terminal

alexis-via:
POSbox image runs Odoo v8, so we should maintain hw_* modules only in the 8.0 branch for the moment.


legalsylvain (approved):
Excellent suggestion. Thanks !

hparfr (approved):

PR #233, opened 2017-12-28
[IMP] custom BillTicket

ovnicraft:
Module to allow customize BillTicket in Restaurant

### DIFF
--- a/pos_custombill/__manifest__.py
+++ b/pos_custombill/__manifest__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+{
+    'name': "Custom Bill Ticket",
+    'summary': """
+    Custom Bill Ticket
+    """,
+    'description': """
+    Custom Bill Ticket
+    """,
+    'author': "Cristian Salamea",
+    'website': "http://www.prisehub.com",
+    'category': 'POS',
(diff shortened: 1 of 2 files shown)


legalsylvain (changes requested):
Hi @ovnicraft. Thanks for your contribution.
About the code, there are some unrespected OCA conventions. (There are no OCA in the author, no manifest, etc...) And you does'nt seem to overload the xml bill ticket. And instead of, you seems to rewrite all the code, that is not the good way for a community module.
More generally, I'm not sure such module should be under OCA umbrella. (I did'nt understood exactly what the module changes), but it seems that it is a custom modification, not a generic one.
Thanks for your comments. 
regards.

ovnicraft:
@legalsylvain hi !  am resolving comments, about rewrite  XML, yes i c&p code, so i consider this module open the road to anyone wants to change BillTicket, no explict changes.
I can add this comments on README

legalsylvain:
Well. In the current state, this module can not be accepted in the OCA, for the reasons explained before. You can move it into a custom repository et make a reference in odoo appstore. So people will have the possibility to find it in appstore and on odoo-code-search.
I so close this PR.
regards.

PR #236, opened 2018-01-05
[FIX] remove unrelated code to pos_pricelist and fix #128

### LINKED ISSUE #128: Previous order remains in POS after ticket is printed
module: pos_pricelist
version: 8.0

**Steps to reproduce**
1) Make a sale, pay and print ticket
2) When button "Next Order" becomes available, do not click it
3) Directly scan (with a barcode reader) a barcode

**Current behavior**
The POS goes back to sale screen with previous items on it with the new one added

**Expected behavior**
Without the module activated, the previous items are cleared and only the new one is on the ticket

legalsylvain:
Hi all,

I was trying to investigate on the bug #128, and for the time being, the bug occurs on a fresh database with pos_pricelist installed.
the reason is the change with an overwrite of the function addProduct (without calling super)

Looking more deeply, I just realized that this function has nothing to do with pos_pricelist and allow to merge lines, even if it's not the last line.
this feature can be great, but has nothing to do with this module.

I so removed the full function : 
- it restore the default merge function;
- it fixes #128.

@imaralux : can you test ?
@AdilHoumadi, could you take a look ?

CC : @StefanRijnhart, @pablocm-aserti, @antespi

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -74,60 +74,6 @@ function pos_pricelist_models(instance, module) {
     });
 
     /**
-     * Extend the order
-     */
-    module.Order = module.Order.extend({
-        /**
-         * override this method to merge lines
-         * TODO : Need some refactoring in the standard POS to Do it better
-         * TODO : from line 73 till 85, we need to move this to another method
-         * @param product
-         * @param options
-         */
-        addProduct: function (product, options) {
-            options = options || {};
-            var attr = JSON.parse(JSON.stringify(product));
-            attr.pos = this.pos;
-            attr.order = this;
-            var line = new module.Orderline({}, {
-                pos: this.pos,
-                order: this,
-                product: product
-            });
-            var self = this;
-            var found = false;
-
-            if (options.quantity !== undefined) {
-                line.set_quantity(options.quantity);
-            }
-            if (options.price !== undefined) {
-                line.set_unit_price(options.price);
-            }
-            if (options.discount !== undefined) {
-                line.set_discount(options.discount);
-            }
-
-            var orderlines = [];
-            if (self.get('orderLines').models !== undefined) {
-                orderlines = self.get('orderLines').models;
-            }
-            for (var i = 0; i < orderlines.length; i++) {
-                var _line = orderlines[i];
-                if (_line && _line.can_be_merged_with(line) &&
-                    options.merge !== false) {
-                    _line.merge(line);
-                    found = true;
-                    break;
-                }
-            }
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Change module version number

legalsylvain:
Thanks for the review @pedrobaeza. Done.

pos_pricelist/static/src/js/models.js
```diff
@@ -73,60 +73,6 @@ function pos_pricelist_models(instance, module) {
...
-                product: product
-            });
-            var self = this;
-            var found = false;
-
-            if (options.quantity !== undefined) {
-                line.set_quantity(options.quantity);
-            }
-            if (options.price !== undefined) {
-                line.set_unit_price(options.price);
```
pedrobaeza, line 104: Maybe this is for setting the correct price and it's needed?

pos_pricelist/static/src/js/models.js
```diff
@@ -73,60 +73,6 @@ function pos_pricelist_models(instance, module) {
...
-            }
-            if (options.discount !== undefined) {
-                line.set_discount(options.discount);
-            }
-
-            var orderlines = [];
-            if (self.get('orderLines').models !== undefined) {
-                orderlines = self.get('orderLines').models;
-            }
-            for (var i = 0; i < orderlines.length; i++) {
```
StefanRijnhart, line 114: I think it's this hidden feature: upstream Odoo only considers the last line to merge with, this change considers all lines on the order. It should probably go into OCB instead.

StefanRijnhart (approved):

pos_pricelist/static/src/js/models.js
legalsylvain: No it's not. you'll find the same line in the Odoo code. 
see original overwritten function : 
https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/static/src/js/models.js#L1066

and

the comment of the function in the OCA module talks about lines merging.
https://github.com/OCA/pos/pull/236/files/c879f74940fbd11ed3f13d75c820fd500980d3c4#diff-747f62729cafe771e948f5cfeb9aff46L81

regards.

pedrobaeza (approved):
OK then

pos_pricelist/static/src/js/models.js
legalsylvain: yep, it's not the place. (and overwrite function is allways a good way to generate bugs).

Ideally, it should be an option. In some cases, users want to merge all lines, sometime no lines. (but it's another story).

thanks for the review @StefanRijnhart.

PR #239, opened 2018-01-20
[ADD] new module pos_timeout

legalsylvain:
This module allow to change the default value of the timeout of the point of sale, for the creation of pos.order. 
the default value is 7.5 seconds. 

https://github.com/odoo/odoo/blob/10.0/addons/point_of_sale/static/src/js/models.js#L780

pos_config

Thanks for your review.


pedrobaeza:
Shouldn't be pos_order_timeout?

legalsylvain:
Hi @pedrobaeza. 
Thanks for your review.
Well, I checked the code of the PoS and this is a only timeout that can be configured. (the others call use standards timeout).
But I can change the name, if it's a blocking point for you.

Regards.

pedrobaeza:
No special problem, so what you prefer.

flotho:
@pedrobaeza , we've tested it with @gaelTorrecillas and it works as expected.
Is it possible to merge?

mourad-ehm (approved):
Code review, not tested

legalsylvain:
thanks for your reviews. Merged.

PR #244, opened 2018-01-24
10.0 mig pos order load

cvinh:


cvinh:
@legalsylvain ok, the PR seems to be correct here

cvinh:
@oca-clabot I just changed my username on Github. Now it's cyril@invitu.com.
I just resent the ICLA and ECLA to info@odoo-community.org

pos_order_load/__manifest__.py
```diff
@@ -0,0 +1,39 @@
...
+{
+    'name': 'POS Order Load and Save',
+    'version': '10.0.1.0.0',
+    'author': 'Akretion,GRAP,Odoo Community Association (OCA)',
+    'category': 'Point Of Sale',
+    'license': 'AGPL-3',
+    'depends': [
+        'point_of_sale',
+    ],
+    'website': 'https://www.akretion.com',
```
chienandalu, line 32: https://www.github.com/OCA/pos

chienandalu (commented):
Amend your migration commit with your new github email so it's linked to your user :slightly_smiling_face:

pos_order_load/models/pos_order.py
```diff
@@ -0,0 +1,97 @@
...
+        return {
+            'id': order.id,
+            'name': order.pos_reference,
+            'partner_id': order.partner_id.id,
+            'orderlines': orderlines,
+        }
+
+    @api.model
+    def _create_draft_order_from_ui(self, orders):
+        for order_tmp in orders:
```
chienandalu, line 86: Not blocking (just curious): Wouldn't it be better to save these draft orders in a temporary session so we can close the current session and be able to load them into another?

legalsylvain: I had this need before, and developped another module for that purpose, to be able to keep a draft order, when closing the session. (no need to create a temporary session I think)

https://github.com/grap/odoo-addons-misc/blob/7.0/pos_keep_draft_orders/__openerp__.py

I think it is out of the scope of this module.

Note : I did'nt port this module because it is now forbidden by law (in France) to have draft orders that could be deleted.
If there are customer that take products but don't pay, I now use for that purpose pos_order_to_sale_order to create delivered pickings. https://github.com/OCA/pos/tree/8.0/pos_order_to_sale_order

regards.

chienandalu: Thanks @legalsylvain :slightly_smiling_face:

pos_order_load/__manifest__.py
cvinh: I think it's solved now

PR #247, opened 2018-02-15
[MIG] pos_pricelist: Migrate to v10 by backporting from v11

yajo:
This module included previously support for taxes and pricelists.

Taxes are now included in upstream Odoo v10, so that feature is removed.

The pricelist feature, however, is supported in upstream Odoo v11. Thus, to ease the migration path, that functionality has been scalpel-backported from there.

This also means a great diff between previous OCA-only code and current Odoo-backport-only code, but with the advantage that the end user gets exactly the behavior that he would have when updating to v11.

Since most of the code has been copied from Odoo, the addon now has to become LGPL, and Odoo has been added as an author.

@Tecnativa


yajo:
Not sure on how to fix the linters... About the tests, it's a lot of time, and the addon will only live for v10, so not much needed IMHO.

pos_pricelist/README.rst
```diff
@@ -1,101 +1,56 @@
...
-- Tax details aggregated by tax at order level
-
-2. Ticket
-
-- Tax details table added at end of printed ticket
-
+#. Go to *Point of Sale > Dashboard* and open the POS session you configured.
+#. Use the new pricelist button to change it on the fly.
+#. When a new order is created, it always has the default pricelist.
+#. When you change to a customer that has a different pricelist, the current
```
chienandalu, line 38: This won't happen if `pos_backend_partner` is installed. Maybe it should be mentioned in the konwn issues section.

pos_pricelist/README.rst
```diff
@@ -1,101 +1,56 @@
...
-4. When we mouseover the price tag, a tooltip is shown to indicate the computation depending on the quantity like this output :
-1x -> 100 €
-3x -> 70 €
-5x -> 50 €
-
-
-Implemented features at backend
--------------------------------
-
-1. Tax details
```
chienandalu, line 72: Not migrated?

pos_pricelist/README.rst
```diff
@@ -1,101 +1,56 @@
...
-
-- Tax details per order line
-- Tax details aggregated by tax at order level
-
-2. Ticket
```
chienandalu, line 77: Not migrated?

pos_pricelist/README.rst
```diff
@@ -1,101 +1,56 @@
...
-
-- Tax details table added at end of printed ticket
-
+#. Go to *Point of Sale > Dashboard* and open the POS session you configured.
+#. Use the new pricelist button to change it on the fly.
+#. When a new order is created, it always has the default pricelist.
+#. When you change to a customer that has a different pricelist, the current
+   order and the listed product prices are udpated accordingly.
```
chienandalu: There's an issue with rules based on qty: unless you reselect the pricelist, unit price is not recalculated wich brings two possible problems:

- A special qty price is not applied when for a given qty of product.
- A special qty price is not disabled when the qty is set below the rule!

pos_pricelist/README.rst
```diff
@@ -1,101 +1,56 @@
...
-2. Fiscal Position of each partner will also be present so taxes will be correctly computed
-(conforming to the fiscal position).
-
-- Implemented Rules are :
-
-1. (-1) : Rule based on other pricelist
-2. (-2) : Rule based on supplierinfo
-3. (default) : Any price type which is set on the product form
-
-3. An new option is introduced in the POS config to let the user show price with taxes in product widget.
```
chienandalu, line 59: Not migrated?

pos_pricelist/README.rst
```diff
@@ -1,101 +1,56 @@
...
-the UI is updated when we change the customer in order to adapt the prices.
-The computation take in account the pricelist and the fiscal position of the customer
-
-4. When we mouseover the price tag, a tooltip is shown to indicate the computation depending on the quantity like this output :
```
chienandalu, line 63: Not migrated?

chienandalu (changes requested):
Thanks @Yajo ! I found some functional issues:

pos_pricelist/README.rst
pedrobaeza: No, original module mixes some concepts, and this is one of them. Taxes goes apart. Any way, Odoo itself now handles taxes correctly in POS.

pos_pricelist/README.rst
pedrobaeza: v11 doesn't include this feature, and it's better to align with that. It's not the usual any way. Think that in a touchscreen, you can't "mouseover".

pos_pricelist/README.rst
pedrobaeza: s/udpated/updated

pos_pricelist/README.rst
pedrobaeza: Not needed. Current ticket now includes that.

pos_pricelist/README.rst
pedrobaeza: Not needed. Included in core

pos_pricelist/i18n/es.po
```diff
@@ -1,141 +1,179 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
-#	* pos_pricelist
+# * point_of_sale
 #
+# Translators:
```
pedrobaeza: Why this large list of translators? They haven't translated the exact terms of this module, so better to remove them.

pos_pricelist/models/pos_config.py
```diff
@@ -0,0 +1,117 @@
...
+    _inherit = 'pos.config'
+
+    # Redefine preexisting field
+    pricelist_id = fields.Many2one(
+        string="Default Pricelist",
+        help="The pricelist used if no customer is selected or if the "
+             "customer has no Sale Pricelist configured.",
+    )
+    # New fields
+    # TODO Prefill with post init hook
```
pedrobaeza: This is finally done, so remove it

pedrobaeza (changes requested):
Little comments

pos_pricelist/i18n/es.po
yajo: I copied preexisting translations from `point_of_sale` v11, since all terms changed to match those. I just filled the missing ones, so I feel I should keep attribution. 😕 

See https://github.com/odoo/odoo/blob/11.0/addons/point_of_sale/i18n/es.po

pedrobaeza: We are not attributing translation that way and it's going to be deleted any way on next Transifex push, so please remove it.

pos_pricelist/README.rst
yajo: Done

pos_pricelist/README.rst
yajo: Good catch! I was missing a piece of backport code. Fixed in 67399a0.

yajo:
All remarks attended.

chienandalu:
I thought I had reported a bug with prices based on qty rules but I can't find it. Basically they won't apply unless the pricelist is reselected.

yajo:
That should be fixed with 67399a0, which is building right now. Could you check after some minutes in runbot please?

chienandalu (approved):

chienandalu:
@Yajo Right! It's working now :slightly_smiling_face:

yajo:
@pedrobaeza could you update your review please?

pedrobaeza:
Travis is now red

yajo:
It should go :heavy_check_mark: now

pedrobaeza:
I'm afraid still not.

pedrobaeza (approved):

PR #251, opened 2018-02-28
[FIX] pos_pricelist: Use lower-version date comparison

yajo:
[The removed function was introduced in momentjs in version 2.11.0](https://momentjs.com/docs/#/query/is-same-or-before/), but [Odoo 10.0 ships 2.8.1](https://github.com/odoo/odoo/blob/b2c321fd1f86a8cb6329e75e6306a657caf17987/addons/web/static/lib/moment/moment.js#L2), so I'm using a different comparison system here that should yield the same results without exceptions.

@Tecnativa

### DIFF
--- a/pos_pricelist/__manifest__.py
+++ b/pos_pricelist/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'POS Pricelist',
-    'version': '10.0.1.0.0',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Fast-tracking as this avoids the module to work.

PR #252, opened 2018-03-01
[8.0][fix] pos_pricelist

legalsylvain:
Before the PR

1. when adding a pos.order.line in the classical UI (not tactile), pos.order.line doesn't have taxes. This is problematical when user create / update refunds.
2. when user create a pos order (in the PoS UI) with a customer that have a specific pricelist, the pricelist is not saved in the pos.order
3. It is not possible to create / update the pricelist of a customer.

This PR fixes those 3 bugs.

Regards.

### DIFF
--- a/pos_pricelist/__openerp__.py
+++ b/pos_pricelist/__openerp__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'POS Pricelist',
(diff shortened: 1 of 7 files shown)


StefanRijnhart (approved):
Thanks!

pedrobaeza (approved):

PR #254, opened 2018-03-07
[10][FIX] Protect crash at runtime if oca-decorator not installed

### LINKED ISSUE #249: @foreach NameError: name 'foreach' is not defined
https://github.com/OCA/pos/blob/d379888fef2a62777e9b248a140b17432766c3e7/pos_pricelist/models/pos_config.py#L55

this generates below error:
  File "/home/mhagag/O10Dist/OCA/pos/pos_pricelist/__init__.py", line 5, in <module>
    from . import models
  File "/home/mhagag/O10Dist/OCA/pos/pos_pricelist/models/__init__.py", line 4, in <module>
    from . import pos_config
  File "/home/mhagag/O10Dist/OCA/pos/pos_pricelist/models/pos_config.py", line 17, in <module>
  File "/home/mhagag/O10Dist/OCA/pos/pos_pricelist/models/pos_config.py", line 55, in PosConfig
    @foreach()
NameError: name 'foreach' is not defined

hparfr:
Fail silently if @foreach not loaded

Fixes #249

### DIFF
--- a/pos_pricelist/__init__.py
+++ b/pos_pricelist/__init__.py
@@ -2,8 +2,9 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from .hooks import post_init_hook
-
 try:
     from . import models
-except NameError: # pragma: no-cover
-    pass
+except ImportError:  # pragma: no-cover
+    import logging
+    _logger = logging.getLogger(__name__)
+    _logger.warn("Missing dependency", exc_info=True)
--- a/pos_pricelist/__init__.py
+++ b/pos_pricelist/__init__.py
@@ -2,4 +2,8 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from .hooks import post_init_hook
-from . import models
+
+try:
+    from . import models
+except NameError: # pragma: no-cover
+    pass
--- a/pos_pricelist/models/pos_config.py
+++ b/pos_pricelist/models/pos_config.py
@@ -4,14 +4,10 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from oca.decorators import foreach
 import logging
 
 _logger = logging.getLogger(__name__)
-
-try:
-    from oca.decorators import foreach
-except ImportError:  # pragma: no-cover
-    _logger.warn("Missing dependency", exc_info=True)
 
 
 class PosConfig(models.Model):
--- a/pos_pricelist/models/pos_config.py
+++ b/pos_pricelist/models/pos_config.py
@@ -4,7 +4,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
-from oca.decorators import foreach
+from oca.decorators import foreach  # pylint: disable=W7935
 import logging
 
 _logger = logging.getLogger(__name__)


legalsylvain:
Hi @hparfr . Thanks for taking care of that ! Just one question, it should not be better to change the try import in the file https://github.com/OCA/pos/blob/d379888fef2a62777e9b248a140b17432766c3e7/pos_pricelist/models/pos_config.py#L55 ?

like : 
```
    try:
        from oca.decorators import foreach
    except ImportError:  # pragma: no-cover
        _logger.warn("Missing dependency", exc_info=True)
        foreach = false
```

or something like that. I fear that your patch will hide other import errors. 
Just a question, reallly not an expert on that point.

hparfr:
Yes it will silent all NameErrors.
I've committed a modified version of the patch.

pedrobaeza (approved):
Well, it's a bit heavy to mask all the files of models, but it can serve.

pedrobaeza:
Any way, please check Travis

hparfr:
Travis says import not in a try catch (in config.py)

legalsylvain:
@Yajo  / @pedrobaeza. could you propose correct fix for this bug introduced by the PR : https://github.com/OCA/pos/pull/247
This repository is currently broken for all users that are pulling this branch.

A quick answer would be appreciated.

pedrobaeza:
Sorry, but we are not able to do it fast. The quick fix is to install the library. If not enough for you, please do it yourself or wait till tomorrow.

yajo (approved):

pos_pricelist/models/pos_config.py
```diff
@@ -4,15 +4,11 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from oca.decorators import foreach
```
yajo: Possibly you need to [disable the pylint message for this line](https://travis-ci.org/OCA/pos/jobs/350388592#L734) because it's handled in a higher level.

Odoo is so stupid that it imports addons when not installed without checking if there's an exception when doing so :man_facepalming: 

Another option would be to:

```python
try:
    from oca.decorators import foreach
except ImportError:
    def foreach():
        return api.one
```

But that's uglier IMHO.

pedrobaeza: There's a better solution proposed by @hbrunn in another issue opened in the library. I don't like too much to have 3 issues/PRs for the same thing.

legalsylvain:
@hbrunn was proposing the following :  (https://github.com/OCA/oca-decorators/issues/10)

```
try:
    from oca.decorators import foreach
except ImportError:
    foreach = lambda x: x
```

what do you think ?

rafaelbn:
Please @pedrobaeza @Yajo @chienandalu could we merge? All bots are :green_heart: ! :+1:

yajo:
So I think; I already approved

chienandalu (approved):

pos_pricelist/__init__.py
```diff
@@ -2,4 +2,9 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from .hooks import post_init_hook
-from . import models
+try:
+    from . import models
+except ImportError:  # pragma: no-cover
+    import logging
+    _logger = logging.getLogger(__name__)
+    _logger.warn("Missing dependency", exc_info=True)
```
yelizariev, line 10: Shall it be ``debug``?

pedrobaeza: Well, better for avoiding yellow runbot. Changing directly.

PR #255, opened 2018-03-09
[FIX] make the module working for product variants

legalsylvain:
due to bad design in odoo 8.0 (not refactored in new API in 8.0), the onchange function should be defined in both models (product.product and product.template).

### DIFF
--- a/pos_to_weight_by_product_uom/models/product_product.py
+++ b/pos_to_weight_by_product_uom/models/product_product.py
@@ -1,16 +1,17 @@
 # coding: utf-8
-# Copyright (C) 2017 - Today: GRAP (http://www.grap.coop)
-# @author: Quentin DUPONT <quentin.dupont@grap.coop>
+# Copyright (C) 2018 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 from openerp import api, models
 
 
-class ProductTemplate(models.Model):
-    _inherit = 'product.template'
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
 
     @api.multi
     def onchange_uom(self, uom_id, uom_po_id):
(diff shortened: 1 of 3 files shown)


quentinDupont (approved):

pos_to_weight_by_product_uom/models/product_product.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    @api.multi
+    def onchange_uom(self, uom_id, uom_po_id):
+        res = super(ProductProduct, self).onchange_uom(uom_id, uom_po_id)
+        if uom_id:
+            res = res and res or {}
```
pedrobaeza: It's enough with `res = res or {}`

pos_to_weight_by_product_uom/models/product_product.py
```diff
@@ -0,0 +1,21 @@
...
+            if not res.get('value', False):
```
pedrobaeza: You can use `val = res.setdefault('value', {})` and then `val['to_weight'] = ...`

pedrobaeza (changes requested):

legalsylvain:
@pedrobaeza. Thanks for your suggestion. modifications done. 
regards.

pos_to_weight_by_product_uom/models/product_template.py
```diff
@@ -12,7 +12,9 @@ class ProductTemplate(models.Model):
     def onchange_uom(self, uom_id, uom_po_id):
         res = super(ProductTemplate, self).onchange_uom(uom_id, uom_po_id)
         if uom_id:
-            if res.get('value', False):
-                res['value']['to_weight'] =\
-                    self.env['product.uom'].browse(uom_id).to_weigh
+            res = res and res or {}
```
pedrobaeza, line 15: The same applies here

rafaelbn (approved):
:+1:

pedrobaeza (approved):
Let's not block more.

PR #257, opened 2018-03-12
[ADD] pos_stock_picking_invoice_link: New Module

chienandalu:
POS Stock Picking Invoice Link
==============================

Links POS generated stock moves and pickings to its corresponding invoice
lines.

Usage
=====

* Create a POS order with stockable products and invoice option checked.
* If you open invoice form in the backend, you must see the related picking in
  the Pickings tab.

cc @Tecnativa


pedrobaeza:
Travis is still red

pos_stock_picking_invoice_link/models/pos_order.py
```diff
@@ -0,0 +1,30 @@
...
+        return res
+
+    def _action_create_invoice_line(self, line=False, invoice_id=False):
+        invoice_line = super(
+            PosOrder, self)._action_create_invoice_line(line, invoice_id)
+        if not line:
+            return invoice_line
+        move = self.env['stock.move'].search([
+            ('picking_id', '=', self.picking_id.id),
+            ('name', '=', line.name)])
```
pedrobaeza, line 25: Please use a harder link, like `product_id`. Any way, we will have the problem when the same product is put several times (for example, when using different lots). Isn't any deterministic way of linking a pos line with an stock move?

chienandalu: Well, only a move is created by each line: https://github.com/odoo/odoo/blob/10.0/addons/point_of_sale/models/pos_order.py#L690

Isn't it deterministic enough?

pedrobaeza: I'm afraid not. As said, this will find multiple moves if you add several times a product, linking all of them to all lines.

POS line 1: Product 1 > Move line 1
POS line 2: Product 2 > Move line 2
POS line 3: Product 1 > Move line 3

Resulting in:

Invoice line 1 from POS line 1, linked to Move line 1 and 3
Invoice line 2 from POS line 2, linked to Move line 2
Invoice line 3 from POS line 3, linked to Move line 1 and 3

chienandalu: I made a test (https://github.com/OCA/pos/pull/257/commits/210533ea5cc97020ea5c77315ef2df570cc94ad0) with your example to show you that doesn't happen. Every pos order line has a unique sequence (determined by an ir.sequence) and its consequent stock move is named after it. So it's very unlikely that we're getting to moves with that premise, isn't it?

pedrobaeza: I don't see in that commit the test you are mentioning, but I'm interpreting from your words that picking line text is composed by not the product name, but another thing?

chienandalu:
Argg, I didn't push it. Yes it names it after an ir.sequence: https://github.com/odoo/odoo/blob/11.0/addons/point_of_sale/models/pos_order.py#L535

pedrobaeza:
Travis is now red.

chienandalu:
Green now

pedrobaeza (approved):

luismontalba (approved):
Tested on runbot

PR #261, opened 2018-03-16
[IMP] pos_lot_selection: clone control no longer needed

chienandalu:
- [x] Needs PR https://github.com/odoo/odoo/pull/23698 to be merged, when clone control is
no longer needed for lots.

cc @Tecnativa

### DIFF
--- a/pos_lot_selection/README.rst
+++ b/pos_lot_selection/README.rst
@@ -29,8 +29,6 @@ Usage
 * There is a new select field with the lots available for that product.
 * If lot quantity can't cover the order line demand it won't be available to
   pick.
-* If there were many lots to fill the value of one of them can be propagated to
-  the rest with the new clone control on the right of the lot field.
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
--- a/pos_lot_selection/static/src/js/chrome.js
+++ b/pos_lot_selection/static/src/js/chrome.js
@@ -13,7 +13,6 @@ odoo.define("pos_lot_selection.chrome", function (require) {
             // Add events over instanced popup
             var events = {
                 "change .packlot-line-select": "lot_to_input",
-                "click .lot-clone": "clone_input",
             };
             packlotline.events = Object.assign(
                 packlotline.events, events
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Let's see if they accept the patch...

chienandalu:
Added reserved quant filter to the selector so only unreserved quants are shown.

pedrobaeza:
But this is still pending of the odoo PR, isn't it?

chienandalu:
Yes, no comment from their side yet

pedrobaeza:
Did you put OPW?

chienandalu:
I didn't but I'm opening a new one to speed it up

rafaelbn:
Hi @chienandalu as https://github.com/odoo/odoo/commit/8f39c652c65c81d87503d64eeef0da3e0048ffc9

I think you could rebase and go ahead with this review

Isn't it?

chienandalu:
@rafaelbn No, that's a commit on `11.0`

rousseldenis:
@chienandalu I confirm odoo patch is on v10.

chienandalu:
Sorry, where? I only see it on OCB

rafaelbn:
Hi @chienandalu , Odoo solved in https://github.com/odoo/odoo/commit/8f39c652c65c81d87503d64eeef0da3e0048ffc9

So please go ahead with this PR please :smile:

rafaelbn (approved):

pedrobaeza:
Not sure this has been merged. @chienandalu please confirm.

chienandalu:
I reviewed this again. That's not the patch involved, but these ones, that fix those first: https://github.com/odoo/odoo/pull/23698/commits

Odoo guys discussed here: https://github.com/odoo/odoo/pull/15735 and decided to not implement it in v10.

It was finally merged in OCB: https://github.com/OCA/OCB/pull/711

So this PR works with an OCB base but not with an Odoo base

pedrobaeza:
What happens then mixing this PR with vanilla Odoo?

chienandalu:
I'd say the functionality would just be lost and the users would go back to the standard behavior.

pedrobaeza:
Yeah, but the module would be useless then, isn't it?

chienandalu:
It'd be less usable, as there was the need to fill the lot for every unit entered (if I remember well)

pedrobaeza:
Well, then I think we shouldn't merge this. It's only something for reducing code in the module, and being an old version, better to stick to full features. We can still use this aggregating the PR in our v10 databases, so I'm closing this, but without deleting branch.

PR #263, opened 2018-03-26
[10][MIG] pos_order_to_sale_order

hparfr:
Did some refactoring due to JS Framework and for better UI.

PR #136, opened 2016-12-09
[ADD] pos_default_customer: establish a default customer in each POS

jonerikceberio:

PR #264, opened 2018-03-26
[10.0][MIG] pos order to sale order

hparfr:
replaces #263


pedrobaeza:
@chienandalu please check this one to see convergences and evaluate to adopt this.

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
-.. figure:: static/description/pos_create_picking_option_2.png
-   :width: 800 px
+   A draft sale order is a quotation. 
+   It can be changed and confirmed later from the backend.
+   No payment can be taken with draft orders.
 
-#. '**Create Delivered Picking**' (by default)
-  A new sale order will be created and confirmed. the associated picking
-  will be marked as delivered.
+#. '**Allow confirmed order**'
```
chienandalu: Doesn't match the setting name

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
+
+   A confirmed sale order without picking.
+   Useful if you don't manage stock or sell services otherwise
+   use 'delivered picking'.
+
+#. '**Allow delivered picking**'
```
chienandalu: Doesn't match the setting name

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
-  have a unique invoice at the end of the month. With that module, you can
-  create a sale order and deliver products every time to keep your stock value
-  correct, and to create a unique invoice, when you want.
+  have a unique invoice at the end of the month. With this module, you can
+  create a sale order and deliver products every time he comes 
+  to keep your stock value correct and to create a unique invoice periodically.
 
 
-Three options are available:
+Four options are available:
```
chienandalu, line 30: This could be moved to configuration section

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
 
 Configuration
 =============
 
 To configure this module, you need to:
 
 #. Go to Point Of Sale / Configuration / Point of Sale
-#. Check the box 'Create Sale Orders'
-#. Select the desired default behaviour
+#. Select the desired default behavior
```
chienandalu: Where?

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
 
-.. figure:: static/description/pos_config_form.png
-   :width: 800 px
 
 Usage
 =====
 
+In the PoS, on the payment screen, new buttons will appear according
+ to options set in the configuration.
```
chienandalu, line 76: What's the workflow?

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
 * Some prepare functions are available in the sale.order model, to overload
   the creation of the sale order.
 
-* You could be interested by another module, pos_sale_order, that completely
-  alter Point of Sale module, avoiding creating Pos Orders, and creating
-  allways Sale Orders.
-  This module is a WIP state, and is available here:
-  https://github.com/OCA/pos/pull/35
+* An additional, not displayed option, is to allow payment for sale orders.
+  It requires an additinnal module (not yet written) to handle it correctly.
```
chienandalu, line 96: Why to include this yet then?

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
+
+
 
 Known issues / Roadmap
 ======================
 
+* Demo not ported yet. It was non functionnal before migration.
 * Because of the poor design of the Odoo Point of Sale, some basic features
```
chienandalu, line 104: Do this still apply?

pos_order_to_sale_order/static/src/xml/pos_order_to_sale_order.xml
```diff
@@ -1,12 +1,26 @@
...
-            <button class="input-button">
-                <t t-esc="widget.display_text"/>
-            </button>
+    <t t-name="DeliveryButtonWidget">
+        <t t-set="isPicking" t-value="widget.isPicking()" />
+        <div t-attf-class='button #{ isPicking ? "highlight": ""}'>
+            <i t-attf-class='fa #{ isPicking ? "fa-arrow-circle-down": "fa-forward"}' />
+            Delivery 
+            <t t-if="!isPicking"> Later</t>
+            <t t-if="isPicking"> Now</t>
```
chienandalu: I think this is uneeded. Just a "Deliver now" text should be shown. If the button is active or not is going to give de clue of the expected behaviour.

chienandalu (changes requested):
Thanks for your work. Tested on runbot. Here are some notes:
- Made a test with lots and they weren't processed in the picking correctly.
- ¿Shouldn't a confirmed order be allowed to generate an invoice if set in the pos?
- Workflow is a little bit weird. I'd expect to have a control like the invoice one and different payment methods to pick from (a deferred one could be set for drafts).

pos_order_to_sale_order/README.rst
hparfr: It's more about what this module do rather than where/how to configure it.

pos_order_to_sale_order/README.rst
hparfr: If some options are selected: some buttons will appear in the pos. User may click on it or not.
If only one option is selected: no additional choices in pos.

Don't know how to state that intelligibly. Any suggestions ?

pos_order_to_sale_order/README.rst
hparfr: To be clear that this modules doesn't handle payments.

What do you suggest then ?

pos_order_to_sale_order/README.rst
hparfr: Yes more than never.

pos_order_to_sale_order/README.rst
```diff
@@ -15,74 +15,87 @@ creation from the Point of Sale.
...
+  create a sale order and deliver products every time he comes 
+  to keep your stock value correct and to create a unique invoice periodically.
 
 
-Three options are available:
+Four options are available:
 
-#. '**Create a draft Order**'
-  A new sale order in a draft mode will be created that can be changed later.
+#. '**Allow PoS order**'
```
legalsylvain, line 32: I'm not sure include this feature in this module is a good idea. I think it should be better in stand-alone module. ```pos_allow_pos_order```. 
- This module extend the functionality of Point Of Sale, not reduce it.
- and overall, this feature is needed in other cases. I think specially to point of sale that could be used in shop by customers, to weight the products and edit labels.
this feature is transveral IMO.

pos_order_to_sale_order/README.rst
legalsylvain: I think that pos_order_to_sale_order should be more near from the original implementation. An other module 
pos_order_to_paid_sale_order could inherit of this module, overloading this module (from python part), and adding a button on payment screen. All other button should be in the ticket form.

pos_order_to_sale_order/__manifest__.py
```diff
@@ -19,6 +22,9 @@
     ],
     'demo': [
         'demo/res_groups.xml',
+        # 'demo/product_template.xml',
```
legalsylvain, line 29: you can fully remove this files (and remove the warning in the readme file).

It's a error in the initial V8 module. My bad.

pos_order_to_sale_order/models/pos_config.py
```diff
@@ -1,27 +1,37 @@
...
-
-    iface_create_delivered_sale_order = fields.Boolean(
-        string='Create Delivered Sale Orders', default=True,
-        help="If checked, the cashier will have the possibility to create"
-        " a confirmed sale Order, based on the current draft PoS Order.\n"
-        " the according picking will be marked as delivered. Only invoices"
-        " process will be possible.")
+    iface_allow_pos_order = fields.Boolean(
+        string='Allow PoS Orders', default=True,
+        help="Allow PoS Orders")
```
legalsylvain: Don't think the helper is helpfull ;-)

pos_order_to_sale_order/models/pos_config.py
```diff
@@ -1,27 +1,37 @@
...
+        string='Allow Delivered Sale Orders', default=True,
+        help="Allow creation of delivered sale orders."
+        " The according picking will be marked as delivered. Only invoices"
+        " process will be possible.",
+        old_name="iface_create_delivered_sale_order")
+
+    iface_allow_order_payment = fields.Boolean(
+        # require an additional module to link payment to sale orders
+        string='Allow Sale Order Payments', default=False,
+        help="Accept payments for confirmed and delivered orders.")
```
legalsylvain, line 32: As I don't see this field used in python and JS file, I think it should be removed.

legalsylvain (changes requested):
Hi @hparfr. 

Why did you moved the button from the ticket form to the payment form ? There is an placeholder in V10, very easy to use in ticket form, to contains all the button.

More generally, I don't understand the creation of the state machine, and I see all the JS very more complex to maintain, and not in the Odoo and OCA spirit at all.
But maybe I missed something !

Thanks for this work.

Seeing the existing module, and the new feature you try to introduce, I'll be more for a modular design like this : 

* ```pos_order_to_sale_order``` : module ported from the V8 version, with buttons still in the bill form. (I see the move in the payment form less intuitive for end users, and more complex to handle in the JS part)

* ```pos_order_to_paid_sale_order``` : module that depend on ```pos_order_to_sale_order```, (and maybe on ```account_cash_invoice``` but it requires refactoring i think, not sure on that point), to allow to handle payment of sale order, handle reconcile, ...

* a module ```pos_allow_pos_order``` : independent that block, or not the the possibility to create pos orders. (new feature, interested, but not dependent on the module ```pos_order_to_sale_order``` IMO).

hparfr:
> Made a test with lots and they weren't processed in the picking correctly.

I didn't test against this module.

>¿Shouldn't a confirmed order be allowed to generate an invoice if set in the pos?

It should be delegate to the to-be-written module which will handle payment.

>  Workflow is a little bit weird. I'd expect to have a control like the invoice one and different payment methods to pick from (a deferred one could be set for drafts).

The payment is out of scope of this module. 

Thanks for your review

pos_order_to_sale_order/models/pos_config.py
hparfr: https://github.com/akretion/pos/blob/20c81a139ee61c5a2010db536275ab23ddf5102d/pos_order_to_sale_order/static/src/js/product_screen.js#L70

hparfr:
Hi,

> Why did you moved the button from the ticket form to the payment form ? There is an placeholder in V10, very easy to use in ticket form, to contains all the button.

There is not much room in the ticket. + It make more sense to take actions in the payment screen since the refactoring of the views by Odoo. For instance, the validate button is not the same screen than the ticket, this small diff implies a lot of changes.

> More generally, I don't understand the creation of the state machine, and I see all the JS very more complex to maintain, and not in the Odoo and OCA spirit at all.

All the states are confined in one single object which is queried by consumers.  Without it, the logic will be splitted across differents widgets, t-if, css:selectors. In the long run it's easier to maintain, to test, and change behavior.

This version is more complex than in the v8. Because it handle more use-cases: like remove payment lines if user choose a draft order, re-use some validation logic of the original pos...

>    pos_order_to_paid_sale_order : module that depend on pos_order_to_sale_order, (and maybe on account_cash_invoice but it requires refactoring i think, not sure on that point), to allow to handle payment of sale order, handle reconcile, ...
Agree on this one.

 >   a module pos_allow_pos_order : independent that block, or not the the possibility to create pos orders. (new feature, interested, but not dependent on the module pos_order_to_sale_order IMO).

There is no need for a module like that. Because, you don't need to block anything if you add other ways (like draft, confirmed orders and picking). The user is already making a choice.

legalsylvain:
> There is not much room in the ticket. + It make more sense to take actions in the payment screen since the refactoring of the views by Odoo. For instance, the validate button is not the same screen than the ticket, this small diff implies a lot of changes.

* I'm not agree with that statement. Since V10 (or maybe 9.0), a new placeholder is available for such button).

image

* force user to click on payment to create a draft sale order is not intuitive, and generate more actions.

hparfr:
Adding buttons in this placeholder can be done in a other PR.

legalsylvain:
> Adding buttons in this placeholder can be done in a other PR.

I don't understand what you propose.

hparfr:
In the use-cases where you want payment and orders (w/w picking) it makes sense to have buttons in the payment screen. It's the logical, and great UI. 

When porting this module, I had to think where to put theses buttons. It was not a good idea to put it back in the ticket. I had to put it somewhere. After some tries, the best place to put it is on the payment screen. It's convenient for all the 6 cases (pos, draft, confirmed, confirmed + payment, picking, picking + payment).

Putting it on the placeholder where you indicate, and accept all the 6 cases rises new questions which are not trivial to implement from a user experience perspective. The best way I see it, is to have buttons on the payment screen (done in this PR) AND offer some of them in the placeholder based on the config selected. 

If you never take payments, I understand it's always 1 click more. If it's too much pain for the customer, this shortcut can be added to this module as pull request by anyone.

legalsylvain:
Well, I did'nt say to put all the button in the bill screen. Just the existing ones : 

> pos, draft, confirmed, confirmed + payment, picking, picking + payment).

* pos -> don't understand what you mean. (natural process ?). For me Nothing to Do.
* draft, confirmed, picking -> on the bill screen (as in V8). nothing to do with the payment screen.
* confirmed + payment / picking + payment -> on the payment screen. It has something to do with the payment, of course ! 

regards.

pos_order_to_sale_order/models/pos_config.py
legalsylvain: It's not the same word so it can not work.

mourad-ehm:
Points of sale was extented to allow managing Pos to sale order workflow.
However payment of sale order is not managed in this module.
So we propose to manage this by :
1- extending the module to integrate to it the payment feature for sale order (solution used by Akretion in odoo 8.0 see https://github.com/akretion/pos/tree/8.0-fix-pos-account-bank-statement/pos_sale_order)
2- Adding a new module for this feature.
In this solution (1 or 2) we add payments on sale_order object (link to bank statement).
We create an user anonymous to allow to create order for not identified customer.
We put a new journal (Pos Sale ORDER) to separate regular invoices from pos anonymous invoices.
Accounting entry are handled by generating invoice. When we close pos session we generate grouped invoice for anonymous customer. And for each not invoiced order with identified customer we generate an invoice too. Then we reconcile invoices with payments (linked on order).

Use case :
A customer who pass to point of sale to take there products that are invoiced at the end of month for the major part of sale orders. So actually if customer pay cash its order we must generate a pos order (to make payment). This is not good for the Follow Up and Statistics. If we add payments management on pos to sale order we can make all the process in sale order (cash payment or late payment). When a company asks for customization of sale (by adding new fields on sale order or sale product with options for example). It is easier to maintain this customization on an object (order of sale) than on 2 objects (sale and order of position).

I think that the solution 1 is easier to code (JS integrated) and maintain.

mourad-ehm:
Hi @legalsylvain, @pedrobaeza, @hparfr, @chienandalu, 
What do you think about my proposition ?

mourad-ehm:
Ping @bealdav

legalsylvain:
Hi @mourad-ehm 
- Well, I'm more in favour on solution 2, as this interesting new feature require extra dependencies.  
- btw, what do you think about what I proposed here https://github.com/OCA/pos/pull/264#pullrequestreview-107969471 ? it's more or less what you propose, or did I missed something ?
- I don't understand the creation of an anonymous customer. Did you planned to handle it with the module ?
- About payment, I think the new module should depend on account_cash_invoice that realizes the work. Don't you think ?

regards.

legalsylvain (changes requested):
Anonymous partner has nothing to do with this module.
If you want such feature, please consider working on this module https://github.com/OCA/pos/pull/136

Regards.

rafaelbn:
Hi @hparfr , as today is friday :smile: could you attend to @legalsylvain and @chienandalu comments please? in order to continue with this PR? Thanks a lot! Regards

legalsylvain:
A quick look in the code shows that problems mentioned in the first comments are not resolved. 
Please ping me when it's ok ! 
kind regards.

hparfr:
@legalsylvain do you talk about *your* first comment ?

legalsylvain:
yes.

legalsylvain:
closing, due to inactivity. Feel free to reopen.

PR #152, opened 2017-02-10
[10.0][WIP][post_list_order] Make possible to list pos orders in the POS (in this moment only available in backend)

ovnicraft:
Add a button to list all orders for active session
* [ ] include state on list
* [ ] search

Please give your feedback about.
button un main screen
captura de pantalla 2017-02-10 a las 12 22 53

List orders

captura de pantalla 2017-02-10 a las 12 23 02

PR #117, opened 2016-09-04
[ADD] 9.0 pos return

legalsylvain:
This module was written to extend the functionality of odoo Point Of Sale about Order returns.

With this module, it is now forbidden to return more products than the quantity of the initial Order.
- A link is created between the returned Order and the initial Order.
- A link is created between the returned Order Line and the initial Order Line.

Review welcome. kind regards.

PR #269, opened 2018-04-26
[MIG] pos_order_return: Migration to 10.0

chienandalu:
PoS Order Return
================

This module extends the functionality of odoo Point Of Sale about POS Order
returns.

With this module, it is now forbidden to return more quantity than the initial
one.

A link is created between the returned Order and the initial Order.
A link is created between the returned Order Line and the initial Order Line.

cc @Tecnativa


chienandalu:
Migrated from https://github.com/OCA/pos/pull/117

rafaelbn:
@chienandalu please, could you check travis? :red_circle:  Thanks

chienandalu:
I'd appreciate your functional review on runbot as well :wink:

rafaelbn:
@chienandalu when I try to choose a customer in the POS I get this error

`Unknown CORS error
An unknown CORS error occured. The error probably originates from a JavaScript file served from a different origin. (Opening your browser console might give you a hint on the error.)`

Maybe is not related to this PR, not sure.

rafaelbn:
@chienandalu I'm not able to complete a return

- Make in the POS an order
- Go to backend
- Return the order
- Press the "Payment" button after that -> Odoo warning

`
Please provide a partner for the sale.
`

So you cannot return... :cry: 

Please check

pos_order_return/README.rst
```diff
@@ -0,0 +1,102 @@
...
+With this module, it is now forbidden to return more quantity than the initial
+one.
+
+A link is created between the returned Order and the initial Order.
+A link is created between the returned Order Line and the initial Order Line.
+
+Usage
+=====
+
+Select an PoS Order an choose either *Return Products* (full return of the
```
rafaelbn, line 21: s/Select/Select in the backend

I expected to make a return from the POS and not need to go to the backend to return and order! Is it possible?

cc @legalsylvain

rafaelbn (commented):
With doubts

pos_order_return/README.rst
chienandalu: It's possible. I'm working on a superseed of pos_order_list (yet to publish) to include it as a dependency here

legalsylvain: > I expected to make a return from the POS and not need to go to the backend to return and order! Is it possible?

> It's possible. I'm working on a superseed of pos_order_list (yet to publish) to include it as a dependency here

What is pos_order_list ? I don't see it in OCA/pos.

In the meantime, I worked (V8) on a module to list the done pos orders. For the time being, it display on PoS front UI the list of the past done / invoiced orders and allow to reprint the bill, but allowing to make a refund from this module can be easely done.

See the current module here (GRAP incubator repository), but I can make a PR against OCA/pos, if OCA members are interested by this module.

https://github.com/grap/grap-odoo-incubator/tree/8.0/pos_done_order_load

(see roadmap parts).

regards.

chienandalu: I meant this PR: https://github.com/OCA/pos/pull/152 I think it can be a good base to extend from (has a little bug yet though).

chienandalu:
Can you test again @rafaelbn ?

pos_order_return/models/pos_order.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+    refund_order_qty = fields.Integer(
+        compute='_compute_refund_order_qty',
+        string='Refund Orders Quantity',
+        digits=dp.get_precision('Product Unit of Measure'),
+    )
+
+    def _compute_refund_order_qty(self):
+        for order in self:
+            order.refund_order_qty = len(order.refund_order_ids)
```
pedrobaeza: This can be best performance implemented with a `read_group` and using count returned value inside the tuple.

pos_order_return/models/pos_order.py
```diff
@@ -0,0 +1,193 @@
...
+            return line
+        # Goes to refund invoice thus it should be positive
+        line.quantity = -line.quantity
+        return line
+
+    def _action_pos_order_invoice(self):
+        """Wrap common process"""
+        self.action_pos_order_invoice()
+        self.invoice_id.sudo().action_invoice_open()
+        self.account_move = self.invoice_id.move_id
```
pedrobaeza, line 73: Why this here in the return?

pedrobaeza (changes requested):

rafaelbn (approved):
Tested again functionally :+1: 

Please check comments and about test. Is it possible to improve them '  44.09% of diff hit (target 82.12%) ' . I know it's complicated to test in POS so maybe it too much for this PR...

pos_order_return/models/pos_order.py
chienandalu: Do you have an example of it?

pos_order_return/models/pos_order.py
chienandalu: I'm replicating the same behaviour that `point_of_sale` has in normal mode

legalsylvain: @pedrobaeza : are you ok with the answer of @chienandalu ?

pedrobaeza: I have to check with him again as I don't understand the reasons for invoicing the return.

legalsylvain:
> Please check comments and about test. Is it possible to improve them ' 44.09% of diff hit (target 82.12%) ' . I know it's complicated to test in POS so maybe it too much for this PR...

No in that case, we can call create_from_ui and other python function to test. No need to create JS test. So I think tests should be added. 

For JS test, please you can inspire from some other PoS module with test like : 
https://github.com/grap/l10n-france/blob/10.0_l10n_fr_certification_pos_offline/l10n_fr_certification_pos_offline/static/src/tests/tests.js

regards.

pos_order_return/models/pos_order.py
chienandalu: It comes from here: https://github.com/OCA/pos/pull/269/commits/00db862fdf891182563ef8a1342347e939da9075#diff-3735e1293bd7b48ab8f6f80fac232206R107 If the original order had an invoice, a refund invoice is made automatically

pedrobaeza: OK, got it.

pedrobaeza (approved):

PR #277, opened 2018-06-22
[8.0] fixes and improvment on pos_picking_load

legalsylvain:
-> fix the load of the field quantity. 
-> add parameter to let the possibility to load custom quantity of pickings
-> prepare function to let the possibility to overrides search function.

Thanks for your review.

### DIFF
--- a/pos_picking_load/__openerp__.py
+++ b/pos_picking_load/__openerp__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'POS Picking Load',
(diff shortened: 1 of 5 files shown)


legalsylvain:
No review on that improvment. @StefanRijnhart, could you take a look on that PR ? Not sure but I think you was using this module.

kind regards.

quentinDupont:
Nice improvments, thank you !

Issue #279, opened 2018-07-05
[pos_pricelist] Dropdown menu default menu error


Ark74:
**Issue:** 
As mentioned on title.

**Version:**
Tested on http://3331405-10-0-9ee349.runbot1.odoo-community.org (Odoo 10)

**Steps to reproduce:**
1. On POS Settings enable Pricelist (bottom of the page)
2. Try to change the **Default Pricelist** on the dropdown menu.
3. Error!
![dd-menu-pos](https://user-images.githubusercontent.com/7444518/42349804-14b86bec-806b-11e8-9e34-3353ade5fdf3.png)

chienandalu:
@Ark74 Check here: https://github.com/OCA/pos/pull/280

Ark74:
Thank you very much for looking into this. :pray: 
Please close this issue at will.

PR #280, opened 2018-07-09
[FIX] pos_pricelist: right operator in default pricelist

chienandalu:
Solves this issue: https://github.com/OCA/pos/issues/279

cc @Tecnativa

### DIFF
--- a/pos_pricelist/views/pos_config_view.xml
+++ b/pos_pricelist/views/pos_config_view.xml
@@ -25,7 +25,7 @@
                             widget="many2many_tags"/>
                         <field
                             name="pricelist_id"
-                            domain="[('id', 'in', available_pricelist_ids)]"
+                            domain="[('id', 'in', available_pricelist_ids and available_pricelist_ids[0] and available_pricelist_ids[0][2] or False)]"
                             options="{'no_create': True}"/>
                         <button
                             name="%(product.product_pricelist_action2)d"
--- a/pos_pricelist/__manifest__.py
+++ b/pos_pricelist/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'POS Pricelist',
-    'version': '10.0.1.0.1',
+    'version': '10.0.1.0.2',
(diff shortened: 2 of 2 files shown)


pos_pricelist/views/pos_config_view.xml
```diff
@@ -25,7 +25,7 @@
                             widget="many2many_tags"/>
                         <field
                             name="pricelist_id"
-                            domain="[('id', 'in', available_pricelist_ids)]"
+                            domain="[('id', 'in', available_pricelist_ids and available_pricelist_ids[0] and available_pricelist_ids[0][2] or False)]"
```
legalsylvain, line 28: This domain will return only one id ? Or did  I miss something ? I think it should return all the ids available in the field available_pricelist_ids.

pedrobaeza: No, it returns all the existing ones, as it's stored as [(6, 0, [`<ids>`])], so [0][2] accesses [`<ids>`], but I didn't know this works on the JS part. @chienandalu is it working?

chienandalu: Yes, it is :slightly_smiling_face: 
image

chienandalu: In v11 the right operator is more transparent, anyway (this bug was due to the literal copy of the view in this backport)

pedrobaeza: Thanks for the explanation

pedrobaeza (approved):
Please change as always module version number

chienandalu:
@pedrobaeza done

legalsylvain (approved):
@pedrobaeza : thanks for the explanation, I didn't know.
@chienandalu : thanks for this patch !

pedrobaeza:
@legalsylvain as explained by my colleague, on v11 these values were homogenized in the new JS views for x2many widget for being plain lists.

Issue #281, opened 2018-07-10
[8.0] pos_pricelist: pos.order.tax ACL group_pos_user unable to create


ssaid:
Currently the acl is defined as
`access_pos_order_tax,pos.order.tax,model_pos_order_tax,point_of_sale.group_pos_user,1,0,0,0
`
When creating a pos_order with this group it fails as expected.
This appears to be a bug, isn't it?

PR #282, opened 2018-07-10
fix: pos.order.tax ACL for group_pos_user

ssaid:
solves #281


legalsylvain:
Hi. well pos.order.tax are created by ```_amount_line_tax```, called by ```_amount_all```.

It's a computed function, so I think that some ACL are not checked. 

Personaly I use this module in production since monthes, and there is no problem.
But yes, It's an error I think.

legalsylvain (approved):
thanks.

PR #288, opened 2018-09-13
[MIG] pos_loyalty 11.0

kirca:


kirca:
Made update to README.rst to follow new guidelines and rebased

pos_loyalty/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+To use this module, you need to:
+
+* Go to *Point of Sale > Configuration > Loyalty Programs* and define a new loyalty program with specific rules and gifts.
+* Assign the loyalty program to the desired Point of Sale.
```
chienandalu: This info should be in CONFIGURATION.rst

I also think this module is complex enough to write a more detailed usage instructions.

pos_loyalty/models/loyalty_reward.py
```diff
@@ -0,0 +1,68 @@
...
+                                         help='The Loyalty Program this reward'
+                                              ' belongs to')
+    gift_product_id = fields.Many2one(comodel_name='product.product',
+                                      string='Gift Product',
+                                      help='The product given as a reward')
+    discount_product_id = fields.Many2one(comodel_name='product.product',
+                                          string='Discount Product',
+                                          help='The product used to apply '
+                                               'discounts')
+    point_product_id = fields.Many2one(comodel_name='product.product',
```
chienandalu, line 44: Shouln't these fields have a `pos_available = True` domain?

pos_loyalty/README.rst
```diff
@@ -0,0 +1 @@
+**This file is going to be generated by oca-gen-addon-readme.**
```
chienandalu: Better to generate it locally to save the later bot commit

chienandalu (changes requested):
Thanks for the great work! I've been testing it on runbot and it works fine overall. Just some comments below:

pos_loyalty/models/loyalty_reward.py
kirca: You are right, only pos products should be available since it wont work otherwise.

chienandalu (approved):
:+1: Great! :smile:

kirca:
Thanks for the review @chienandalu !

legalsylvain:
Hi @kirca, could you cherry pick this commit ? 
https://github.com/OCA/pos/pull/294
thanks !

pos_loyalty/models/pos_config.py
```diff
@@ -1,6 +1,6 @@
-# -*- coding: utf-8 -*-
 # Copyright 2004-2010 OpenERP SA
 # Copyright 2017 RGB Consulting S.L. (https://www.rgbconsulting.com)
+# Copyright 2018 Lambda IS DOOEL <https://www.lambda-is.com>
```
legalsylvain: You added copyright in files you did'nt changed. (done many times).

pos_loyalty/views/templates.xml
```diff
@@ -3,6 +3,7 @@
     <template id="assets" inherit_id="point_of_sale.assets">
         <xpath expr="." position="inside">
             <script type="text/javascript" src="/pos_loyalty/static/src/js/pos.js"></script>
+            <script type="text/javascript" src="/pos_loyalty/static/src/js/tests.js"></script>
```
legalsylvain: Is it a good way to include JS test ?
I mean, this file will be loaded allways, in production, even if it's not necessary.

regards.

legalsylvain (changes requested):
Some minor remarks and a question about JS test. 

thanks a lot for porting this module, and adding test ! great addition.

kind regards.

pos_loyalty/models/pos_config.py
kirca: encoding declaration `# -*- coding: utf-8 -*-` line was removed on all .py files. Not sure if that constitutes as a change.

pos_loyalty/views/templates.xml
kirca: I can't see a reliable way on how to distinguish between a production and not-production session. Please share if you have an idea.

Also the inherited template already includes the tests from the point_of_sale so I don't see why this should be considered as an exception.

kirca:
@legalsylvain, I've included the commit and added my comments in your review. Let me know if you have something to add. Thanks

pos_loyalty/models/pos_config.py
legalsylvain: I don't think so. copyright is generally based on code developpers have written. In that case, it's just a cosmetic changes, due to OCA decision.

pos_loyalty/views/templates.xml
legalsylvain: Excuse me, it was not about production or non production but about with demo data or without. In that case, adding a new xml file, imported in the 'demo' data of the manifest will let it available for test (what we want) but avoid to load it on production. (but yes, it should be done in the same inherit_id = point_off_sale.assets".
What do you think ?

pos_loyalty/models/pos_config.py
kirca: Sorry but I don't agree with this argument. I think overall **all** changes should be considered, even cosmetic ones, since they increase the maintainability of the file/project. Otherwise there is no point in doing them.

Although this is a really trivial case so I'll just remove the copyright.

pos_loyalty/views/templates.xml
kirca: That is a good idea, it depends on demo data so it wont work if that isn't loaded anyway. Thanks, I'll make the change

kirca:
Hi @legalsylvain, can you please check latest changes. Thanks!

kirca:
@legalsylvain can you check if we can move this forward. Thanks

legalsylvain (approved):
Hi @kirca. 

my apologyze for the late. Code review (no test, runbot seems to be off)

thanks for porting this module !

kirca:
Hi @legalsylvain, no apology is needed. Thanks again for the review and merge!

PR #290, opened 2018-09-20
[8.0][FIX] Error pos_pricelist compare Dates 

RodrigoBM:
Error pos_pricelist compare Dates in function find_valid_pricelist_version.

The var date that initialize with new Date(), and this have the date and the time in the moment of the creation. If you have a pricelist with a date_start or date_end those dates only have the date, and the times inicilice to 0. If you compare for example date_end (Thu Sep 20 2018 00:00:00) >= date (Thu Sep 20 2018 14:24:56), the comparison is wrong, when this has to be correct.

Regards


legalsylvain:
Propose to close for lake of review, and because 8.0 is quite obsolete.
Thanks @RodrigoBM for your proposal.
 
kind regards.

PR #291, opened 2018-09-24
[10][FIX] pos_product_template : don't load module in web backend

hparfr:
Otherwise it's causing JS issues in the backend.


PierrickBrun (approved):

legalsylvain:
thanks @hparfr !
fast tracking trivial patch.

PR #293, opened 2018-09-27
ADD pos_picking_delayed

legalsylvain:
This PR add a new module that delay the creation of the stock.picking created when a pos.order is created.

This module is interesting specially in a context of Synchroneous Point Of Sale that is introduced by certification modules like 'l10n_fr_pos_cert' because in such cases, the bill will be printed only when the pos order is created (after the call of the function create_from_ui) and the creation of the picking is the process that takes the most time.

- See https://github.com/odoo/odoo/pull/26314#issuecomment-422949266 for more information
- see the readme for a complete description


pos_picking_delayed/models/pos_order.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo import api, models
+
+
+class PosOrder(models.Model):
+    _inherit = 'pos.order'
+
+    @api.model
+    def create_from_ui(self, orders):
+        return super(PosOrder, self.with_context(
+            picking_creation_delayed=True)).create_from_ui(orders)
```
chienandalu: Wouldn't it be interesting to add a configuration by PoS?

legalsylvain: Hi @chienandalu ! Thanks for your review ! 

Indeed, I added a boolean field for the setting, on pos.config, and added a test.

pos_loyalty/models/pos_order.py
```diff
@@ -24,7 +24,7 @@ def create_from_ui(self, orders):
         res = super(PosOrder, self).create_from_ui(orders)
         for order in orders:
             order_partner = order['data']['partner_id']
-            order_points = order['data']['loyalty_points']
+            order_points = order['data'].get('loyalty_points', 0)
```
flotho: @legalsylvain Are you sure that this part should be in this branch ?

pos_picking_delayed/models/pos_order.py
flotho: Hi @legalsylvain , actually you don't use the option you set up picking_creation_delayed if I'm correct
Won't it be more readable to directly pass the context with the value of your option

flotho:
Hi @legalsylvain , Some little comments on this one but it lokks good to me anyway.

Good job !

pos_loyalty/models/pos_order.py
legalsylvain: This is a trivial patch of pos_loyalty (Travis fail with the current PR, without this patch, when I create an order during the test of pos_picking_delayed) but I can revert it and make it in another PR to make the cherry pick working more easy, you're right.

pos_picking_delayed/models/pos_order.py
chienandalu: I guess it would be more logical and simple to check the pos config here and return the `super()` with the context flag or not.

pos_loyalty/models/pos_order.py
legalsylvain: Done in https://github.com/OCA/pos/pull/294.

pos_picking_delayed/models/pos_order.py
legalsylvain: Not so easy. in create_from_ui, I don't have an easy access to the current pos_config, because it's an @api.model function, and ``orders`` is just a big json. I think parsing the json to know the session, and then the config, and then check the value of picking_creation_delayed will be more line of code.

The test is done later in ``create_picking`` function. (test of the context then test of the pos config setting).

chienandalu: You're right there :slightly_smiling_face: Although you'd save splitting the recordset call in `create_picking`. I can't tell wich method would perfomance best, anyway :slightly_smiling_face:

chienandalu (approved):

pos_picking_delayed/models/pos_order.py
legalsylvain: Yes, I split but I checked the code of ``create_picking`` and the first lines include a ``for order in self:``
So I don't understand how the code of this module could make the process slower. 

Réf : https://github.com/odoo/odoo/blob/10.0/addons/point_of_sale/models/pos_order.py#L643

don't you think ?

regards.

chienandalu: Yes, my guess is that your option is the right one, therfore I approved :smile:

pos_picking_delayed/models/pos_order.py
```diff
@@ -0,0 +1,38 @@
...
+                if not order.config_id.picking_creation_delayed:
+                    super(PosOrder, order).create_picking()
+            return True
+        else:
+            return super(PosOrder, self).create_picking()
+
+    @api.model
+    def create_delayed_picking(self):
+        orders = self.search([
+            ('state', '!=', 'draft'), ('picking_id', '=', False)],
```
flotho: I think we have to be more explicit in the search parameters
If you sell something else than consu/product then the search will return more and more pickings and will possibly be stucked on multiple orders that have only service products.
maybe a search on lines or a computed fields indicating that picking is expecting would do the trick.

regards

legalsylvain:
Hi @flotho, thanks for the remarks. I changed the code, adding a new boolean field "has_picking_delayed" that will be set to False when create_picking is called, even if picking is not created.

Ready for review.

Regards.

pos_picking_delayed/models/pos_order.py
```diff
@@ -0,0 +1,46 @@
...
+    def create_from_ui(self, orders):
+        return super(PosOrder, self.with_context(
+            picking_creation_delayed=True)).create_from_ui(orders)
+
+    def create_picking(self):
+        if self.env.context.get('picking_creation_delayed', False):
+            for order in self:
+                if not order.config_id.picking_creation_delayed:
+                    super(PosOrder, order).create_picking()
+                    order.write({'has_picking_delayed': False})
```
flotho: I think a more precise control has to be set here, for old pickings, it set the fields to true even if a picking already exists for the order. This causes huge computation load and slow down all the process.

flotho (changes requested):

pos_picking_delayed/models/pos_order.py
legalsylvain: You're right, the default value was bad. I set it to False, and changed the algorigthm.
https://github.com/OCA/pos/pull/293/commits/2bb9653d8a30d170b5323cd383f3561bde8bda61

Should be ok, now.
regards.

chienandalu:
@flotho All your comments were attended. Can you make your final review?

pos_picking_delayed/models/pos_order.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Custom Section
+    @api.model
+    def create_delayed_picking(self):
+        orders = self.search([
+            ('state', '!=', 'draft'),
+            ('has_picking_delayed', '=', True),
+        ], order='date_order')
+        for order in orders:
+            order.sudo(order.user_id.id).with_context(
```
rousseldenis, line 54: Even if this is called through cron, you will create two environments (one for 'sudo' and one for 'with_context').

Why don't you use queue_job process ? Moreover, it will allow you to follow the picking creation status in job lines.

rafaelbn (approved):
Thanks @legalsylvain :+1:

pos_picking_delayed/models/pos_order.py
legalsylvain: > Even if this is called through cron, you will create two environments (one for 'sudo' and one for 'with_context').

Do you have an alternative ?

> Why don't you use queue_job process ? Moreover, it will allow you to follow the picking creation status in job lines.

Indeed, could you be a great improvement. I personaly don't use queue_job, but feel free to make a PR to improve this module ! Otherwise, we can add a simple remark in the roadmap section.

rousseldenis: > > Even if this is called through cron, you will create two environments (one for 'sudo' and one for 'with_context').
> 
> Do you have an alternative ?

One could be grouping orders by company and changing context once.

> 
> > Why don't you use queue_job process ? Moreover, it will allow you to follow the picking creation status in job lines.
> 
> Indeed, could you be a great improvement. I personaly don't use queue_job, but feel free to make a PR to improve this module ! Otherwise, we can add a simple remark in the roadmap section.

Ok for ROADMAP

chienandalu:
@legalsylvain can you squash your commits and add those notes to the README?

legalsylvain:
- roadmap added
- squatch done 

regards, and thanks for your review.

chienandalu:
Thanks to you @legalsylvain ! :slightly_smiling_face:

PR #294, opened 2018-09-28
[FIX] Travis is failing when test are creating orders via create_from_ui when pos_loyalty is installed

legalsylvain:
Trivial patch. 
See : https://github.com/OCA/pos/pull/293#pullrequestreview-159774945
thanks for your review.

CC : @chienandalu, @flotho .


chienandalu (approved):
:+1:

legalsylvain:
fast tracking this patch, to fix #293 PR.

PR #295, opened 2018-10-08
[FIX] add pos_fix_search_limit dependency 

mourad-ehm:
add pos_fix_search_limit dependency to fix template lacking bug if more than 100 variants.
@hparfr

### DIFF
--- a/pos_product_template/__manifest__.py
+++ b/pos_product_template/__manifest__.py
@@ -1,7 +1,7 @@
 # -*- encoding: utf-8 -*-
 {
     'name': 'POS - Product Template',
-    'version': '10.0.1.0.2',
+    'version': '10.0.2.0.1',
     'category': 'Point Of Sale',
     'author': "Akretion,Odoo Community Association (OCA)",
     'summary': 'Manage Product Template in Front End Point Of Sale',
@@ -9,6 +9,7 @@
     'license': 'AGPL-3',
     'depends': [
         'point_of_sale',
+        'pos_fix_search_limit',
     ],
     'data': [
         'view/view.xml',


hparfr (approved):
Thanks @mourad-ehm

legalsylvain (approved):
Makes sense. not tested. :+1: 
thanks @mourad-ehm !

mourad-ehm:
ping @bealdav @chafique-delli could you review this PR ?

chafique-delli (approved):
It's ok for me.

pedrobaeza:
I don't see correct to add a new dependency in a stable module. This will break all the existing installations that don't have this module installed. You can add a recommendation of installing that module, but not to force it.

pedrobaeza:
@legalsylvain you have merged while I was writing previous comment...

legalsylvain:
image

Yes. concurrent work.

@pedrobaeza : what about to increment revision of this module ? It should fix the problem, don't you think ?

pedrobaeza:
I'm afraid not. Take into account the following scenario:

- You have previous version of `pos_product_template`.
- You don't have `pos_fix_search_limit` installed.
- Update code for your repositories.
- Service restarted.
- `pos_product_template` module won't be loaded because there's an unresolved dependency.
- It's not auto-installed even in the case you have the module accessible on your addons path.

hparfr:
Increment the major version number allows to introduce breaking changes, isn't it ?

legalsylvain:
> - Update code for your repositories.
> - Service restarted.

Yes, this is because a step is missing. When restarting, you should ``-u modules``. (at least, the modules where the code changed, or ``-u all`` otherwise). If you don't want to ``-u``, you should not update your code. It will fail a lot of time. (if a xml file is added or updated for exemple)

I tested, and it works correctly. If I add uninstalled ``module_A`` as a dependence of ``module_B`` that is installed, and if I do ``-u module_B``, it will install ``module_A``

I understand you that maybe some production environment that doesn't deploy correctly will fail but there is a bug, that make the module quite unusable, and this dependency fix it. So it's important.

legalsylvain:
> Increment the major version number allows to introduce breaking changes, isn't it ?

I'm not very clear on that point. For me it's "informational" changes. And also I think that if you change the version number, Odoo will propose you to update the code, via the UI. (but I don't use this feature)

pedrobaeza:
@legalsylvain with regular Odoo (not OpenUpgrade), AFAIK, updating a module with a dependency not installed doesn't install that dependency. Are you sure about this?

About correctly deployed, it's also interpretable, as I deploy my systems correctly, but limiting the exposed addons for avoiding possible side effects or limiting customer actions installing not covered modules.

legalsylvain:
> @legalsylvain with regular Odoo (not OpenUpgrade), AFAIK, updating a module with a dependency not installed doesn't install that dependency. Are you sure about this?

Just tested on a 8.0, based on OCA/OCB. and yes. Maybe not in more recent version.

> About correctly deployed, it's also interpretable, as I deploy my systems correctly, but limiting the exposed addons for avoiding possible side effects or limiting customer actions installing not covered modules.

It was not a critic ! I do sometimes the same for any good reasons ! but yes, I think that if we pull code, we should update code. (that is the objective of the module from acsone that check if code changed, doesn't it ?)
regards.

pedrobaeza:
Hehe, yes, it was more our module and upgraded by Acsone, but we detected that need and do it in first place in collaboration with LasLabs :wink: 

I'm pretty sure in v10 auto-install is not happening, but hey, I won't fight anymore as I'm not a user of this myself and I don't want to be always the nitpicking man.

legalsylvain:
> Hehe, yes, it was more our module and upgraded by Acsone

sorry, my bad ! 

> I'm pretty sure in v10 auto-install is not happening, but hey, I won't fight anymore as I'm not a user of this myself and I don't want to be always the nitpicking man.

Just tested in a fresh database with point_of_sale installed. (OCA/OCB/10.0)
- add a dependency in ``point_of_sale`` to ``crm``
- restart, with ``-u point_of_sale``
- ``crm`` is installed

pedrobaeza:
OK, great, then I should change my assumptions about this.

PR #297, opened 2018-10-19
[11.0][ADD] pos_order_mgmt: New module

chienandalu:
This module extends the functionality of the PoS frontend allowing to load
already done PoS Orders in order to be able to operate over them, being able to
reprint past tickets o return them.

Based on @legalsylvain work here https://github.com/grap/grap-odoo-incubator/tree/8.0/pos_done_order_load but reconducted to manage the loaded orders from a list screen and filling the shortcomings of that one.

Hopefully it can be useful as a base for other modules.

cc @Tecnativa


pos_order_list/models/pos_config.py
```diff
@@ -0,0 +1,24 @@
...
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    iface_load_done_order = fields.Boolean(
+        string='Load Done Orders',
+        default=False,
```
legalsylvain:
```suggestion
        default=True,
```

We can guess that if users install the module, he want to see the feature enabled.

legalsylvain (changes requested):
Hi @chienandalu. Thanks for porting this module !
Looks nice. (one comment inline about default value.)
other things : 
- could you take too from the original module, the screen shot in static / description and reference in the readme part (or add new one) ? It's good in OCA appstore to see what is the screen provided by the module.
- I have an error, testing on runbot, when I try to reprint or reload the pos.order. did you faced the same problem ?

kind regards.

image

chienandalu:
Mm, I'm testing but I couldn't reproduce the error either from new or reloaded orders. What steps did you follow?

chienandalu:
Ok, I tested with Firefox and failed as you described. It should be fixed right now.

pos_order_list/README.rst
```diff
@@ -0,0 +1,110 @@
...
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/184/11.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module extends the functionality of the PoS frontend allowing to load
+already done PoS Orders in order to be able to operate over them, being able to
+reprint past tickets o return them.
```
legalsylvain:
```suggestion
reprint past tickets or return them.
```

legalsylvain (changes requested):
Hi. Thanks !
Well, testing on rubot, Seems that this module has feature already placed in another OCA module 'pos_order_return'. 
So actually, you can see that if you return an order on runbot, with the other module, an other error occures when validating.

image


solution I see : 
- merge the two modules. (big work) 
- make your depend on the other one, and send the value of the refunded order. returned_order_id
https://github.com/OCA/pos/blob/11.0/pos_order_return/models/pos_order.py#L14
(more easy, just overload the JS function that prepare the order.

what do you think ?

kind regards.

chienandalu:
@legalsylvain It was kinda tricky (we're dealing with synced and unsynced orders) but I could make both modules get along together :slightly_smiling_face:

pedrobaeza:
@legalsylvain is everything correct now?

chienandalu:
@legalsylvain Could you re-review again?

rafaelbn:
Hello @legalsylvain ! Could you please RE-Review this PR? We are running it in production and I think that it could be very useful to have it merged for more visibility about OCA POS improvements in App Store. As much users they use as more mature will be. It looks like other Payment modules are more popular.

Thanks

pedrobaeza:
hahaha, we (the 2) have requested at the same time the same thing.

pedrobaeza:
(this is the result of an internal chat conversation)

legalsylvain:
Hi @chienandalu. First, thanks for the changes.
All is working like a charm. (tested on runbot) reprint, refund.
the only missing feature that in pos_order_return you can not refund many times the same order. (more precisely, for each line, you can not refund more than the original quantity).
this is handled with that part of code in ``pos.order.line`` : https://github.com/OCA/pos/blob/11.0/pos_order_return/models/pos_order.py#L135

## Use case
**via back office (pos_order_return)** 
- select an order A, make a refund, confirm.
- select the order A again, make a refund : the refund is empty. OK

**Front office (pos_order_mngmt)**
- select an order A, make a refund, confirm.
- select the order A again, make a refund  You can refund many times. KO

This is visible with the partial refund wizard :
image

**solution 1** :  you introduce returned_line_id in your Javasript code (can be a big work, to handle all the cases)

**solution 2** : you introduce a roadmap / know issue in your rst file, mentionning that the check in front office is not implemented. (less work ! ;-)) 

kind regards.

otherwise, LGTM !

chienandalu:
> he only missing feature that in pos_order_return you can not refund many times the same order

@legalsylvain Ok, I'll let it for a future revision as we invested a lot of time in this already :wink:

legalsylvain (approved):
OK. I permit me to update the roadmap file in your branch to mention that this feature is partially implemented.

let's go merging this PR.

thanks for this work !

pedrobaeza:
@chienandalu Travis is red

chienandalu:
@pedrobaeza Yes, there's an issue with pos_loyalty yet to be resolved: https://github.com/OCA/pos/issues/309

pedrobaeza:
OK, then let's merge it right now

PR #300, opened 2018-10-26
 [MIG] pos_ticket_logo: Migration to 11.0

chienandalu:
A different receipt template is used if the PoS ticket is printed via web or
via proxy. In the case the ticket is printed via web (through the browser) the
company logo isn't printed. This module adds it.

cc @Tecnativa


LoisRForgeFlow (approved):
Tested on runbot :+1:

pedrobaeza (approved):

PR #308, opened 2018-11-15
[FIX] pos_loyalty: frontend test failing in CI

kirca:
Add delay in steps to reduce timing errors when tests are running


kirca:
Build still fails so it is not a timing problem. I'll close this and open an issue.

PR #312, opened 2018-12-04
[FIX] display /kg for to_weight products

legalsylvain:
Trivial patch. restore the display of '/kg' for to_weight product that is hidden where pos_pricelist is installed.

before this patch :

image

After this patch :

image

CC : @quentinDupont

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -577,7 +577,8 @@ function pos_pricelist_models(instance, module) {
                         price = this.pos_widget.format_currency(price);
                         if (k == 0) {
                             if (product.to_weight) {
-                                $(product_ui).find('.price-tag').html(price + '/Kg');
+                                $(product_ui).find('.price-tag').html(price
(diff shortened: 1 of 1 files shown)


rousseldenis:
@legalsylvain I don't really know this flow but IMHO it should depends on product unit you defined on product no ?

legalsylvain:
hi @legalsylvain. Thanks for your review.

You're right ! 

https://github.com/odoo/odoo/blob/8.0/addons/point_of_sale/static/src/xml/pos.xml#L754

I'll do the change.

kind regards.

quentinDupont:
Thanks for the patch ! :)

legalsylvain:
done. Note : travis is failing for unrelated reason. (master 8.0 is fucked https://travis-ci.org/OCA/pos/branches)

joj828 (approved):
Code Review , test
it works fine.
Thank you for your contribution

rousseldenis:
@legalsylvain Yes, it seems that pylint does not like the js in pos_pricelist... Maybe someone should fix the code...

rousseldenis (approved):

PR #313, opened 2018-12-06
Migrate pos_default_payment_method from v10

CGenie:
Seems to be working in v11, with 1 small change for the views pos box id.


pedrobaeza:
Thanks for this, but you should preserve commit history following method in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0#howto

PR #317, opened 2018-12-30
[10.0][MIG] pos picking load

legalsylvain:
Port 'pos_picking_load' from 8.0 to 10.0 serie

### Refactoring part

* Use new readme OCA files
* update description and screen shots
* remove obsolete workflow on sale order
* use new PoS buttons box
* split js files in models.js and widget.js files
* fix pylint warning (a lot !)
* JS refactoring. (simplification, + add prepare function)
* make the current picking id / name persistent if the PoS is reloaded

### New feature and improvment part
*  display the name of the picking in the point of sale, (if a picking is loaded)
* add new checks, to avoid a picking to be loaded multiple times in many PoS tabs
* improve picking selection. (behaviour similar to the partner selection) 

You can read the full description [here](https://github.com/legalsylvain/pos/tree/10.0-mig-pos_picking_load/pos_picking_load)

Thanks for your review.


legalsylvain:
CC : original reviewers @StefanRijnhart, @hparfr

legalsylvain:
I don't understand why codecov is red and coveralls is green. 

codecov says that two files are red, even if I didn't changed them.
image

pos_picking_load/demo/sale_order.xml
```diff
@@ -0,0 +1,53 @@
+<openerp><data>
```
kirca, line 1: Perhaps it is wise and change the top element in all xml files to `</odoo>` to future proof.

pos_picking_load/models/pos_order.py
```diff
@@ -0,0 +1,68 @@
...
+        for order in self:
+            # Cancel Picking
+            order.origin_picking_id.action_cancel()
+            order.origin_picking_id.write({'final_pos_order_id': order.id})
+
+            # Ignore Delivery exception of the Sale Order
+            sale_orders = sale_order_obj.search([
+                ('procurement_group_id', '=',
+                    order.origin_picking_id.group_id.id)])
+            sale_orders.signal_workflow('ship_corrected')
```
kirca, line 33: Workflows are deprecated and this has no effect. Maybe it would be suitable to call `sale_orders.action_done()` here instead?

pos_picking_load/static/src/xml/pos_picking_load.xml
```diff
@@ -0,0 +1,88 @@
...
+                        <i class="fa fa-arrow-circle-o-right"></i>
+                        Select
+                    </span>
+                    <span class="searchbox">
+                        <input placeholder="Search Picking" />
+                    </span>
+                </section>
+                <section class="full-content">
+                    <div class="window">
+                        <section class="subwindow collapsed">
```
kirca, line 40: This `section` doesn't seem to be needed in v10, it pushes the selection list down. I think it can be dropped.

kirca (changes requested):
Hi @legalsylvain ,

Great work! Please check my comments.

pos_picking_load/demo/sale_order.xml
legalsylvain: you mean, replacing by ``<odoo>`` ? indeed.

kirca: yes :)

pos_picking_load/static/src/js/widget.js
```diff
@@ -0,0 +1,289 @@
...
+                }).fail(function (error, event) {
+                    self.handle_errors(error, event);
+                });
+        },
+
+        render_list: function (pickings) {
+            var contents = this.$el[0].querySelector('.picking-list-contents');
+            contents.innerHTML = "";
+            var line_list = document.createDocumentFragment();
+            _.each(pickings, function (picking) {
```
hparfr, line 177: pickings.forEach ?

hparfr (approved):
Small remark about JS style.
Code read, didn't test.

I'm not convinced about the use cases. I think, the mix of sale order / pos order, cancellation, products not available in the pos, etc. will lead to edge cases hard to debug.

legalsylvain:
closing due to inactivity. module available for V12.

PR #321, opened 2019-01-19
[11.0] [MIG] pos_customer_display

rupa-acespritech:
POS Customer Display
====================

This module adds support for Customer Display in the Point of Sale. As incredible as it may seem, the Odoo POS doesn't have native support for a customer display, even on Odoo 10. So, if the customer cannot see the screen of the cashier, he will not be able to check the price of each product scanned by the cashier, he won't be able to see the total amount, etc. This module provides a solution to this problem by adding support for good old POS LCDs, which are often made of 2 lines of 20 caracters.

This module is designed to be installed on the *main Odoo server*. On the
*POSbox*, you should install the module *hw_customer_display*. But you will certainly prefer to use `pywebdriver <https://github.com/akretion/pywebdriver>`__ instead of the POSbox. Compared to the POSbox, Pywebdriver has several advantages:

* smaller footprint: no need to have a full-blown Odoo with PostgreSQL on the computer of the cashier (or his small Linux-based PC connected to the hardware, like the RaspberryPi for the POSbox),
* availability of an Ubuntu package, for easier deployment,
* native support for the customer display, payment terminal, etc.
* nice test/diagnosis Web interface.


Swapnesh-SerpentCS:
This PR was made for Training at [OCA Sprints India #3](https://odoo-community.org/event/oca-sprints-india-3-2019-01-19-98/register)

pedrobaeza:
Please sign the CLA and check Travis and Runbot status

chienandalu (changes requested):
Please, review runbot errors

legalsylvain:
@rupa-acespritech : any news ? 

thanks !

PR #322, opened 2019-01-23
[v10] fix apply pricelist when adding product to order.

Martronic-SA:
I've faced the problem that when I first choose the pricelist in a new order, I see the prices of procuts impacted accordingly, but when I add the product to the order, they have the base price and not the reduced one.
So i checked the javascript and found that on order line initialisation you are setting base product price or pricelist price.
The modification I've made takes the pricelist price if a pricelist is set on order, otherwise it will take the default product price.
I don't really understand you code:
`this.set_unit_price(`
`    options.product.price ||`
`    this.product.get_price(`
`        this.order.pricelist,`
`        this.get_quantity()`
`    )`
`);`
in which case the order line would take the pricelist price in the code above? if price is not set on product?

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -206,11 +206,13 @@ odoo.define("pos_pricelist.models", function (require) {
     models.Orderline.prototype.initialize = function (attr, options) {
         _Orderline_initialize.apply(this, arguments);
         if (options.product) {
-            options.price ||
-            this.product.get_price(
-                this.order.pricelist,
-                this.get_quantity()
-            )
+            this.set_unit_price(
+                options.price ||
+                this.product.get_price(
+                    this.order.pricelist,
+                    this.get_quantity()
+                )
+            );
         }
     };
     models.Orderline.prototype.init_from_JSON = function () {
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -206,13 +206,8 @@ odoo.define("pos_pricelist.models", function (require) {
     models.Orderline.prototype.initialize = function (attr, options) {
         _Orderline_initialize.apply(this, arguments);
         if (options.product) {
-            this.set_unit_price(
-                options.product.price ||
-                this.product.get_price(
-                    this.order.pricelist,
-                    this.get_quantity()
-                )
-            );
+            var unit_price = this.order.pricelist ? this.product.get_price(this.order.pricelist,this.get_quantity()) : options.product.price;
+            this.set_unit_price(unit_price);
         }
     };
     models.Orderline.prototype.init_from_JSON = function () {
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -206,8 +206,11 @@ odoo.define("pos_pricelist.models", function (require) {
     models.Orderline.prototype.initialize = function (attr, options) {
         _Orderline_initialize.apply(this, arguments);
         if (options.product) {
-            var unit_price = this.order.pricelist ? this.product.get_price(this.order.pricelist,this.get_quantity()) : options.product.price;
-            this.set_unit_price(unit_price);
+            options.price ||
+            this.product.get_price(
+                this.order.pricelist,
+                this.get_quantity()
+            )
         }
     };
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is the problem then also on v11 core (main code is from there)?

yajo:
Overall the patch makes sense, but just to be sure...

@Martronic-SA Could you please provide a screencast to let us see the problem live, and the fix?

@pedrobaeza Could you provide the core code link please?

Thanks!

Martronic-SA:
@Yajo I have no screen cast now but here is what happen:
If I open the POS, then add a product to the order and then change the pricelist, I see that the prices are impacted in the order and also in the products on the right. that's ok.

But if I choose the pricelist first, I see that the prices of the product are impacted in the product list on the right (ok), but then if I add one product to the order, the product in the order has the default price, and not the price shown in the product list on the right. see below:


I don't know if you can reproduce?

pedrobaeza:
@Yajo, you were the one that created `pos_pricelist` from v11, so I can't tell you where is the code you caught for making this extraction.

pos_pricelist/static/src/js/models.js
```diff
@@ -206,13 +206,8 @@ odoo.define("pos_pricelist.models", function (require) {
     models.Orderline.prototype.initialize = function (attr, options) {
         _Orderline_initialize.apply(this, arguments);
         if (options.product) {
-            this.set_unit_price(
-                options.product.price ||
```
yajo, line 210: OK, I have digged through it and the fix is to remove `.product` in this line. Could you please push the new patch?

yajo (changes requested):

pos_pricelist/static/src/js/models.js
Martronic-SA: you mean on line 210 replace
options.product.price ||
by
options.price ||
?
I just test the change and rewrite the commit.

yajo: Yes, I mean that. :blush:

pos_pricelist/static/src/js/models.js
```diff
@@ -207,7 +207,7 @@ odoo.define("pos_pricelist.models", function (require) {
         _Orderline_initialize.apply(this, arguments);
         if (options.product) {
             this.set_unit_price(
-                options.product.price ||
+                options.price ||
```
chienandalu, line 210: Why not to implement it the same way Odoo does upstream?: https://github.com/odoo/odoo/blob/11.0/addons/point_of_sale/static/src/js/models.js#L1350

chienandalu (changes requested):
Looks good as tested in runbot.

Just rebase you commits and bump version in manifest to 10.0.1.0.3

Minor and not blocking comment as well

pos_pricelist/static/src/js/models.js
yajo: It's actually what we're doing. Well, different code but same result, isn't it?

Martronic-SA:
@Yajo 
@chienandalu 
Finally it seems that the pricelist in pos is working well even without that piece of code!
I've commented the whole initialize function defined in pos_pricelist models.js under // Patch Ordreline and it is taking the right price wether I set the pricelist after or befor having added th product to the cart.
So is this really required?
Maybe I've missed something?

pos_pricelist/static/src/js/models.js
```diff
@@ -207,7 +207,7 @@ odoo.define("pos_pricelist.models", function (require) {
         _Orderline_initialize.apply(this, arguments);
```
yajo, line 207: Remove all of this initialize code, as you said, please.

yajo (changes requested):
Indeed you're probably right and this is not needed!

Look, in v10, [`initialize()` will call `this.set_quantity(1)`](https://github.com/odoo/odoo/blob/877fe07f7418e2a70256ec0f2127e8007604ba1d/addons/point_of_sale/static/src/js/models.js#L1103).

Further below in current file exists an override on `set_quantity()` that takes into account the pricelist. So, basically, all this `initialize()` is useless. Please remove it.
https://github.com/OCA/pos/blob/f6340cb00c73554ae4f55095b4273043afb64eb8/pos_pricelist/static/src/js/models.js#L222-L234

PR #323, opened 2019-01-23
[FIX] pos_order_return: return picking and invoices

chienandalu:
- Odoo already prepares a refund invoice in case negativa
- Return pickings weren't being auto validated as regular ones.

cc @Tecnativa

### DIFF
--- a/pos_order_return/models/pos_order.py
+++ b/pos_order_return/models/pos_order.py
@@ -48,11 +48,12 @@ class PosOrder(models.Model):
 
     def _prepare_invoice(self):
         res = super(PosOrder, self)._prepare_invoice()
-        if not self.returned_order_id:
+        if not self.returned_order_id.invoice_id:
             return res
         res.update({
-            'origin': self.name,
-            'type': 'out_refund',
+            'origin': self.returned_order_id.invoice_id.number,
+            'name': _(
+                'Return of %s' % self.returned_order_id.invoice_id.number),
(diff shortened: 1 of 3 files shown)


pos_order_return/models/pos_order.py
```diff
@@ -48,11 +48,10 @@ def _blank_refund(self, res):
 
     def _prepare_invoice(self):
         res = super(PosOrder, self)._prepare_invoice()
-        if not self.returned_order_id:
+        if not self.returned_order_id or not self.returned_order_id.invoice_id:
```
pedrobaeza: Isn't the same to put `if not self.returned_order_id.invoice_id:` directly?

chienandalu: Sure :)

chienandalu:
Test are still failing on pos_loyalty for this branch

pedrobaeza:
But is main branch also red?

rafaelbn (approved):
This PR is a FIX that work! :+1: 

Travis errors are not related to this PR so no reason for blocking it

rafaelbn:
After reviewing functionally:

- Description in the Credit Note should be Reason of return
- Source in the Credit note should be the number of the invoice refunded

@chienandalu

chienandalu:
@rafaelbn Fixed

pedrobaeza:
Travis is red

chienandalu:
Same with this one, @pedrobaeza

Issue #309, opened 2018-11-15
pos_loyalty tests failing in CI


kirca:

kirca:
https://github.com/OCA/pos/pull/308 eliminates possibility that it is related with timing. The output html in the logs after the step fails looks correct. Also I've tested successfully locally with same version of phantomjs (2.1.1).

legalsylvain:
which version ?

kirca:
11.0

chienandalu:
I'd say it's failing on integration with Odoo's tests. Can you try to add (like here: https://github.com/OCA/pos/blob/11.0/pos_order_return/tests/test_pos_order_return.py#L9):

```python
@common.at_install(False)
@common.post_install(True)
```

kirca:
Postponing tests till end doesn't affect the issue https://github.com/OCA/pos/pull/310

kirca:
Maybe CI maintainer or someone who has access can check more closely if phantomjs tests fail in general or it is specific to this module.

PR #326, opened 2019-02-12
[11.0][FIX] frontend tests

kirca:
Test `point_of_sale` in travis to check if there is issue with phantomjs. Related to https://github.com/OCA/pos/issues/309

### DIFF
--- a/pos_lot_selection/README.rst
+++ b/pos_lot_selection/README.rst
@@ -62,6 +62,7 @@ Contributors
 * David <david.vidal@tecnativa.com>
   * (https://www.tecnativa.com)
 * Ignacio Ales <ignacio.ales@guadaltech.es>
+* Kiril Vangelovski <kiril@lambda-is.com>
 
 Do not contact contributors directly about support or help with technical issues.
 
--- a/pos_lot_selection/static/src/js/chrome.js
+++ b/pos_lot_selection/static/src/js/chrome.js
@@ -1,4 +1,5 @@
 /* Copyright 2018 Tecnativa - David Vidal
+   Copyright 2019 Lambda IS DOOEL <https://www.lambda-is.com>
    License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). */
 
 odoo.define("pos_lot_selection.chrome", function (require) {
@@ -14,7 +15,7 @@ odoo.define("pos_lot_selection.chrome", function (require) {
             var events = {
                 "change .packlot-line-select": "lot_to_input",
             };
-            packlotline.events = Object.assign(
+            packlotline.events = _.extend(
                 packlotline.events, events
             );
             // Add methods over instanced popup


chienandalu:
@kirca I was testing this locally and found out that the problem seems to rely on the interaction with this part of `pos_lot_selection` code: https://github.com/OCA/pos/blob/11.0/pos_lot_selection/static/src/js/chrome.js#L10-L37

It's a part that injects some code into the instanced widget of lot input. I can't guess where the problem relies though. Yo can raise the error locally running the two modules tests altogether.

kirca:
@chienandalu , thanks. I'll check if I can reproduce it and if it is possible update the test case so that it doesn't break.

chienandalu:
Ok, let's keep this PR open to try to greenify the branch :wink:

kirca:
@chienandalu , it is finally :green_apple:

chienandalu (commented):
Was it en ES6 support thing from phantomjs all along?

kirca:
@chienandalu , that is right. It is explained in the commit message.

chienandalu (approved):

chienandalu:
Great! Thanks!

pos_lot_selection/__manifest__.py
```diff
@@ -6,6 +6,7 @@
     'version': '11.0.1.0.0',
     'category': 'Point of Sale',
     'author': 'Tecnativa,'
+              'Lambda IS,'
```
pedrobaeza: Sorry, but this change doesn't grant you co-authorship in the module. Please add yourself only in contributors.

pedrobaeza (changes requested):

yajo (approved):

pos_lot_selection/__manifest__.py
kirca: Sure. Just curious, what guidelines do you follow on what qualifies as a change for authorship?

pedrobaeza: There are no objective metrics, but a matter of sense about the amount of time/code involved around the module itself. In this case, it's very thankfully that you have fixed a problem in a module that is not yours, but the module essence and the amount of work invested for developing it is not something that you have needed to invest. Don't you think?

kirca: My opinion is that anyone who modifies a file is a co-author and has copyright on those files. If order is preserved (top to bottom) it can be seen who is original author. Qualifying based on amount of work done is quite subjective and if it will be used as a criteria we should define it in the [guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst), example _only improvements grant co-authorship and can be listed in the manifest, authors of bugfixes and other changes can be added in contributors_. I'll remove myself from the authors now although I did spend a surprising amount of time debugging and researching this :wink: .

pedrobaeza: One think is the copyright of the file, and another the module co-authorship. If you have a proposal for "objectivitizing" the module co-authorship in the guidelines, please do it. I'm all opened for avoiding these misunderstandings.

kirca: OK, I'll see if I can draft a proposal. Thanks for the review.

pedrobaeza (approved):

PR #327, opened 2019-02-12
[FIX] pos_order_mgmt: reprint tickets with fp

chienandalu:
When the PoS is using fiscal positions it can lead to errors when
computing the reprinted ticket value because the tax mapping function
computes them from the current order instead of from the one belonging
to the line.

So we set our loaded order temporarily as the selected one
just to bring it back to the current one after it's printed.

cc @Tecnativa

### DIFF
--- a/pos_order_mgmt/static/src/js/widgets.js
+++ b/pos_order_mgmt/static/src/js/widgets.js
@@ -37,6 +37,13 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
                 return this._super();
             }
             this.pos.from_loaded_order = false;
+            // When reprinting a loaded order we temporarily set it as the
+            // active one. When we get out from the printing screen, we set
+            // it back to the one that was active
+            if (this.pos.current_order) {
+                this.pos.set_order(this.pos.current_order);
+                this.pos.current_order = false;
+            }
             return this.gui.show_screen(this.gui.startup_screen);
         },
     });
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Isn't another way of doing this without the hack?

@Yajo you are more expert in JS than me, what do you think?

chienandalu:
The problem resides in this line: https://github.com/odoo/odoo/blob/11.0/addons/point_of_sale/static/src/js/models.js#L1684

There's no way to compute the taxes right if the loaded order is not temporarily selected

pedrobaeza:
Can't you do a trick like the one we use in Python having a context and overriding the `get_order` function to return another thing in that case?

chienandalu:
I could store the order in a global variable to be catched by get_order(), but I should disable it almost in the same way as I do here.

Maybe @Yajo knows of a more similar way python does it.

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -37,6 +37,10 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
                 return this._super();
             }
             this.pos.from_loaded_order = false;
+            if (this.pos.current_order) {
```
yajo, line 43: Please add a comment here explaining this code chunk.

yajo (approved):
The POS is designed to work with 1 SO per tab. This module itself is actually reusing one visual tab for more than 1 SO (although creating 1 in the background, but the user doesn't notice, except for the tab counter), so actually setting that one as the current one temporarily until the user exits this module's features and goes back to the normal POS mode is what makes most sense.

If the POS had a read-only mode, it would indeed make more sense to load a new tab with the chosen SO in read-only mode, and then print or return it from there, but that's not supported and it would mean refactoring some parts of the POS, and this module almost completely.

To make it more fun, most of this uses Backbone, which makes overrides even more complex.

So, after analyzing all of this together with @chienandalu, although we could make this digging down the innermost methods, it would end up with the same result, but would make the code more complex without real benefits, so I think current approach is better.

Just one little inline comment to make this behavior more obvious.

rafaelbn (approved):
OK thanks

rafaelbn:
Please before mergin wait to see bots :green_apple:

pedrobaeza:
Green bots

PR #328, opened 2019-02-27
[ADD] pos_require_product_quantity: raise popup if orderline quantity is 0

robinkeunen:
A popup is raised if product quantity is set to 0 for one of the order lines.

I'm new to javascript, I'm eager for comments. Most notably, I could not find a straightforward way to to go to the next line in the body of the popup.


chienandalu (commented):
Hi, @robinkeunen The same here. Rebase to get the last CI fixes

pos_require_product_quantity/static/src/js/pos.js
```diff
@@ -0,0 +1,59 @@
...
+        var _t = core._t;
+
+        screens.ActionpadWidget = screens.ActionpadWidget.include({
+            renderElement: function () {
+                var self = this;
+                this._super();
+
+                this.$('.pay').click(function () {
+
+                    if (self.pos.config.require_product_quantity) {
```
ivantodorovich, line 25: Two things here:

1) I guess the idea is to prevent the cashier from getting to the payment screen, too.. not just showing the popup.
For this I recommend doing it like this:

https://github.com/OCA/pos/blob/8eca935e1a9fe17976071d7ccf2084eac8654cfc/pos_event_sale/static/src/js/screens.js#L41-L75

This way you respect the original event handlers, which may have been altered by other modules.



2) It's better to avoid messing with the event callbacks if the setting is disabled, than evaluating in the callback.

pseudo-code:

```js
            renderElement: function () {
                if (!this.pos.config.require_product_quantity) {
                    return this._super.apply(this, arguments);
                }

                // your custom code after this..
            }
```

ivantodorovich (changes requested):

PR #330, opened 2019-02-27
[ADD] eslintrc file to have the possibility to mass eslint --fix JS code

legalsylvain:
Hi all,

There a lot of lint warning in this repository. Most of them are not a problem, but the mass of warning hides potential problem. I so like to clean Travis warning, using ``eslint --fix`` options.

(see logs : https://travis-ci.org/OCA/pos/jobs/493867890)

I propose to add the eslintrc file into the top of the repository.
Once done : 
-> install eslint package ``sudo npm install eslint -g``
-> run ``eslint --fix ./`` in any module subfolder, to fix all the lint warning automatically (and check all the "non automatically fixable" bugs to fix)

Thanks for your insights.

If it's OK, I'll do the same PR in all the branches > 9.0

In a second time, it will be possible to define custom configuration for lint checks.

CC : @chienandalu, @pedrobaeza, @kirca, @sbidoul

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+    "extends": "standard"
+};


kirca:
@legalsylvain , does `eslint` catch the same errors as the current JS linter in travis? Not sure if there is standard formatting rules in JS as is the case with Python PEP8 so could be that we will get again errors. But it seems like a nice tool and auto fix option sounds great :-), so maybe we can also suggest for it to be the default option since there is no specific one mentioned in the [guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#4javascript). This also means that it can be used in the CI too for consistency.

legalsylvain:
Hi @kirca. Thanks for your review.

> does `eslint` catch the same errors as the current JS linter in travis? 

i'm not sure. maybe @OCA/community-maintainers could answer to this question.

As far as I tested, the eslint check is more strict that the travis check. (In that direction, it's ok).

here is the extract of the eslint check. (all the things that could not be fixed automatically) : 
::

    20:1   error  'openerp' is not defined                                    no-undef
    21:3   error  Read-only global 'module' should not be modified            no-global-assign
    23:7   error  '_t' is assigned a value but never used                     no-unused-vars
    44:43  error  Expected '===' and instead saw '=='                         eqeqeq
    61:33  error  Identifier 'product_list' is not in camel case              camelcase
    74:60  error  Typeof comparisons should be to string literals          valid-typeof
    587:49  error  Unexpected newline between function name and paren        func-call-spacing
    271:2   error  Mixed spaces and tabs                                     no-mixed-spaces-and-tabs
    311:13  error  Unexpected tab character                                  no-tabs

Some checks are maybe a matter of taste (``camelcase``), but I thinks it's good to fix ``no-undef``, ``no-global-assign``, ``no-unused-vars``, etc...


I'm not a lint expert and I did not benchmarked all the JS checks solutions. This is an intend to reduce the problem of huge warning logs, and provide a tools for all the contributors in OCA/pos (and also OCA/web I guess).
- After merging, it's quite easy to ask to contributors to install eslint and to run it.
- Also it is maybe possible to run it by bot, after each merge. (like readme generation) ? CC : @sbidoul 

here is the current travis log of the branch 9.0 : https://travis-ci.org/OCA/pos/jobs/465595716
``Your code has been rated at -2.77/10``
(yes, negative notation seems to be possible !)

Once ran, here the new log (https://github.com/OCA/pos/pull/331) : https://travis-ci.org/OCA/pos/jobs/499417491
``Your code has been rated at 9.33/10``

I see it easier to review and a good first step. Don't you think ?

regards.

kirca:
Hi @legalsylvain ,

I agree. Lets try it for OCA/pos first and suggest to use it for all PRs. We can add this info in the root README. Once we see how best to use it we can propose it for other projects also.

kirca:
OCA/pylint-odoo seems to be the linting tool that runs on travis, and it already uses eslint for JS https://github.com/OCA/pylint-odoo/pull/97 .

chienandalu:
Indeed eslint is already available (example in a 11.0 CI log: https://travis-ci.org/OCA/pos/jobs/493671209). This are the rules configured : https://github.com/OCA/pylint-odoo/blob/master/pylint_odoo/examples/.jslintrc

Do you miss something?

Issue #334, opened 2019-03-14
11.0 -> 12.0 pos_order_mgmt splitting in several module


PierrickBrun:
I will soon migrate pos_order_mgmt from 11.0 to 12.0.

I think this is a good occasion to split it it in several modules:

- `pos_order_mgmt` to display old pos_orders
- `pos_order_mgmt_reprint` to reprint tickets
- `pos_order_mgmt_return` to return orders

That way we can only install what we need in our project.

Also that could be an occasion to integrate pos_order_return into pos_order_mgmt_return.

I'm not sure  that I will need `pos_order_mgmt_return` so I may not migrate that part

Any thoughts about this @chienandalu @legalsylvain ?

chienandalu:
It can be interesting. There are some monolithic methods that should be considered to make them extensible, mainly this one: https://github.com/OCA/pos/blob/11.0/pos_order_mgmt/static/src/js/widgets.js#L184

Or this one: https://github.com/OCA/pos/blob/11.0/pos_order_mgmt/static/src/js/widgets.js#L262

chienandalu:
I close this as the module migration is alredy on the go

Issue #333, opened 2019-03-13
[BUG] [11.0] [pos_order_mgmt] Incompatibility with pos_restaurant


carlos-domatix:
Hello,

- Go to OCA runbot / pos.
- Install pos_restaurant.
- Open session on pos.
- Click on cart button to return or view orders.
- You will get this javascript error:
![Screen Shot on Mar 13th at 11_15 AM](https://user-images.githubusercontent.com/28001484/54271192-5921dd00-4581-11e9-8663-6e727f4840b4.png)

Other bugs detected:
- Open session on pos.
- Select any table.
- Click on cart button to return or view orders.
- Click on back button.
- When you click on the previously selected table, it does not open.

chienandalu:
Hi, @carlosDomatix Thanks for you report.

Indeed we didn't count with the restaurant screens and that behaviour and it needs to be fixed.

- For the first case, the cart icon should be hidden in the table selection screen.
- The second one seems a little bit more complex, since there's some wrong interaction the table widget.

Feel free to PR with a fix :smile:

PierrickBrun:
Hi @chienandalu,
I've fixed the first problem in this PR https://github.com/OCA/pos/pull/337 . Fortunately, I did not have to hide the button on the floors screen.
The second problem is, as you said, more complex and I still don't know how to fix it

PR #337, opened 2019-03-21
[MIG][12.0] port pos_order_mgmt

PierrickBrun:
With modules split as discussed in #334


pedrobaeza:
A migration script will be needed that installs the 2 extra modules.

pos_order_mgmt/readme/USAGE.rst
```diff
@@ -0,0 +1,19 @@
+Once the PoS is loaded, you'll find a shopping trolley icon (🛒) in the top
+bar that grants access to the order list screen.
+
+.. image:: /pos_order_mgmt/static/description/order-mgmt-icon.png
```
legalsylvain: replace by 
``.. image:: ../static/description/order-mgmt-icon.png``
With that syntax, 

- it will work locally in github interface if you see the USAGE.rst file in your browser
- and the main readme of the module will be correctly generated (ref : https://github.com/OCA/maintainer-tools/pull/397)

PierrickBrun:
after trying to split it into several modules, it appeared to be more time consuming than I first thought.

I've migrated the module without splitting but I modified the pos.config to allow to choose to allow reprint or returning done orders.

I think the returning part is working well but I' don't know the 11.0 module and pos_order_return enough to be sure about it.

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -151,34 +151,40 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
...
-                        orderlines: order.get_orderlines(),
-                        paymentlines: order.get_paymentlines(),
-                    }));
-                this.pos.set_order(this.pos.current_order);
-                this.pos.current_order = false;
-            } else {
-                this.pos.reloaded_order = order;
-                this.gui.show_screen('receipt');
-                this.pos.reloaded_order = false;
+            //this is only a double check, the button is invisible
```
legalsylvain: why adding a double check, if the button is invisible ?

legalsylvain (commented):
> after trying to split it into several modules, it appeared to be more time consuming than I first thought.

> I've migrated the module without splitting but I modified the pos.config to allow to choose to allow reprint or returning done orders.

Thanks @PierrickBrun for porting this module.

> I think the returning part is working well but I' don't know the 11.0 module and pos_order_return enough to be sure about it.

Well, ``pos_order_return`` was developed to 
- link returned order to the original order, (what your code does)
- to block the possibility  to return more quantity as the original one. This feature was developped only in backoffice part and had to be implemented in the front office.
As in new version of Odoo PoS, it is not possible to make an order in a backoffice part, we could consider ``pos_order_return`` as fully obsolete on the second point.

2. Did you checked if the module is compatible with ``pos_restaurant`` ? ref : https://github.com/OCA/pos/issues/333. CC @chienandalu, @carlosDomatix 

kind regards.

other wise, LGTM. (code review / no test).

PierrickBrun:
Looks like it works better with pos_restaurant now

fuentes73 (approved):
Tested and working OK

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -0,0 +1,404 @@
...
+            });
+            return order;
+        },
+        _prepare_orderlines_from_order_data: function(order, order_data, orderLines) {
+            _.each(orderLines, function(line) {
+                // In case of local data
+                if (line.length === 3) {
+                    line = line[2];
+                }
+                var product = self.pos.db.get_product_by_id(line.product_id);
```
fuentes73, line 303: Is missing at the beginning `var self = this;`
So this line throws the error: `Cannot read property 'db' of undefined`

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -0,0 +1,404 @@
...
+            var orderLines = order_data.line_ids || order_data.lines || [];
+            self._prepare_orderlines_from_order_data(order, order_data, orderLines);
+            if (order_data.return) {
+                order.return = true;
+                // A credit note should be emited if there was an invoice
+                order.set_to_invoice(order_data.origin_invoice_id);
+                // We'll refunded orders once they are synced
+                order.returned_order_id = order_data.id || order_data.name;
+                order.origin_name = order_data.pos_reference || order.returned_order_id;
+                return order;
```
fuentes73: Testing more in depth the module I also seen that, if you return and order and reload the page without change anything, the "return" info is lost.
return-pos-order


A possible solution might be to add `order.trigger('change');` before return the order
```suggestion
                order.trigger('change');
                return order;
```

PierrickBrun:
thank you for the review @fuentes010 , I've addressed the changes

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -0,0 +1,406 @@
...
+                if (self.unknown_products.length > 0) {
+                    self.gui.show_popup('error-traceback', {
+                        'title': _t('Unknown Products'),
+                        'body': _t('Unable to load some order lines because the ' +
+                            'products are not available in the POS cache.\n\n' +
+                            'Please check that lines :\n\n  * ') + self.unknown_products.join("; \n  *"),
+                    });
+                    correct_order_print = false;
+                }
+                if (correct_order_print && action === 'print') {
```
Kerrim-: Hi,

I'm using this code for a project and wanted to ask if it is possible to change this part to be a little easier to extend. For example by putting it in some kind of dispatch method or like this:

```suggestion
                if(correct_order_print){
                    self["action_" + action](order);
                }
```

**Pros**: Easier to extend (You only need to register the event and write the method)
**Cons**: If you need different checks before executing the action or you want to use actions not starting with action this suggestion won't work. (But the dispatch method still would.)

Kerrim-: Also if there is more work to do with this module I would be glad to help :)

pedrobaeza:
Please add #354 here

pos_order_mgmt/static/src/js/widgets.js
PierrickBrun: Hi @Kerrim- , No problem about accepting a PR from my side :+1: I only use the module for reprinting tickets so I did not dig a lot into returns

ivantodorovich:
Functional test only.

I only encounter this issue (very small one):
On 11.0. the order list window closes automatically when clicking the refund order button.

**12.0:**
9Lx4sxJ4Rw


**11.0:**
H4VJPthjQu

ivantodorovich:
It's probably because of this line:
https://github.com/OCA/pos/pull/337/commits/e548fccd518bf2653723a1107d058b50a41dcdb5?file-filters%5B%5D=.js&owned-by%5B%5D=#diff-3619c2549c163cd99597081619ddb2c4R273

PierrickBrun:
@ivantodorovich nice catch :+1:

ivantodorovich (approved):

chienandalu:
Hi, @PierrickBrun , maybe you want to take a look at this: https://github.com/OCA/pos/pull/361

pos_order_mgmt/models/pos_order.py
```diff
@@ -0,0 +1,131 @@
...
+        readonly=True,
+    )
+    refund_order_ids = fields.One2many(
+        comodel_name='pos.order',
+        inverse_name='returned_order_id',
+        string='Refund Orders',
+        readonly=True,
+    )
+    refund_order_qty = fields.Integer(
+        compute='_compute_refund_order_qty',
```
legalsylvain, line 29: Where is this function defined ?

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -0,0 +1,450 @@
...
+            var order_data;
+            var dataset = event.target.parentNode.dataset;
+            if (dataset.orderId) {
+                self.load_order_data(parseInt(dataset.orderId, 10))
+                .then(function(order_data) {
+                    self.order_action(order_data, action);
+                })
+            } else {
+                _.each(this.orders, function (order_data) {
+                    if (order.uid === dataset.uid) {
```
legalsylvain: I don't understand in which case dataset.orderId is not defined, and order is not defined anywhere.

legalsylvain (changes requested):
Hi @PierrickBrun. 

* some minor errors inline. 
* there are also a lot of lint warning on Travis. 
* Coverage is low because there is no python test
* Adding a ribbon is cool, but generate a display bug in the total place : 
image




I propose to fix all that points this week, in a PR against your branch.

Otherwise, thanks a lot for this work ! 
kind regards.

pos_order_mgmt/models/pos_config.py
```diff
@@ -0,0 +1,40 @@
...
+        oldname='iface_load_done_order',
+    )
+
+    iface_copy_done_order = fields.Boolean(
+        string='Duplicate Done Orders',
+        default=True,
+        help='Allows to duplicate already done orders in the frontend',
+    )
+
+    iface_load_done_order_max_qty = fields.Integer(
```
legalsylvain: This field is not used anywhere.

pos_order_mgmt/models/pos_config.py
```diff
@@ -0,0 +1,40 @@
...
+        default=True,
+        help='Allows to reprint already done orders in the frontend',
+        oldname='iface_load_done_order',
+    )
+
+    iface_return_done_order = fields.Boolean(
+        string='Return Done Orders',
+        default=True,
+        help='Allows to return already done orders in the frontend',
+        oldname='iface_load_done_order',
```
legalsylvain: there are two field with the same oldname value. 'iface_load_done_order'. why ?

legalsylvain:
Hello, I finished the port of this module, fixing some points, and adding test. (coverage from 43% to 71%).
I did a PR against this PR : https://github.com/OCA/pos/pull/373
You can test it on the runbot of the PR.

regards.

pos_order_mgmt/models/pos_config.py
PierrickBrun: the field iface_load_done_order is splitted, so if you used it, you want both the functions once migrated

legalsylvain (approved):

legalsylvain:
@fuentes010, @ivantodorovich : various changes since your last review. do you want to update your review ? 
thanks.

rafaelbn (changes requested):
Please don't merge until @chienandalu  make last review tomorrow, thanks!

chienandalu (approved):
Tested on runbot and code review. Thanks for the good work!

legalsylvain:
Hi @chienandalu, thanks for your review.

/ocabot merge

PR #340, opened 2019-03-27
[8.0][FIX] pos_pricelist

legalsylvain:
protect pricelist_id get to avoid error in tests if other modules if pos_pricelist is installed


legalsylvain:
Trivial patch. Merging. Travis failed for unrelated reason.

PR #341, opened 2019-04-02
[FIX] pos_order_return: ensure test correctness

chienandalu:
- In a CI test environment, we could end up with pricelists with different
currencies, wich leads to exceptions. This commit ensures the test.

cc @Tecnativa


pedrobaeza (approved):

yajo (approved):

chienandalu:
Thanks :)

Fast tracking this

PR #343, opened 2019-04-25
[9.0] [MIG] POS Restricted Customer List

ThomasBinsfeld:
Migration of https://github.com/OCA/pos/tree/8.0/pos_restricted_customer_list.


rousseldenis (approved):
Code review

rousseldenis:
Travis is red for lint errors related to other modules

pos_restricted_customer_list/__openerp__.py
```diff
@@ -1,20 +1,24 @@
...
-    'license': 'AGPL-3',
+    'name': "POS: restricted customer list",
+    'version': '9.0.1.0.0',
+    'development_status': "Production/Stable",
+    'category': "Point Of Sale"'',
+    'summary': "This module will limit the download of customer data to "
+                "only those customers where this has been specifically "
+                "requested.",
+    'author': "Therp BV, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/pos/tree/9.0/"
```
chienandalu, line 14: https://github.com/OCA/pos should be enough

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
```diff
@@ -1,25 +1,25 @@
...
-                    return true; // domain already added
+    PosModels.PosModel = PosModel.extend({
+        initialize: function (session, attributes) {
+            var self = this;
+            for (var i = 0; i < self.models.length; i++) {
+                var model = self.models[i];
+                var model_name = model.model;
+
+                if (model_name === 'res.partner') {
+                    model.domain = [
```
chienandalu: what if another module tries to extend the domain?

```javascript
model.domain = model.domain.push(['available_in_pos','=',true]);
```

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
```diff
@@ -1,25 +1,25 @@
...
-			if (m.model !== 'res.partner')
-				return false;
-			// check if not already done by someone else
-            for(var i = 0; i < m.domain.length; i++) {
-                var domain_tuple = m.domain[i];
-                if (domain_tuple[0] === 'available_in_pos') {
-                    return true; // domain already added
+    PosModels.PosModel = PosModel.extend({
+        initialize: function (session, attributes) {
+            var self = this;
```
chienandalu, line 11: I'd say this isn't needed

pos_restricted_customer_list/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+
+from openerp import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    available_in_pos = fields.Boolean(
+        string="Available for POS",
+        default=False,
```
chienandalu, line 13: Shouldn't this be `True`? What happens when a cashier creates a customer from the frontend, does it become unavailable later?

chienandalu (changes requested):
I couldn't properly test it on runbot, as some other modules in this branch seem to have bad interactions. Here are some comments anyway:

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
ThomasBinsfeld: I prefer to always write `var self = this;` in case it's useful in the future. It doesn't cost anything.

pos_restricted_customer_list/models/res_partner.py
ThomasBinsfeld: There was no default value in the v8 module (so it was `False`), i just made it explicit here. I prefer not to change the behavior in the migration PR.

chienandalu: Ok, but, what happens in the case I mentioned?

pos_restricted_customer_list/__openerp__.py
ThomasBinsfeld: It would be enough, indeed. I just chose the other way suggested by the [OCA guidelines](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11).

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
ThomasBinsfeld: That doesn't work.

chienandalu: Well, try an alternative, but you shouldn't substitute the domain, but extend it (as the original module does)

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
chienandalu: Well, you don't put code for the just in case :sweat_smile: If it's useless now, remove it, as it makes it easier to read.

pos_restricted_customer_list/__openerp__.py
chienandalu: Ok :+1: I prefer the shorter anyway as it's a line you don't have to worry about in future migrations

benwillig (approved):
Maybe there could be a ```_get_partner_domain``` method but otherwise LGTM :+1:

pos_restricted_customer_list/models/res_partner.py
ThomasBinsfeld: It will be available in the POS where the cashier created it until it's closed.

chienandalu: Until that pos is reloaded. Imagine I create a customer for a ticket and... oops, I accidentally pressed F5... now I don't have the customer anymore and probably an error raises for the ticket I had opened

ThomasBinsfeld: It disappears only after closing and posting the entries so no error. But I guess it's not intuitive. I would say that the `default` in the field's definition must remain `False` but `available_in_pos` must be set to `True` when creating a partner from a POS. Wyt?

ThomasBinsfeld: Done

rousseldenis:
9.0 branch is red as Travis could not launch

chienandalu:
@ThomasBinsfeld I fixed 9.0 travis, so you can rebase to get green travis

ThomasBinsfeld:
> @ThomasBinsfeld I fixed 9.0 travis, so you can rebase to get green travis

@chienandalu Thanks a lot! Rebased.

chienandalu (approved):

PR #344, opened 2019-05-07
[MIG] 12.0 pos_loyalty

PierrickBrun:

PR #356, opened 2019-05-29
12.0 mig pos loyalty

cvinh:


chienandalu (commented):
Thanks @cvinh , but this module is already being migrated and discussed in https://github.com/OCA/pos/pull/344

PR #358, opened 2019-05-29
[10.0][FIX] pos_pricelist: Wrong comparison

fuentes73:
`self.product_tmpl_id` is a tuple `(id, name)` so the comparison is always wrong

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -142,7 +142,7 @@ odoo.define("pos_pricelist.models", function (require) {
             }
 
             var pricelist_items = _.filter(pricelist.items, function (item) {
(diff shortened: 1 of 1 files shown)


chienandalu (commented):
This is backported from Odoo v11 itself, so there would be a bug there as well (https://github.com/odoo/odoo/blob/11.0/addons/point_of_sale/static/src/js/models.js#L1282)

Can you give some steps to reproduce the issue?

(I'm fixing the branch .travis here: https://github.com/OCA/pos/pull/359)

chienandalu:
Hi, @fuentes010 . Please, rebase your branch to get the last CI changes and get green travis

fuentes73:
@chienandalu thank's for the review.
I've tested on runbot and cannot reproduce the error.
I guess there is something wrong on our side.

PR #360, opened 2019-06-04
12.0 mig pos margin

wpichler:
Nothing changed beside the manifest...


pos_margin/i18n/es.po
```diff
@@ -35,7 +35,8 @@ msgstr ""
 
 #. module: pos_margin
 #: model:ir.model,name:pos_margin.model_pos_order_line
-msgid "Lines of Point of Sale"
+#, fuzzy
```
legalsylvain, line 38: I think there is some errors in the translation file. there are a lot of "fuzzy".

pos_margin/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (C) 2017 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'POS Margin',
+    'version': '11.0.1.0.0',
```
legalsylvain: Should be 12.0.1.0.0

pos_margin/models/pos_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+        multi='multi_margin', digits=dp.get_precision('Product Price'))
+
+    # Compute Section
+    @api.multi
+    @api.depends('product_id', 'qty', 'price_subtotal')
+    def _compute_multi_margin(self):
+        for line in self:
+            if not line.product_id:
+                line.purchase_price = 0
+                line.margin = 0
```
legalsylvain, line 28: ``for line in self.filtered(lambda x: x.product_id):``

can replace line 25-> 29.

legalsylvain:
some minor remarks. LGTM otherwise.

thanks for porting this module.

travis is red because the version is not correct. (should be 12).

pos_margin/models/pos_order_line.py
chienandalu: You don't even need the lambda :slightly_smiling_face: :
```python
for line in self.filtered('product_id'):
```

pos_margin/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'category': 'Point Of Sale',
```
chienandalu, line 8: the website key is missing as well

pos_margin/models/pos_order_line.py
legalsylvain: ow thanks !

wpichler: this is correct for this small piece of the code - but there is also an "else" condition in the original code. So i would have to create to for loops here - one filtered for all lines with product_id - one filtered with all lines without product_id...

I think its ok as it is

pos_margin/__manifest__.py
wpichler: ok

pos_margin/__manifest__.py
wpichler: added the website key now

pos_margin/models/pos_order_line.py
chienandalu: I didn't notice that :)
An alternative:
```python
self.filtered(lambda x: not x.product_id).write({
    'purchase_price': 0,
    'margin': 0,
})
for line in self.filtered('product_id'):
    line.purchase_price = line.product_id.standard_price
    line.margin = line.price_subtotal - (
        line.product_id.standard_price * line.qty)
```

chienandalu: Anyway, it can stay as it is :)

pos_margin/models/pos_order_line.py
```diff
@@ -0,0 +1,32 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+import openerp.addons.decimal_precision as dp
+
+
+class PosOrderLine(models.Model):
+    _inherit = 'pos.order.line'
+
+    # Columns Section
```
chienandalu, line 12: Maybe this obvious comments can go away ;)

pos_margin/models/pos_order_line.py
legalsylvain: I don't think so. I think setting default values is not necessary.
this part of code should work.

```
for line in self.filtered('product_id'):
    line.purchase_price = line.product_id.standard_price
    line.margin = line.price_subtotal - (
        line.product_id.standard_price * line.qty)
```

pos_margin/models/pos_order_line.py
legalsylvain: ;-) Indeed. My bad.

pos_margin/models/pos_order_line.py
wpichler: i think setting default is necessary - else it will be None and not zero - this could cause problems in other calculations. So i do tend to keep the original code.
And - there is no real computation cost here - as the result gets stored - so no problem.

pos_margin/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'POS Margin',
+    'version': '12.0.1.0.0',
+    'category': 'Point Of Sale',
+<<<<<<< HEAD
+    'website': 'http://www.github.com/OCA/pos',
+=======
+>>>>>>> 5107be1... [FIX] fixes margin calculation with different uom
```
chienandalu: You've got serveral merge conflict in your last commit

pos_margin/models/pos_order.py
```diff
@@ -0,0 +1,37 @@
...
+        for line in self.env['pos.order.line'].search([]):
+            line.env.add_todo(line._fields['purchase_price'], line)
+            line.env.add_todo(line._fields['margin'], line)
+            line.recompute()
+
+    @api.multi
+    def recompute_order_margins(self):
+        for order in self.env['pos.order'].search([]):
+            order.env.add_todo(order._fields['margin'], order)
+            order.recompute()
```
legalsylvain, line 37: I don't think that the two buttons are correct. Recompute margin after the initial computation will corrupt data, as the standard price changed, etc...

legalsylvain (changes requested):

pos_margin/models/pos_order_line.py
```diff
@@ -0,0 +1,35 @@
...
+    @api.depends('product_id', 'qty', 'price_subtotal')
+    def _compute_multi_margin(self):
+        for line in self:
+            if not line.product_id:
+                line.purchase_price = 0
+                line.margin = 0
+            else:
+                # if used in combination with module which does add the uom_id to line
+                if hasattr(line, 'uom_id') and line.uom_id.id != line.product_id.uom_po_id.id:
+                    line.purchase_price = line.product_id.uom_po_id._compute_price(line.product_id.standard_price, line.uom_id)
```
legalsylvain, line 32: should not be uom_po_id, but uom_id by default.

legalsylvain:
Hi  @wpichler. I allowed me to finish the migration of pos_margin in that PR : https://github.com/OCA/pos/pull/385

I kept the lazy dependency to ``pos_order_line.uom_id`` if the field exist, and and add you as co-author.
added also test, and handle multi currency context, as in the sale_margin module.

So closing this PR. If there is any problem with that, please reopen it.
otherwise, your review is welcome in #385. 
kind regards.

PR #361, opened 2019-06-19
[FIX] pos_order_mgmt: loaded orders date to locale

chienandalu:
- Loaded orders where in UTC

cc @Tecnativa

### DIFF
--- a/pos_order_mgmt/static/src/js/widgets.js
+++ b/pos_order_mgmt/static/src/js/widgets.js
@@ -316,6 +316,13 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
                 args: [query || '', this.pos.pos_session.id],
             }).then(function (result) {
                 self.orders = result;
+                // Get the date in local time
+                _.each(self.orders, function (order) {
+                    if (order.date_order) {
+                        order.date_order = moment.utc(order.date_order)
+                            .local().format('YYYY-MM-DD HH:mm:ss');
+                    }
+                });
             }).fail(function (error, event) {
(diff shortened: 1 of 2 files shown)


pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -316,6 +316,14 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
                 args: [query || '', this.pos.pos_session.id],
             }).then(function (result) {
                 self.orders = result;
+                // Get the date in local time
+                _.each(self.orders, function (order) {
+                    if (order.date_order) {
+                        var date = moment.utc(order.date_order).toDate();
+                        order.date_order = moment(date).format(
+                            'YYYY-MM-DD HH:mm:ss');
```
Tardo: Here can use ```moment.utc(order.date_order).local().format('YYYY-MM-DD HH:mm:ss')```

Tardo (approved):
LGTM, code review :+1:

pos_order_mgmt/static/src/js/widgets.js
chienandalu: Thanks @Tardo I put it as you said :)

pedrobaeza:
I would say Odoo has already functions for doing this directly (or why don't you pass tz on RPC for getting the date from server directly in proper timezone?)

chienandalu:
I'd pass a context if I had any, but I don't

pedrobaeza (approved):
OK then

PR #362, opened 2019-06-24
[11.0] New module sales & pos sales report

Martronic-SA:
This new module "merges" sales report and pos sales report under the standard sales report.
Technically it creates two more postgresql views, the first one that unifies sale.order and pos.order, the second one that unifies sale.order.line and pos.order.line.
Then the standard sale.report model is altered to get the lines data from sales unified and sales lines unified.
In addition, the "Sales" button on the product will show the number of sales in orders and pos ordes cumulated and when you click on it you will get the list of sales lines unified.
sales_stats


pedrobaeza:
It seems there's a problem with your branch. Any way, isn't is this what `pos_sale` module does?

Martronic-SA:
@pedrobaeza 
This is a bit different:
The sales count on product is based on sale.report model. As long as the module adds pos sales in sale.report, the sales count on products automatically reflects the pos & sales counts.
The Sales button on product links to the sale.report model with links to the corresponding Orders or Sales Orders: here below on the left you have the links to the SO and the Main/ which are POS orders.

And finally the Sales report has more data  than the pos_sale report which is mainly based on sales teams.

pedrobaeza:
OK, please reflect this on the README and you should clean the branch, as 149 modified files are not correct.

rafaelbn:
Hi @Martronic-SA , Odoo makes this in Odoo v12 maybe you could take a look into v12 and make the barkport? Is available in Odoo Community. Regards

Martronic-SA:
Ok, I didn't know about v 12.0 module. Therefore that module is not very useful here as long as it will become obsolete with v12.0

chienandalu:
Well, indeed it has pos_sale has been present since 11.0 :slightly_smiling_face:

Martronic-SA:
@chienandalu  yes but honestly in v11 this is not really satisfying. they whouldn't have completely rewritten it otherwise.

PR #363, opened 2019-06-25
[12.0][MIG] pos_session_pay_invoice

jarroyomorales:
Migration to 12.0 of pos_session_pay_invoice. No major changes.
Added the dependency of https://github.com/OCA/account-payment/pull/240 in `oca_dependencies.txt`


pos_session_pay_invoice/readme/INSTALL.rst
```diff
@@ -0,0 +1,2 @@
+* In order to install this module, you will need to install first the
```
legalsylvain: I think this file is not necessary, as it is redondant with manifest.

legalsylvain (approved):
a minor remarks. Otherwise LGTM ! Thanks for porting this usefull module.

tested on runbot. :+1:

pos_session_pay_invoice/readme/INSTALL.rst
jarroyomorales: yes, you are right, I just removed it.

oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+account-payment https://github.com/mamcode/account-payment 12.0-mig-account_cash_invoice
```
chienandalu: Let's wait for https://github.com/OCA/account-payment/pull/240 to be finished and merged, otherwise it could be overseed since the changes left aren't very big

etobella (approved):
@jarroyomorales can you change dependencies for the new PR OCA/account-payment#276

jacobus101:
@jarroyomorales don't forget this module please!

jarroyomorales:
@jacobus101 @chienandalu Sorry! It's green now :green_apple:

jacobus101:
Thanks @jarroyomorales this community is so cool!

jacobus101:
@etobella @chienandalu @legalsylvain guys don't forget this one please!

legalsylvain:
I'm asking myself if this module could be bad used if user make try to make payment on the account.invoice, in the back-office, when the session is not closed.

So, I'm thinking to make a dependency with the module [pos_invoicing](https://github.com/OCA/pos/tree/12.0/pos_invoicing) to avoid that case and to show explicitely to the user, that the invoice will be marked as paid, when the session will be closed.

Any point of view about that point ?

etobella:
@legalsylvain The system will not allow you to pay it twice, It will fail and give a warning.
For example, if you use this and then you reconcile it manually, the sessions will not be able to be closed, as the invoice has already been reconciled. 
The option you propose is good as an initial option, but what happen if we use two payments?
Maybe the best option would be to show the proposed reconciliations on the invoice, but it is hard to achieve.

legalsylvain:
>  The option you propose is good as an initial option, but what happen if we use two payments?

Thanks for you answer !

What about to change the module ``pos_invoicing`` changing the field ``has_pending_payment`` boolean, by a float ``pos_pending_payment_amount`` ? And so prevent "to much payment".
It will take all the cases, and will avoid such mistakes. Don't you think ?

etobella:
Probably it is better, however, I have a doubt, `pos_invoicing` should depend on this or this should depend on `pos_invoicing`?

legalsylvain:
``pos_session_pay_invoice`` should depend on ``pos_invoicing`` but it requires a refactoring of pos_invoicing. Could you just add a section in the roadmap. We'll add the improvment when porting there modules in v 13+.

jarroyomorales:
Okay, I just added it to the roadmap. @legalsylvain @etobella

legalsylvain:
Finally, I'll try to make the changes on your PR V12. Expect a PR in the next week.

flotho:
@legalsylvain , didi you push your PR , do you need review ?

jarroyomorales:
@legalsylvain @flotho Maybe we can proceed merging this and then do the PR?

etobella:
@OCA/pos-maintainers I think this is ready for merging

legalsylvain:
Hi. sorry for the late. I don't see easy way to make a dependency between both modules. let's go ahead, and we'll see later to improve this module.

/ocabot merge patch

PR #368, opened 2019-07-07
[12.0][MIG] pos_to_weight_by_product_uom

legalsylvain:
- Migrate pos_to_weight_by_product_uom from 8.0 to 12.0. 

main changes : 
- renaming model and files, to the change related to the new ``uom`` module
- create readme files
- update images

- commit to review : https://github.com/OCA/pos/commit/2ba5c3c728cb787481512e63578ed934bff23917

CC : @hparfr, @quentinDupont


pos_to_weight_by_product_uom/models/uom_uom.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2017, Grap
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class UomtUom(models.Model):
```
hparfr: why "t" in class name ?

pos_to_weight_by_product_uom/models/product_template.py
```diff
@@ -0,0 +1,15 @@
...
+# @author: Quentin DUPONT <quentin.dupont@grap.coop>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.onchange('uom_id')
+    def _onchange_uom_id(self):
```
hparfr, line 11: Why onchange instead of api.depends or related field ?

hparfr (approved):
Some no blocking remarks about the design.

pos_to_weight_by_product_uom/models/product_template.py
legalsylvain: The objective of these feature is to propose a value for 'to_weight', when user is creating a product. not to force it. Maybe some products can be with unit 'Kilo', but will be only sold by 'kilo', without the possibility to weight it.

pos_to_weight_by_product_uom/models/uom_uom.py
legalsylvain: typo fixed. Thanks !

legalsylvain:
@hpar : answered and fixed. Is it OK ? (or happy holidays, if don't answer this month)

hparfr:
I did approve it.

legalsylvain:
oh sorry, did'nt seen.

houssine78:
@legalsylvain what is the difference with the pos_to_weight module ?

legalsylvain:
@houssine78 which module ?

houssine78:
@legalsylvain this one https://github.com/AwesomeFoodCoops/odoo-production/tree/9.0/extra_addons/pos_price_to_weight

legalsylvain:
This module set the to_weight field, regarding the uom.
The other compute in the PoS product price from barcode.

houssine78:
ok ! sorry, I though that the module has been renamed.

anyway do you think that pos_price_to_weight could come under OCA/pos umbrella ?

legalsylvain:
> anyway do you think that pos_price_to_weight could come under OCA/pos umbrella ?

Of course ! Feel free to propose a PR, I'll review it for sure.

Regarding this module, did you reviewed it ? 

regards.

PR #371, opened 2019-07-12
[8.0][FIX] pos_payment_terminal

azmeuk:
pos_payment_terminal is supposed to work with [hw_telium_payment_terminal](https://github.com/OCA/pos/blob/8.0/hw_telium_payment_terminal/), but the telium controller [expects to be told](https://github.com/OCA/pos/blob/8.0/hw_telium_payment_terminal/controllers/main.py#L132) the number of decimals of the currency used for the transaction.

This pull requests make pos_payment_terminal communicate the numbers of decimals to the posbox proxy.

I suppose it should also be ported to branch 9.0?

### DIFF
--- a/pos_payment_terminal/static/src/js/pos_payment_terminal.js
+++ b/pos_payment_terminal/static/src/js/pos_payment_terminal.js
@@ -11,9 +11,10 @@ openerp.pos_payment_terminal = function(instance){
     module = instance.point_of_sale;
 
     module.ProxyDevice = module.ProxyDevice.extend({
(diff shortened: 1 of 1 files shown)


legalsylvain:
LGTM. Thanks. Fast tracking this patch.

FYI, authors of ``pos_xxx`` (for device modules) (GRAP, Akretion) in the OCA are mostly using an alternative to the posbox and so don't use ``hw_xxx`` modules.

ref : https://github.com/akretion/pywebdriver

PR #375, opened 2019-07-19
[11.0][ADD] pos_disable_lot_input: New module

chienandalu:
This module allows to prevent the manual input of lots in the PoS fronted.

This module is better used in combination with
`stock_pack_operation_auto_fill`, wich allows to autoassign lots to lines and
fill quantities based on stock policies. Some workflow tips provided below.

cc @Tecnaitva


pedrobaeza:
What happens if you don't have the other module installed? Picking operations are correctly completed?

chienandalu:
> What happens if you don't have the other module installed? Picking operations are correctly completed?

No, the lots should entered on every picking that would need it.

pedrobaeza:
Then I see an inconsistency...

chienandalu:
I wanted to keep the feature as generic as possible, as maybe other users want to implement a different type of automatation

pedrobaeza:
Talked directly and module will be redirected to `pos_lot_auto_assign`

chienandalu:
Closing this as the full workflow can be achieved setting off "Use Existing Lots/Serial Numbers" in PoS related picking type and setting the proper removal strategy for each location/product category.

Sorry about the noise

MeetKD:
@chienandalu @rafaelbn @pedrobaeza It would be great if you can share link of pos_lot_auto_assign module.

I have set the  Use Existing Lots/Serial Numbers for pos related picking type but it will still ask to enter Lot/Serial Number in pos front end.

chienandalu:
@MeetKD You have to set it off (https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/static/src/js/models.js#L1402)

MeetKD:
@chienandalu thank you for the reference :+1:

PR #378, opened 2019-07-23
12.0 IMP pos_order_mgmt

ivantodorovich:
@PierrickBrun @legalsylvain Hi guys! Thanks for the effort on the other PR, btw.

here are two small improvements on action_print:

- Print invoices instead of tickets, if the order was invoiced
- Destroy the order after printing, so that it's removed from the browser's localStorage

### DIFF
--- a/pos_order_mgmt/static/src/js/widgets.js
+++ b/pos_order_mgmt/static/src/js/widgets.js
@@ -171,13 +171,6 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
         },
 
         action_print: function (order_data, order) {
-            // If it's invoiced, we just print the invoice
-            if (order_data.to_invoice) {
-                return this.pos.chrome.do_action('point_of_sale.pos_invoice_report', {
-                    additional_context: { active_ids: [order_data.id] }
-                })
-            }
-            // Otherwise, we load the order to generate the ticket
             // We store temporarily the current order so we can safely compute
             // taxes based on fiscal position
             this.pos.current_order = this.pos.get_order();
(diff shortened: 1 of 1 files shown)


legalsylvain (changes requested):
Hi @ivantodorovich. Thanks of a lot for your contribution.

> *  Destroy the order after printing, so that it's removed from the browser's localStorage

:100: ! You fix something I didn't understood, when I ported this module. Thanks ;-)

> *  Print invoices instead of tickets, if the order was invoiced

I'm in favour in a more modular approach. Even it is an invoiced order, maybe the cashier want to print the bill with his thermical printer. (mainly because perhaps, the cashier only has a thermical printer) So, I propose to make the following changes : 

- Add a new button print_invoice (displayed only for invoiced order).
- add a new function action_print_invoice with the code you wrote.

        action_print_invoice: function (order_data, order) {
            return this.pos.chrome.do_action('point_of_sale.pos_invoice_report', {
                additional_context: { active_ids: [order_data.id] }
            }
        }

What do you think ?

kind regards.

ivantodorovich:
@legalsylvain Actually that was my first approach, but adding another button made the order list view ugly and it just seemed unnecessary.

How about now? It's similar to the odoo standard: if you are Invoicing the order you get both the receipt and the invoice.

legalsylvain (approved):
Hi ! 
Now is perfect. printing both is the best behaviour IMO, so cashier can give the bill, or print the downloaded pdf.
kind regards.
LGTM, code review, no test.

/ocabot merge patch

PR #380, opened 2019-07-26
IMP Improve configuration page. Allow to disable order mgmt easily. And IMP css

ivantodorovich:
captura


pos_order_mgmt/models/pos_config.py
```diff
@@ -10,26 +11,31 @@
 class PosConfig(models.Model):
     _inherit = 'pos.config'
 
+    iface_order_mgmt = fields.Boolean(
+        string='Order Management',
```
legalsylvain, line 15: Should be default to True.
- updating the module will disable the feature otherwise.
- when installing the module, it is expected the feature is enabled.

pos_order_mgmt/models/pos_config.py
```diff
@@ -10,26 +11,31 @@
...
+        help='Allows to manage orders in the frontend',
+    )
+
     iface_reprint_done_order = fields.Boolean(
```
legalsylvain, line 20: Maybe rename all the field too. (With oldname) to be consistent. 
What do you think ?

legalsylvain (changes requested):
Thanks for the improvment of the UI.
Small remarks. Lgtm otherwise.
Code review no test.

pos_order_mgmt/models/pos_config.py
ivantodorovich: Yeah, you're right

Done

pos_order_mgmt/models/pos_config.py
ivantodorovich: Well.. I prefer not to do that now.
I'm already working on another module that depends heavily on this one, so I don't want to break compatibility.

Personally I prefer to restrict field renaming/removing to migrations. Unless it's of importance

ivantodorovich:
Hi @legalsylvain, you think it's ok now?

legalsylvain:
Hi, for the rename of the field, you should continue using Majuscule in each word, as previously. That's an OCA convention. (https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#37models)

Reprint orders --> Reprint Orders

Otherwise, could be merge. regards.

pedrobaeza:
@legalsylvain not really: there's no convention on uppercase for strings. Only for technical names. In fact, I don't like the abuse of upper everywhere (and in Spanish is discouraged).

legalsylvain:
Hi @pedrobaeza. Well, in fact it was an advice, I received when I began to contribute in OCA repositories, and I don't know where it is written, but it is followed. 

I really don't care about Uppercase or not, but this should be homogeneous, in the same application, and it seems to be the rule. (and in the OCA guidelines sample, it is allways uppercase).

See odoo code source : 
- https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_invoice.py#L1678
See OCA code source : 
- https://github.com/OCA/server-tools/blob/12.0/auditlog/models/rule.py#L66

And, yes, in French too, it is discouraged, but we are talking about english words.
ref odoo code source : 
https://github.com/odoo/odoo/blob/12.0/addons/account/i18n/fr.po#L145 

kind regards.

ivantodorovich:
I've renamed the field to use majuscules for the word 'Order'. I agree that in english seems to be the standard.

Also took the liberty to add 2 more commits to this pr. Very small improvements that help the client-list table to render correctly when the customer name is too long

legalsylvain (approved):
Thanks for all this improvments. Happy to see druidoo contributing on OCA modules. :-)

ivantodorovich:
Thanks!!
We are happy to contribute, too :smile:

PR #381, opened 2019-07-31
[ADD][12.0] New module pos_place (to handle roaming sale)

legalsylvain:
Hi all, 

Proposing a new PoS module ``pos_place`` for users that are making sales out of a shop. (typically, that are one day, in a place A, and the next day, in another place).

this module add a new model ``pos.place`` users can manage.

pos_place_tree

Then, in the point of sale, user has the possiblity to select the place where he is : 

pos_front_end_ui

At the end, user can make statistics about best places : 

report_pos_order_graph

CC : @chienandalu, @quentinDupont 

thanks for your review ! 
kind regards.

#GRAPOCA


pedrobaeza:
This is basically FSM (field service management - https://en.wikipedia.org/wiki/Field_service_management). I think you should depend on https://github.com/OCA/field-service/tree/12.0/fieldservice for having proper base models. I know it seems too complex, but if the piece gets complicated later also for PoS, you are already covered.

legalsylvain:
Hi @pedrobaeza,

Thanks a lot for pointing ``fieldservice`` !
I did not know this project, so I'm currently testing.

In a first sight, I'm not in favour to make dependencies on ``fieldservice`` module, because it's a rather enormous project.

This module ``pos_place`` is very simple, and just add a field on ``pos.order`` model. 
``fieldservice`` is installing a lot of extra modules ``base_geolocalize``, ``partner_fax``, ``web_timeline`` ... and 4 level of locations : territory, branch, district, region, ... with associated workers (fsm.person)... routes, orders (fsm.order) !

**without fieldservice**

image

**with fieldservice**

image

It seems that fieldservice is a rather young repo (since 11.0, CI is all red, etc...) 
I think your proposal could be applied later if fieldservice is a little refactorised splitting concepts by theme to have lighter dependency, and to avoid to install a lot of useless models (useless in the case of this features of course). (having fax on partner model is not mandatory for ``pos_place`` module). Not sure if such refactoring is in the roadmap of fieldservice modules.

Alternatively, it's possible to develop the module ``fieldservice_pos`` but it is out of the scope of this PR. If some peoples try to develop such module, I will review it with pleasure.

kind regards.

@max3903, any extra point of view about this ?

pedrobaeza:
OK, let's keep it simple for now.

quentinDupont (changes requested):
I tested the code and it works like a charm in a clean install !
I suggest an improvement in this PR : https://github.com/legalsylvain/pos/pull/2

quentinDupont (approved):

pos_place/__manifest__.py
```diff
@@ -0,0 +1,37 @@
...
+    'demo': [
+        'demo/res_groups.xml',
+        'demo/pos_place.xml',
+    ],
+    'images': [
+        'static/description/pos_front_end_ui.png',
+        'static/description/pos_config_form.png',
+        'static/description/pos_order_search.png',
+    ],
+}
```
bealdav, line 37: Not mandatory but you could consider to replace ' by " because black in futures versions will steal your commits

pos_place/models/pos_order.py
```diff
@@ -0,0 +1,18 @@
...
+class PosOrder(models.Model):
+    _inherit = 'pos.order'
+
+    # Columns section
+    place_id = fields.Many2one(
+        string='Place', comodel_name='pos.place')
+
+    @api.model
+    def _order_fields(self, ui_order):
+        res = super(PosOrder, self)._order_fields(ui_order)
```
bealdav:
```suggestion
        res = super()._order_fields(ui_order)
```

pos_place/models/pos_place.py
```diff
@@ -0,0 +1,25 @@
...
+    # Columns section
+    code = fields.Char(required=True, size=6)
+
+    name = fields.Char(required=True)
+
+    active = fields.Boolean(default=True)
+
+    company_id = fields.Many2one(
+        string='Company', comodel_name='res.company',
+        default=_default_company_id)
```
bealdav, line 25: Not mandatory but lambda fct is maybe a better solution because your method `_default_company_id` here is not overridable

pos_place/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,9 @@
+To configure this module, you need to:
+
+* Add your users to the new group 'Point of Sale - Places' (User or Manager)
+
+* Go to Point of Sale / Configuration / Places
+
+* Create your places
+
+.. figure:: ../static/description/pos_place_tree.png
```
bealdav, line 9: I don't know it was possible to specify a path like that, really nice

bealdav (approved):
Code review only
Interesting feature, thanks a lot
Some non mandatory typo could be fixed or not

pos_place/readme/CONFIGURE.rst
legalsylvain: indeed, that the only way to make the images displayed in all cases : 
- on github, in the CONFIGURE.rst view
- in the main README.rst, the path will be correctly generated by oca-gen-readme
- so in the apps store (and OCA store) it is OK
- locally (apps / local modules) it is also OK.

pos_place/models/pos_place.py
legalsylvain: you're right. I'll do it.

pos_place/__manifest__.py
legalsylvain: Well, It will be done automatically when migrating in version > 12.0.
In 12.0 it is not an OCA convention.

pos_place/models/pos_place.py
legalsylvain: Caramba ! Too late ! @sebastienbeau merged the PR !

PR #384, opened 2019-08-13
[12.0][MIG] pos_cash_move_reason

legalsylvain:
* [MIG] Port module from 8.0 to version 12.0.
* [REF] Don't use ``product.product`` model for Reasons, because Odoo remove
  the fields ``expense_pdt`` ``and income_pdt`` from the model.
  Use instead a new model ``pos.move.reason`` for this purpose.
* [REF] Doesn't inherit from ``cash.box.in`` and ``cash.box.out`` model,
  as there are bad designed and doesn't allow clean inheritance.
  Instead, use new transient model ``wizard.pos.move.reason``.


legalsylvain:
CC : @adrienpeiffer, @quentinDupont

quentinDupont (commented):
**demo datas**
account_account demo data is loaded but account are not created.. (no time to investigate more but i have to check "Show Full Accounting Features" for demo user, maybe that's related ?)
→so  pos_move_reason demo data don't have account, neither journal

**Otherwise**
Works perfectly in clean odoo12

I didn't review code

gregory-moka:
Hello @legalsylvain 
This request can't be merge ?

JordiBForgeFlow (approved):
:+1: Tested. Works as expected.

quentinDupont (approved):

PR #387, opened 2019-08-20
[ADD] pos_check_session_state

legalsylvain:
Add a new simple module, that check that the PoS session is still in a opening state.
It prevents error, if a user close the session in a back office, and another waiter is using one.

**configuration**

pos_config_form

**Result**

error_message


chienandalu (changes requested):
Nice module :)

Some remarks:

- As the pos window is blocked, it'd be nice to have a button in the dialog to close it:
image

- Config parameter could be in a more integrated style:
image

- Check lint errors

legalsylvain:
Hi all.
Thanks for your review.
@MeetKD : I fixed your points.
@chienandalu : I refactored the code. (more simple, no more new Popup Object), and add a button OK. When clicked the PoS is closed that is the expected behaviour, because it is not possible to use it anymore.

could you update your review ? 

thanks !

pos_check_session_state/static/src/js/chrome.js
```diff
@@ -0,0 +1,71 @@
...
+                    clearInterval(self.intervalIDCheckSessionState);
+                }
+            })
+            .fail(function(error, event){
+                // Prevent error if server is unreachable
+                event.preventDefault();
+            });
+        }
+    });
+
```
MeetKD, line 70: Can we remove extra line ?

pos_check_session_state/views/view_pos_config.xml
```diff
@@ -0,0 +1,31 @@
...
+                        <label for="check_session_state_frequency"/>
+                        <div class="text-muted">
+                            The value is set in seconds.
+                        </div>
+                        <div class="content-group mt16">
+                            <field name="check_session_state_frequency" required="1"/>
+                        </div>
+                    </div>
+                </div>
+
```
MeetKD, line 26: Can we remove extra line here too ?

MeetKD (changes requested):

legalsylvain:
@ MeetKD said:

> Can we remove extra line ?

I personally don't see why. Matter of taste.... But sure, feel free to make a PR, if you think it's important !

In the meantime, could you update your review ? 

Thanks !

MeetKD (approved):
@legalsylvain Yes its matter of taste :+1:  Updated Review for now.

legalsylvain:
Hi @rafaelbn, could you unblock this PR ? I think i made the required changes.

thanks !

chienandalu (approved):

chienandalu:
Sorry for the late review

/ocabot merge nobump

PR #391, opened 2019-08-30
[8.0][FIX] Print status in hw_telium_payment_terminal

### LINKED ISSUE #372: [8.0][hw_telium_payment_terminal] status is not implemented
[This line](https://github.com/OCA/pos/blob/8.0/hw_telium_payment_terminal/controllers/main.py#L272) shows that there is no way to know the status of the payment terminal.
For example, http://127.0.0.1:8069/hw_proxy/status always shows the terminal as "connecting", even when it is fully functional.

- A not-so-bad solution would be to display a message like "unknown status, your payment terminal might be OK" to not lead people to understand the the terminal is actually connecting.
- A good solution would be to really implement a way to check that the terminal is connected.

azmeuk:
This should partially fix #372

Statuses will not be updated on-demand, but they will be updated as the terminal is used. This can be useful for debugging (actually, this could have saved me some time :) ).

### DIFF
--- a/hw_telium_payment_terminal/controllers/main.py
+++ b/hw_telium_payment_terminal/controllers/main.py
@@ -239,6 +239,15 @@ class TeliumPaymentTerminalDriver(Thread):
                 self.device_name, self.device_rate,
                 timeout=3)
             logger.debug('serial.is_open = %s' % self.serial.isOpen())
+
+            if self.serial.isOpen():
+                self.set_status("connected",
+                                "Connected to {}".format(self.device_name))
+            else:
+                self.set_status("disconnected",
+                                "Could not connect to {}"
+                                .format(self.device_name))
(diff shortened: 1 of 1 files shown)


rafaelbn:
@azmeuk hi, check please travis :red_circle:

azmeuk:
I fixed the travis linter tests that concerned the lines I have edited. It still failed, do you want to fix the other failures on hw_telium_payment_terminal?

pedrobaeza:
You can rebase again that I have fixed RST formatting issues. As this branch is very old, the RST check was not yet implemented on that days.

azmeuk:
The rebase is done.

pedrobaeza (approved):

rafaelbn (approved):

PR #393, opened 2019-09-29
[11.0][ADD] pos_cashier_login

kirca:
Require cashier to be scanned before each new PoS order.

pos_cashier_login


pos_cashier_login/static/src/js/main.js
```diff
@@ -0,0 +1,40 @@
...
+odoo.define('pos_cashier_login.main', function (require) {
+    'use strict';
+
+    var gui = require('point_of_sale.gui');
+    var screens = require('point_of_sale.screens');
+    var chrome = require('point_of_sale.chrome');
+    var pos_models = require('point_of_sale.models');
+    var SuperPosModel = pos_models.PosModel.prototype;
+
+    pos_models.PosModel = pos_models.PosModel.extend({
```
hparfr, line 12: you may use an include instead:
```js
pos_models.PosModel.include({
...
})
```

hparfr (approved):
The code looks good to me.

pos_cashier_login/static/src/js/main.js
kirca: Tried but it is a backbone model and it doesn't have such a method.

kirca:
@hparfr thanks for review!

pos_cashier_login/static/src/js/main.js
hparfr: ok :)

PierrickBrun (approved):

chienandalu (approved):
Thanks @kirca :+1:

PR #395, opened 2019-09-30
[MIG] 12.0 payment terminal

PierrickBrun:


legalsylvain (changes requested):
sorry @PierrickBrun but your PR is quite hard to review because you blackified all the module.

1) for the time being, black is not an OCA convention.
2) if it become an OCA convention, please blackify in a separate commit to make easy the review.

thanks.

PierrickBrun:
@legalsylvain I've divided it into two commits

pos_payment_terminal/__manifest__.py
```diff
@@ -1,14 +1,14 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     'name': 'POS Payment Terminal',
-    'version': '10.0.0.1.1',
+    'version': '12.0.0.1.0',
     'category': 'Point Of Sale',
     'summary': 'Manage Payment Terminal device from POS front end',
-    'author': "Aurélien DUMAINE,Akretion,Odoo Community Association (OCA)",
+    'author': "Akretion, Odoo Community Association (OCA)",
```
legalsylvain: why did you removed the author ?

pos_payment_terminal/static/src/xml/pos_payment_terminal.xml
```diff
@@ -3,7 +3,7 @@
     <t t-extend="PaymentScreen-Paymentlines" >
         <t t-jquery=".col-name" t-operation="append">
             <t t-if="line.cashregister.journal.payment_mode and widget.pos.config.iface_payment_terminal">
-                   <button class="payment-terminal-transaction-start"  t-att-data-cid='line.cid'>Start transaction</button>
+                   <button class="payment-terminal-transaction-start"  t-att-data-cid='line.cid'>Send to terminal</button>
```
rousseldenis, line 6: Why this change? IMHO user 'functionnaly' understands more 'Start transaction' than send one. But it's my feeling

rousseldenis (approved):
Code review. LGTM

legalsylvain (changes requested):
Code review & test

minor remarks. on journal, it seems that it is missing a group. The field is not good displayed : 

image

more important problem, if the iot box / pywebdriver is not responding, I can not confirm the order. Did you faced the same problem ?

image

regards.

PierrickBrun:
@legalsylvain I've been able to reproduce, this needs some refactoring I think

pos_payment_terminal/__manifest__.py
PierrickBrun: I've left only the organisations in this field, isn't it how it is supposed to work ? Aurélien and Alexis are on the top of the file

pos_payment_terminal/static/src/xml/pos_payment_terminal.xml
PierrickBrun: I thought it was easier to understand but this is only my point of view. Not changing it should be better for the users in case of migrations

legalsylvain:
@PierrickBrun : any news on that PR ? 

regards.

PierrickBrun:
@legalsylvain None at all and I don't think I will continue working on this since I realized the hardware I wanted to use it with is not compatible with https://github.com/akretion/pypostelium

legalsylvain:
supersed by https://github.com/OCA/pos/pull/440

PR #400, opened 2019-10-16
12.0 mig pos report session summary

wpichler:


chienandalu (commented):
Wasn't the original one a backport of later Odoo core features?

wpichler:
The original one was in odoo v9 (https://github.com/odoo/odoo/blob/9.0/addons/point_of_sale/views/report_sessionsummary.xml) - was removed in odoo v10. So it went into oca/pos

PierrickBrun (approved):

legalsylvain (approved):
Tested on runbot. Works perfectly. Thanks.

PR #402, opened 2019-10-17
[FIX] pos_order_mgmt: Xml receipt renders different values

chienandalu:
- In some cases, the rendered ticket is not fine when we directly send
to XmlReceipt print.
- To get rid of this kind of issues, we allways go to the receipt screen
wich will allways give us the same results for both options (html and
escpos).

cc @Tecnativa TT19501

### DIFF
--- a/pos_order_mgmt/static/src/js/widgets.js
+++ b/pos_order_mgmt/static/src/js/widgets.js
@@ -163,23 +163,15 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
             // taxes based on fiscal position
             this.pos.current_order = this.pos.get_order()
             this.pos.set_order(order);
-            if (this.pos.config.iface_print_via_proxy) {
-                this.pos.proxy.print_receipt(QWeb.render(
-                    'XmlReceipt', {
-                        receipt: receipt,
-                        widget: this,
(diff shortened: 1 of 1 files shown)


pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -163,23 +163,15 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
...
-                this.gui.show_screen('receipt');
-                this.pos.reloaded_order = false;
-            }
+            this.pos.reloaded_order = order;
+            var skip_screen_state = this.pos.config.iface_print_skip_screen;
+            // Disable temporarily skip screen if set
+            this.pos.config.iface_print_skip_screen = false;
+            this.gui.show_screen('receipt');
+            this.pos.reloaded_order = false;
+            // Se skip screen to whatever previous state
```
pedrobaeza: s/Se/Set

pedrobaeza (approved):

chienandalu:
This has been tested in production by some weeks now.

pedrobaeza:
@legalsylvain maybe you can give second review

sergio-teruel (approved):

PR #403, opened 2019-10-18
[11.0] pos_cashier_login: use the default OCA icon

kirca:


legalsylvain (approved):
Hi @kirca. Generaly, self merging PR without review is not a good practice in OCA.

About that PR, LGTM. Thanks.

kirca:
Hi @legalsylvain , I thought I can save some time and not have someone review since is just metafiles which I forgot to edit properly when initially contributing it. But you are right, this shouldn't be done in general. Thanks for taking the time to review and merge!

PR #405, opened 2019-10-28
[FIX] pos_picking_delayed: ensure test correctness

chienandalu:
In a CI test environment, we could end with pricelists with different
currencies or default product taxes. This commit ensures the test correctness.

cc @Tecnativa TT18625

### DIFF
--- a/pos_picking_delayed/tests/test_module.py
+++ b/pos_picking_delayed/tests/test_module.py
@@ -15,6 +15,7 @@ class TestModule(TransactionCase):
             'name': 'Test POS product',
             'type': 'product',
             'available_in_pos': True,
+            'taxes_id': False,
         })
         self.pricelist = self.env['product.pricelist'].create({
             'name': 'Test pricelist',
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

chienandalu:
@legalsylvain This is should go fast. Can you check?

pos_picking_delayed/tests/test_module.py
```diff
@@ -24,21 +25,21 @@ def setUp(self):
                 'base': 'list_price',
             })]
         })
-        # Create a new pos config and open it
-        self.pos_config = self.env.ref('point_of_sale.pos_config_main').copy()
+        self.pos_config = self.env.ref('point_of_sale.pos_config_main')
```
legalsylvain: Why did you removed .copy() ?

legalsylvain (commented):
One question. Lgtm otherwise.

pos_picking_delayed/tests/test_module.py
chienandalu: I had some issues with some defaults not beign applied or provoking problems, so I found that in the end having the default pos_config was more stable after all.

legalsylvain: The copy allow to use DB, and run test, during development. (Otherwise, if the session is opened, the tests fail), so I'd like to keep it, _if possible_.
I dont understand how copy config could break default value, could you be more precise ?
Thanks.

legalsylvain:
thanks for this patch.

/ocabot merge

pos_picking_delayed/tests/test_module.py
chienandalu: Sylvain you were too fast! :sweat_smile: :   I just wanted to retest my CIs and in this manner they still fail. I'd say I that in order to keep the copy() when the instance currency is different from the demo data one I have to assign the pricelist to both of them... I'll test in another PR :smile:

PR #406, opened 2019-11-06
[9.0][ADD] pos_payment_restriction

rousseldenis:


pos_payment_restriction/static/src/js/models.js
```diff
@@ -0,0 +1,15 @@
...
+*/
+
+odoo.define('pos_payment_restriction.models', function (require) {
+    "use strict";
+
+    var models = require('point_of_sale.models');
+
+    var _paymentlineproto = models.Paymentline.prototype;
+    models.Paymentline = models.Paymentline.extend({
+    });
```
chienandalu: I guess this is WIP, isn't it? :sweat_smile:

rousseldenis: Done

legalsylvain:
Hi @rousseldenis. thanks for sharing this module. 
I didn't made a deep review, but in the first sight, I'd better see the settings, in the journal model, to allow to disable / enable both feature payment method by payment method.
What do you think ?

rousseldenis:
> Hi @rousseldenis. thanks for sharing this module.
> I didn't made a deep review, but in the first sight, I'd better see the settings, in the journal model, to allow to disable / enable both feature payment method by payment method.
> What do you think ?

It's quite linked to POS behaviour, so IMHO it should stay there.

legalsylvain:
> It's quite linked to POS behaviour, so IMHO it should stay there.

answered here : https://github.com/OCA/pos/pull/598#issuecomment-773224298

PR #407, opened 2019-11-08
[FIX] pos_stock_picking_invoice_link: ensure test correctness

chienandalu:
When the currency of test environment is different from the one of de
demo data, the test will fail. We must ensure that we create a pricelist
with the current company defaults.

cc @Tecnativa TT18619


chienandalu:
This can be fast tracked once tests have passed

pedrobaeza (approved):
Yeah, this kind of test resilience is not always got with tests. Fast-tracking through ocabot merge.

PR #410, opened 2019-11-17
[MIG] migrate module to v12

houssine78:
This module fix the issue on picking when there is two lines on 
the pos order for the same product, with one of lines with a 0 quantity.
The lines with quantity set to 0 are removed before the pos order is 
processed to avoid such issue.

### DIFF
--- a/pos_margin/i18n/es.po
+++ b/pos_margin/i18n/es.po
@@ -20,12 +20,12 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 #. module: pos_margin
-#: model:ir.model.fields,field_description:pos_margin.field_pos_order_line_purchase_price
+#: model:ir.model.fields,field_description:pos_margin.field_pos_order_line__purchase_price
 msgid "Cost Price"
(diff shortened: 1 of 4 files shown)


houssine78:
@legalsylvain this module could be interesting for the repo. Any comments welcome. The fix brought is quiet straight forward.

pedrobaeza:
IMO this should be fixed at odoo/OCB level instead of here. Any way, are you sure this is not fixed already?

houssine78:
hi @pedrobaeza yes I just tested if it still occurs today and it does.
The best option would be to have it fixed on Odoo or OCB but as usual it would be considered as a feature request. To have it on OCB I think it should be at least considered by odoo as a bug and not a feature at least until v12 is still supported.

I can make a PR on Odoo and OCB if you think it's the best option.

pedrobaeza:
Why this has been merged?

legalsylvain:
I dont understand anything on this PR. There are no diff in the changes section,and it has been merged by OCA bot, without any order.

pedrobaeza:
OK, it seems the branch has been aligned with main branch, being detected as merged.

PR #413, opened 2019-11-19
[FIX] pos_picking_delayed: ensure test correctness

chienandalu:
https://github.com/OCA/pos/pull/405 was prematurely merged :smile: . This one can be fast-tracked once CI's pass

cc @Tecnativa

### DIFF
--- a/pos_picking_delayed/tests/test_module.py
+++ b/pos_picking_delayed/tests/test_module.py
@@ -25,11 +25,13 @@ class TestModule(TransactionCase):
                 'base': 'list_price',
             })]
         })
-        self.pos_config = self.env.ref('point_of_sale.pos_config_main').copy()
-        self.pos_config.write({
+        pos_config_main = self.env.ref('point_of_sale.pos_config_main')
+        # Ensure tests correctness despite default currency
+        pos_config_main.write({
             'available_pricelist_ids': [(6, 0, self.pricelist.ids)],
             'pricelist_id': self.pricelist.id,
         })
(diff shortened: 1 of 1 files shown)


legalsylvain:
Hi don't understand the reason of this patch. Could you point the error ?
Travis is green, and all seems OK.

chienandalu:
> Could you point the error ?

Again? :sweat_smile: The reason is to ensure tests correctness despite default currency. Otherwise, there is an exception due to the default pricelists default currency wich could be different. So:

- A pricelist must be created on the test
- Such pricelist must be assigned to de pos.config used.
- If we just copy the pos.config, the exception raises.
- So in this patch, we ensure the copied pos.config has no such issues.

pedrobaeza (approved):
I don't have clear why the previous code didn't work, but this is the same harmless IMO, so go ahead.

chienandalu:
Because the of the copy()

chienandalu:
@legalsylvain This is the exact error. And it's due to the reasons I pointed out.
```
ValidationError: ('All available pricelists must be in the same currency as the company or as the Sales Journal set on this point of sale if you use the Accounting application.', None)
```
In this last patch I respect your implementation of the test but just preventing this fail.

Let's merge it and go on with other stuff :wink:

legalsylvain:
OK. On that point, not sure to have understood finally ;-), as Travis branch is green. (https://travis-ci.org/OCA/pos/branches) But no problem at all to merge, and let's go talking about more interesting point.

Thanks for the patch !

/ocabot merge

PR #417, opened 2019-11-21
[FIX] pos_order_mgmt: Xml receipt renders different values

chienandalu:
Forward Port of #402

- In some cases, the rendered ticket is not fine when we directly send
to XmlReceipt print.
- To get rid of this kind of issues, we allways go to the receipt screen
wich will allways give us the same results for both options (html and
escpos).

cc @Tecnativa


pedrobaeza (approved):
If CIs go well

legalsylvain (approved):
Make Sense. 

code review. LGTM. thanks.

PR #418, opened 2019-11-27
[MIG] pos_order_return: Migration to 12.0

chienandalu:
Standard migration

cc @Tecnativa TT20736


pos_order_return/models/product_template.py
```diff
@@ -10,4 +10,5 @@ class ProductTemplate(models.Model):
 
     pos_allow_negative_qty = fields.Boolean(
         string='Allow Negative Quantity on PoS',
+        default=True,
```
pedrobaeza, line 13: Why this default change?

pos_order_return/wizard/pos_partial_return_wizard.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2016-2018 Sylvain LE GAL (https://twitter.com/legalsylvain)
+# Copyright 2018 Lambda IS DOOEL <https://www.lambda-is.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class PosPartialReturnWizard(models.TransientModel):
+    _name = 'pos.partial.return.wizard'
```
pedrobaeza, line 9: Missing `_description` (runbot warning comes from here)

pedrobaeza (approved):

pos_order_return/models/product_template.py
chienandalu: First, to avoid default annoying behaviors or tests fails. It's a nices feature to have, but wherever we got this module in an instance we finally had to set this setting to True by default.

pedrobaeza: OK

chienandalu:
Waiting for reviews here :)

pos_order_return/models/product_template.py
rousseldenis: I would say that it will be nice to have this feature (allow negative) in another module (separate functionality).

rousseldenis (approved):
Code review

pos_order_return/models/product_template.py
legalsylvain: @chienandalu : Could you just change the readme of configuration . 

Replace : 

> In some cases, you may want to let the possibility to allow negative quantity

by 

> In some cases, you may want to **block** the possibility to allow negative quantity ... etc...

legalsylvain (changes requested):
A little remarks regarding description.

The rest looks good to me.

thanks ! :+1: (code review / no test)

legalsylvain:
Damn'it ! concurrent review !

@chienandalu could you take into account my remarks here ? 

https://github.com/OCA/pos/pull/418#discussion_r356556011

thanks !

pedrobaeza:
Yeah, sorry, but having default=True isn't the redaction of the README correct?

legalsylvain:
```
In some cases, you may want to let the possibility to allow negative quantity
in a PoS Order, without mentioning initial order. This can happen for special
products like returnable products, etc.

In that case, a checkbox is possible on Product Form View to allow such case
```

Reading that text, I suppose that by default, it is not allowed and, _in some cases_, it is possible to allow negative quantity. That is wrong, because, by default, negative quantity are now allowed, due to that change introduced in that PR. Don't you think ?

pedrobaeza:
Yeah, right. It makes sense. Just note that the change is done for keeping the behavior without this module (standard doesn't block by this reason). And it's included here for forcing to use returns instead of using negative quantities.

legalsylvain:
>  Just note that the change is done for keeping the behavior without this module

Yes, no problem regarding the change. I think it's good !

my remark was only about the description.

pedrobaeza:
@chienandalu please change the README according comments. As you are PSC, I think this can be done as direct commit without PR.

chienandalu:
Sure, thanks @legalsylvain

chienandalu:
Changed: https://github.com/OCA/pos/commit/95481f2a10c5ac30e28e4c839bb1af73dab8b778

legalsylvain:
Thanks !

PR #421, opened 2019-12-09
[12.0][ADD] pos_ticket_without_price

dalonsod:
This module allows to print a gift ticket, which is the same ticket than normal but without prices and taxes.

In previous versions of Odoo and POS engine, [there was originally a module](https://github.com/OCA/pos/tree/8.0/pos_gift_ticket) with the same name and functionality. I've tried to port this one, but developing it from scratch.

CC @ismaelcj @chienandalu


legalsylvain (changes requested):
Hi. Thanks for sharing this module.
2 points 
I think that you should not use replace operator on the ticket. You risk to break inheritance mechanism of other pos modules are installed. You could use css ans invisible instead.

I see the name of the module a little ambiguous. Gift sugger that it is not paid, that is not true. I think that somethink like pos_ticket_without_price should be better.
What do you think ?
Regards.

dalonsod:
Thanks, @legalsylvain !!

1. In order to make invisible instead of using replace mechanism, could you help me giving an example of that? I'm newbie in QWeb and POS and I'm not sure what is the best approach to achieve it.

2. No problem with the name change!!

legalsylvain:
I guess using ``t-operation="attributes"`` and adding a class or an attribute that make the content invisible.
I think that it should be used for the 3 ``replace`` and the ``inner`` operations.

Qweb Ex : 
https://github.com/odoo/odoo/blob/12.0/addons/website_sale_link_tracker/static/src/xml/website_sale_dashboard_utm_chart.xml#L4

thanks !

dalonsod:
Done!!!

legalsylvain:
Thanks a lot. Tested on runbot, It works well. I have a only one UX problem. Step to reproduce : 

- Once an order is validated
- click on "Print Receipt" -> OK: the regular receipt is printed
- Click on "Print Receipt without price" -> OK: the without price receipt is displayed and printed
- click on "Print Receipt" ->KO: the without price receipt is still displayed and printed

dalonsod:
> 
> 
> Thanks a lot. Tested on runbot, It works well. I have a only one UX problem. Step to reproduce :
> 
>     * Once an order is validated
> 
>     * click on "Print Receipt" -> OK: the regular receipt is printed
> 
>     * Click on "Print Receipt without price" -> OK: the without price receipt is displayed and printed
> 
>     * click on "Print Receipt" ->KO: the without price receipt is still displayed and printed

Now it works, but I'm not sure what's the proper way to change the click event handler, could you check it? The code is at https://github.com/OCA/pos/pull/421/files#diff-14cd01dcdd944ffcaf8245dc097273dcR19

lmiguens-solvos:
@dalonsod please check this code to override the click event handler:
https://github.com/OCA/pos/blob/12.0/pos_access_right/static/src/js/pos_access_right.js#L160

I think it shoud be a better solution than yours

legalsylvain:
thanks @lmiguens-solvos !

dalonsod:
Request changes done

lmiguens-solvos (approved):
LGTM 👍

chienandalu (changes requested):
I miss the XmlReceipt support. It should not take more than override the proper template I think.

dalonsod:
> 
> 
> I miss the XmlReceipt support. It should not take more than override the proper template I think.

Well, if proxy mode is set, the current behavior is printing directly without showing the Receipt screen... So I can't call the ticket without price printing instead of default one, only in web mode I can. So, my current approach is not compatible with proxy mode :(

Any idea of how to solve this? Thanks!!

legalsylvain:
@dalonsod no. Feel free to add a ROADMAP.rst file in the readme folder, to mention that it's not compatible for the time being with the proxy mode.
Maybe You'll receive a PR that improve your module.
kind regards.

dalonsod:
> 
> 
> @dalonsod no. Feel free to add a ROADMAP.rst file in the readme folder, to mention that it's not compatible for the time being with the proxy mode.
> Maybe You'll receive a PR that improve your module.
> kind regards.

Thanks for the help. Added roadmap and usage files.

dalonsod:
@chienandalu @legalsylvain is this PR ok or are there pending changes?

Thanks!!

legalsylvain:
Let's go merging this PR. improvment can be done later, regarding proxy / non proxy mode.

Thanks for your contribution !

legalsylvain (approved):

PR #385, opened 2019-08-15
12.0 mig pos margin

legalsylvain:
migrate pos_margin

- from 11.0 to 12.0
- [REF] OCA convention (readme folder, etc...)
- [IMP] handle uom_id, based on the work of @wpichler here https://github.com/OCA/pos/pull/360
- [FIX] reuse algorithm of ``sale_margin``, so handle correctly multicurrency context.
- [ADD] test, improving coverage.

supersed : #360

PR #423, opened 2019-12-12
12.0 add margin percent pos margin

legalsylvain:
Hi all. 
some improvement on ``pos_margin`` module, following same logic as the module [OCA/margin-analysis/account_invoice_margin](https://github.com/OCA/margin-analysis/tree/12.0/account_invoice_margin/)

- [ADD] me as maintainer ; 
- [ADD] percent margin on pos.order and pos.order.line
- [IMP] view, adding margin and margin percent on tree view

CC : 
- original reviewers of #385 @wpichler, @rousseldenis
- @quentinDupont 

pos_order_form
pos_order_tree


pos_margin/models/pos_order.py
```diff
@@ -11,14 +11,29 @@ class PosOrder(models.Model):
...
+    @api.depends('lines.margin', 'lines.price_subtotal')
     def _compute_margin(self):
         for order in self:
-            order.margin = sum(order.mapped('lines.margin'))
+            tmp_margin = sum(order.mapped('lines.margin'))
+            tmp_price_subtotal = sum(order.mapped('lines.price_subtotal'))
+            order.update({
+                'margin': tmp_margin,
+                'margin_percent': tmp_price_subtotal and (
+                    tmp_margin / tmp_price_subtotal * 100),
```
rousseldenis:
```suggestion
                    tmp_margin / tmp_price_subtotal * 100) or 0.0,
```

legalsylvain: Hi @rousseldenis, thanks for your review.

I see your suggestion unnecessary. Don't you think ? 
``if not tmp_price_subtotal`` the result will be False, so will be converted to 0.0

But no problem to add it. Just to understand.

rousseldenis: Yes, indeed. I was used to that notation and be explicit :-)

legalsylvain: I added it.

robyf70 (approved):
LGTM

GSLabIt:
@legalsylvain any chance to add Margin (%) in reports graph and pivot? Seems like Margin Rate is referred to some other thing.

legalsylvain:
@GSLabIt I don't think so. iMHO Margin percent are useless in report because Odoo will make sum or average of percent, giving wrong values.
Anyway, It's not in the scope of this PR, but feel free to try to add this feature, I'll review it.
Regards.

pedrobaeza:
You can do proper margin percentage computation intercepting `read_group` method and doing the operation "on the fly". See an example in https://github.com/OCA/hr/blob/2674c06104bee007756dd3d8f541e52d9fcf0bc5/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py#L237-L239

legalsylvain:
Hi @GSLabIt. Did you have some time to make a PR, regarding @pedrobaeza suggestion ? 
Otherwise, I propose to merge this PR and see after for some improvment.

kind regards.

GSLabIt:
> Hi @GSLabIt. Did you have some time to make a PR, regarding @pedrobaeza suggestion ?
> Otherwise, I propose to merge this PR and see after for some improvment.
> 
> kind regards.

@legalsylvain actually not much time for that, may be better merging it and then having PR to improve as suggested.

PR #424, opened 2019-12-12
[NEW][10.0] POS order remove line

robyf70:
Add a new module to remove pos order line


legalsylvain:
Hi @robyf70. Thanks for this module .Could you check travis syntax error ? 
Thanks ! 

Ref : https://travis-ci.org/OCA/pos/jobs/624270199#L1147

robyf70:
> Hi @robyf70. Thanks for this module .Could you check travis syntax error ?
> Thanks !
> 
> Ref : https://travis-ci.org/OCA/pos/jobs/624270199#L1147

@legalsylvain Yep! Doing that!

pos_order_remove_line/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+**English**
```
legalsylvain: [nitpicking] I think that it is unncessary, as all the description are allways in English.

legalsylvain:
Hi @robyf70. Justed on runbot. Seems to work like a charm.

- UX remark
-> the size of the button is smaller that the other PoS button. I think it's a problem, because can not be used very on tactile screen.
-> Also a border is expected to mention to the user that it is a button. (as all the other buttons in the PoS)
-> the image is generally in recent Odoo version a bin, not a red cross 

**Current image** 
image

**What it could be expected** : (with PoS css and bin, this is a painted image ;-))
image


Otherwise, LGTM thanks !

pos_order_remove_line/readme/DESCRIPTION.rst
robyf70: Yep! Done!

robyf70:
@legalsylvain I've changed the icon to a standard odoo bin and now the icon looks a button. I've already tested the trash bin icon on a tablet and a touch screen the usability looks pretty ok to me

legalsylvain (approved):
image

Nice ! 

Well, i still think that the button is too small, but matter of taste ! Non blocking point.

:+1: (code review + Test).

Thanks.

eLBati (approved):
Thanks

chienandalu (commented):
Wouldn't it be interesting to have the deletion button visible only on selected lines?

robyf70:
> Wouldn't it be interesting to have the deletion button visible only on selected lines?

I think it's much quicker to delete whatever line without having selected it

chienandalu:
> I think it's much quicker to delete whatever line without having selected it

Sure, but it can be prone to errors with clumsy users :sweat_smile:

robyf70:
> > I think it's much quicker to delete whatever line without having selected it
> 
> Sure, but it can be prone to errors with clumsy users

Indeed, but doesn't look like a big problem. at least the operator has to retype/scan the deleted item, no?

chienandalu:
Ok. It was just a thought ;)

chienandalu (approved):

legalsylvain:
> Sure, but it can be prone to errors with clumsy users sweat_smile

We can keep in mind of the road map to have an option that enable confirmation "Are you sure you want to delete this line". (the same message as ``pos.order`` actually)
Something like a boolean on ``pos.config`` model ``has_clumsy_users``. But not in the scope of this PR I guess. Could be added later.

regards.

/ocabot merge

PR #428, opened 2019-12-17
[ADD] 12.0 pos_mail_receipt module

PierrickBrun:


chienandalu (commented):
Is this a backport of v13 functionality?

PierrickBrun:
> Is this a backport of v13 functionality?

I did not know v13 included this. So this is not a backport.

This module allows you to send an e-mail to a customer without creating a partner, this is better to not force it to be more RGPD compliant.
Also in this module you send the ticket after validating the payment and not before, which is more user friendly.

quentinDupont:
Thank you for this module !

- CSS for the email input should be changed a little to be centered

`margin-left: 50px;`

- I didn't received mail from runbot (maybe that's normal ?)

- And it was not clear if the mail was sent after clicking "OK". A green warming message "The email was sent" would be nice !

PierrickBrun:
I "fixed" the textinput

You can see the failed sent mails in Settings -> Technical -> Email -> Emails (you need to be in debug mode)

I'll change the button color once clicked to avoid having to handle notifications (unless it is easy but I don't know how)

pos_mail_receipt/models/pos_order.py
```diff
@@ -0,0 +1,70 @@
...
+from odoo import fields, models, api, _
+
+
+class PosOrder(models.Model):
+    _inherit = "pos.order"
+
+    email_receipt_sent = fields.Boolean()
+
+    @api.model
+    def send_mail_receipt(
```
legalsylvain, line 20: I think that this function should be done asynchroneously, by queue job. 
don't you think ? 

like this module : https://github.com/OCA/pos/tree/12.0/pos_picking_delayed

pos_mail_receipt/static/src/js/screens.js
```diff
@@ -0,0 +1,93 @@
...
+
+            return rpc.query({
+                    model: 'pos.order',
+                    method: 'send_mail_receipt',
+                    args: [order, options["email"], options["body_from_ui"]],
+                }, {
+                    timeout: timeout,
+                })
+                .then(function (result) {
+                    console.log(result);
```
legalsylvain: to remove. ;-)

pos_mail_receipt/models/pos_order.py
```diff
@@ -0,0 +1,70 @@
...
+        self, pos_reference, email, body_from_ui, force=True
+    ):
+        order = self.search([("pos_reference", "=", pos_reference)])
+        if len(order) < 1:
+            return _("Error: no order found")
+        if order.email_receipt_sent:
+            return _("E-mail already sent")
```
legalsylvain: the returned values are not used. I think that a logger should be better. Don't you think ?

legalsylvain (commented):
Some remarks inline.

FYI, here is the code used by Odoo, on master branch, to send email.

https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/models/pos_order.py#L634

General question : 
I don't understand the interest to send a receipt before the payment ? 

Thanks !

pos_mail_receipt/models/pos_order.py
PierrickBrun: I'd prefer avoiding this dependency, emails are queued by default also.

I think it is better to have the e-mail sent quickly though, sometimes you need to have it to justify that you did not stole or to have a reduced parking fee.

PierrickBrun:
@legalsylvain I've addressed the problems you saw and added tests

pos_mail_receipt/models/pos_order.py
legalsylvain: > I think it is better to have the e-mail sent quickly though, sometimes you need to have it to justify that you did not stole or to have a reduced parking fee.

Hi thanks for your answer. my thought was 
- about the compatibility of this module with the certification module. (that displays the hash on the bill). But I have no time to test, and the feature was introduced in V13. (I make some developmment for V12, but not yet merged).
- IFAIK, I don't think that if you use ``send_force``, the email is not queued. I think that it will wait for the sent of the mail, and I don't think that it is in the spirit of the PoS module, to wait for server response.
ref https://github.com/OCA/pos/pull/428/files#diff-cc4623d2c77df155a2db869e209c63a2R63

But well, I don't think if it's real issues, and have no time to propose a better design, so I abstain on that point, and it's not a blocking point for me. Let's go reviewing and merging this PR, and see if it's OK.

Thanks for this work !

PierrickBrun:
@legalsylvain @quentinDupont 
I think this is in good shape now.
I had trouble applying good CSS to the PDF generated by the backend, but in the end I got it to work

mclaeysb:
@PierrickBrun Could you explain why the `force=True` option is used by default when the mail is sent, and when the `create_from_ui()` with `force=False` is used? I'm worried the session could be hindered when the email can't be sent.

PierrickBrun:
> @PierrickBrun Could you explain why the `force=True` option is used by default when the mail is sent, and when the `create_from_ui()` with `force=False` is used? I'm worried the session could be hindered when the email can't be sent.

When the request is immediately sent (ie. The PoS is connected to the server), there is a `force=True` to ensure the e-mail is sent right away. This is useful because you may need it to prove you did not steal or to have a free parking for instance.
In that case, since the record is already created when we decide to send the email or not, the email is created from a separated endpoint.

When the request is sent afterwards (ie. The PoS gets back connection) there is no `force=True` because in that case it does not matter to send the email quickly.
In that case the email is sent with the same request as the record creation, in the `create_from_ui`

pos_mail_receipt/static/src/js/screens.js
```diff
@@ -0,0 +1,99 @@
...
+                            'body':  error.data.debug
+                        });
+                         this.$('.button.email').removeClass("highlight");
+                    }
+                    if(connection_problem){
+                        self.gui.show_popup('error',{
+                            'title': _t('The e-mail could not be sent'),
+                            'body': _t('Check your internet connection and try again.'),
+                        });
+                         this.$('.button.email').removeClass("highlight");
```
mclaeysb: I'm testing this branch and getting an error when the email can't be sent, stating `this.$ is not a function` referring to this line. I think this `this` should be `self`, created on line 46, as the `this` is lost here. Same goes for the `this` on line 82. This replacement fixed it for me.

mclaeysb:
credit to @vvrossem for helping with this find!

pos_mail_receipt/static/src/js/screens.js
```diff
@@ -0,0 +1,99 @@
...
+        },
+        email: function() {
+            var self = this;
+            var email = false;
+            var body_from_ui = this.$('.pos-receipt-container').html()
+            if( this.pos.get_order().get_client() && this.pos.get_order().get_client().email ) {
+                self._send_email_server(this.pos.get_order().name, {"email": this.pos.get_order().get_client().email, "body_from_ui": body_from_ui});
+            } else {
+                this.gui.show_popup('textinput', {
+                    'title':_t('E-mail address to use'),
```
mclaeysb: Getting the error that `_t()` is in not found. It should be declared:
```js
odoo.define(...
    var core = require('web.core');
    var _t = core._t;
```

PierrickBrun: Hi @mclaeysb  , thanks for the review :) Would you mind committing your fixes directly to the branch ? I think you have write access to it

mclaeysb: Sure! Sorry for the inconvenience..

flotho (approved):
eassy patch

LGTM

Thanks

vdewulf:
Hello all,
Hello @mclaeysb and @PierrickBrun 
I tested the last version on one of our test database.

1) Send ticket to a customer who has an email adress.
I clicked on the send receipt button, which became green.
The email was sent correctly.

I configured some text on the header/footer in the pos config. Le sent ticket in attachment contains the modifications.

2) Selling without selecting any customer.
When I click on the button to sent the ticket, a pop up opens to indicate an email adress.
The email was sent correclty.

Something somewhat surpising : it is possible to re-click on the button, and enter a new email adress. But no ticket will be sent again.
The best would be to prevent the user to click again on the button.
But still we can live with that at the moment.

3) I cut the Wifi and made a sale.
I was able to click on the button to sent the receipt. There was no error.
The wifi signal was red with the number 1 (to inform there is 1 non synchronised sale).
I re-activated the Wifi. I clicked on the 1 number. Everything was synchronised and the email was in the queue.

=> for me everyting works fine.

cesarlr:
We need to upload it to PyPi. Is it ok? https://github.com/coopiteasy/pos/pull/36 @mclaeysb

mclaeysb:
Hi @cesarlr! For this PR we are waiting for the OCA (@legalsylvain @chienandalu ?) to consider it :)

pos_mail_receipt/static/src/css/print.css
```diff
@@ -0,0 +1,129 @@
...
+    width: 100%;
+    border: 0;
+    table-layout: fixed;
+}
+.pos .pos-sale-ticket table td {
+    border: 0;
+    word-wrap: break-word;
+}
+
+@page {
```
legalsylvain, line 53: All this part will alter all the pdf rendered via ``web.report_assets_pdf``. Don't you think ? 
If yes it should be avoided.

pos_mail_receipt/models/pos_order.py
```diff
@@ -0,0 +1,86 @@
...
+            order.id, force_send=force, email_values=email_values,
+        )
+        order.email_receipt_sent = True
+
+    @api.model
+    def create_from_ui(self, orders):
+        res = super(PosOrder, self).create_from_ui(orders)
+        for order in orders:
+            if "email" in order["data"]:
+                self.send_mail_receipt(
```
legalsylvain, line 87: Well. Thinking regarding queue, I still think that we should do it by queue. Using queue_job for other project, it will not delay the process. Queue job are executed as soon as possible, so you'll have a delay, only if the server is busy, but the same occures with the current design.

I understand it's not in the scope of your PR, but could you add a private function ``_send_mail_receipt()`` called by ``send_mail_receipt`` with all the code. With that design, it will be easy to add a module ``pos_mail_receipt_queue_job`` inheriting

legalsylvain (changes requested):
2 remarks inline. 
Could you also fixed travis. timeout is reached, just after the test you are introducing.

legalsylvain:
some rebase should be done on this branch. (conflict, a lot of file modified, ...)

ivantodorovich:
The branch is still messy unfortunately. Make sure you push with `--force` after rebasing.
Ideally, for this PR, only 1 commit should be here: `[ADD] pos_mail_receipt`

flotho:
Hi @PierrickBrun , any chance to make a push rebase so that everyone could benefit from the huge amount of energy you spent on this?

Regards

PierrickBrun:
Hi @flotho , done. I quickly tested it and it looks like nothing's been broken in the process

legalsylvain (approved):
Just tested again. Still think some things should be changed, but It's not a good thing to block this PR. (sorry, my bad). 

Propose to merge and let improve this module after (or wait for the odoo core feature, in more recent version).

/ocabot merge patch

carmenbianca:
@legalsylvain This won't merge because travis is failing.

legalsylvain:
@PierrickBrun : could you take a look on travis ? 

/ocabot rebase

carmenbianca:
@legalsylvain I'll rebase.

legalsylvain:
still red !

victor-champonnois:
@legalsylvain The travis CI is red because of a timeout. Do you have an idea on  how to fix this ?

legalsylvain:
@victor-champonnois : not sure that is a time out problem. when executing test of ``pos_mail_receipt`` something seems blocked. 

``No output has been received in the last 10m0s``...

@PierrickBrun what do you think ?

victor-champonnois:
Note to @carmenbianca : I did not explicitly resolve the promise in the code. 
- I am not sure it is needed, since the screen is locked until the mail is sent (or raise an errors)
- Also after some trials I did not find a way to do it 
    - await does not work (I have an error telling me it must be used inside an async function/module/class, although I specifically added async to the function)
    - using .resolve() in the `email` function does not work either (`.resolve() is not a function` error)

victor-champonnois:
> @victor-champonnois : not sure that is a time out problem. when executing test of `pos_mail_receipt` something seems blocked.
> 
> `No output has been received in the last 10m0s`...
> 
> @PierrickBrun what do you think ?

@legalsylvain You're right, apparently the problem is related to a call to `_wkhtmltopdf` in the tests.

victor-champonnois:
@legalsylvain This is tricky. The timeout happens when the test call `send_mail_receipt()`, [itself calling `_run_wkhtmltopdf`](https://github.com/OCA/pos/blob/df17842e7cdce02eae825c1133c905dc2ae884f5/pos_mail_receipt/models/pos_order.py#L52) to convert the receipt to a pdf. 
`_run_wkhtmltopdf` run wkhtmltopdf in a subprocess. 
I tried to run the command via the command line and my conclusion is that it is stuck because of the head: 
```
                <link type="text/css" rel="stylesheet" href="/web/content/572-85cd6e9/web.report_assets_pdf.0.css"/>
                <link type="text/css" rel="stylesheet" href="/web/content/569-a6cdbb7/web.report_assets_common.0.css"/>
                <link type="text/css" rel="stylesheet" href="/web/content/570-a6cdbb7/web.report_assets_common.1.css"/>
```
When I remove it everything is fine. I think these links cannot be accessed in the test environment. 
I don't know how to fix this though. It seems impossible to remove the header this high in the call stack. 
Maybe we need to simply remove this test ?

legalsylvain:
I dont know. @ivantodorovich any idea on that topic ?

ivantodorovich:
I've never encountered this myself but I see something that bugs me here:
https://github.com/OCA/pos/blob/df17842e7cdce02eae825c1133c905dc2ae884f5/.travis.yml#L16

I'm not sure that's the right way of adding `wkhtmltopdf` dependency, as Odoo requires very specific versions of it.

On the oca repo template we have this: https://github.com/OCA/oca-addons-repo-template/blob/8f81ed76dd5ad0e315fb3b9729e3d7c9eeb4c867/src/%7B%25%20if%20ci%20%3D%3D%20'Travis'%20%25%7D.travis.yml%7B%25%20else%20%25%7D.t2d.yml%7B%25%20endif%20%25%7D.jinja#L67

Sadly the `12.0` branch is not using copier at all, but maybe you could try adding the travis dependency in the same way.
Not sure it'll fix the issue, but maybe it's worth trying.



> I tried to run the command via the command line and my conclusion is that it is stuck because of the head ... https://github.com/OCA/pos/pull/428#issuecomment-1164544986

Maybe `wkhtmltopdf` doesn't have access to the odoo server on the test environment? Kind of reminds me of the issues you can get due to misconfiguration and `report.url` system parameter

victor-champonnois:
@ivantodorovich Thanks a lot, it works :)

victor-champonnois:
(not ready to merge because new changes has been made since the approval)

pos_mail_receipt/static/src/js/screens.js
```diff
@@ -48,9 +48,16 @@ odoo.define("pos_mail_receipt.screens", function (require) {
             var self = this;
             options = options || {};
             var timeout = typeof options.timeout === 'number' ? options.timeout : 7500;
-            this.$('.button.email').addClass("highlight");
+            var timeout_db_sync = typeof options.timeout_db_sync === 'number' ? options.timeout : 5000;
```
carmenbianca:
```suggestion
            var timeout_db_sync = typeof options.timeout_db_sync === 'number' ? options.timeout_db_sync : 5000;
```

Alternatively, just use the same `timeout` as the line above it.

pos_mail_receipt/static/src/js/screens.js
```diff
@@ -48,9 +48,16 @@ odoo.define("pos_mail_receipt.screens", function (require) {
...
+            var i = 0
+            var waiting_interval = 1000
             while (self.pos.get("synch").state == "connecting") {
-                await sleep(1000);
+                await sleep(waiting_interval);
+                i = i + waiting_interval
+                if (i >= timeout_db_sync){
+                    self.gui.show_popup('error', _t('Timeout syncronising the database'));
```
carmenbianca: This popup is not very helpful to the user.

```suggestion
                    self.gui.show_popup('error', _t('Aborted the sending of the e-mail receipt; could not synchronise with the database in time.'));
```

carmenbianca: Also, add a `self.lock_screen(false)` line here if the `lock_screen(true)` line is moved

pos_mail_receipt/static/src/js/screens.js
```diff
@@ -48,9 +48,16 @@ odoo.define("pos_mail_receipt.screens", function (require) {
...
+                    return
+                }
             }
             self.lock_screen(true);
```
carmenbianca: This should be moved _before_ the loop where it checks for the DB synchronisation. The user could be waiting several seconds, and immediate feedback is good.

.travis.yml
```diff
@@ -17,7 +17,9 @@ addons:
 env:
   global:
   - VERSION="12.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
-
+  {%- if include_wkhtmltopdf %}
+  - WKHTMLTOPDF_VERSION="0.12.5"
+  {%- endif %}
```
ivantodorovich: Actually this is how it should look like.

```suggestion
  - WKHTMLTOPDF_VERSION="0.12.5"
```

`{% - ... - %}` are jinja expressions only used by the repository template.

Adding them here unfortunately has Travis not even run because of a malformed `.travis.yml` file, so in fact we don't know yet if it's fixed or not, because CI didn't run 😢

carmenbianca (changes requested):
Some comments on the recent commits by @victor-champonnois 

Awesome work :)

If you have time, make sure to run `prettier` on `screens.js`. There are a few non-idiomatic lines here and there, but it's not ultimately important.

pos_mail_receipt/static/src/js/screens.js
victor-champonnois: done, you're right, it's simpler this way.

pos_mail_receipt/static/src/js/screens.js
victor-champonnois: > This popup is not very helpful to the user.

Is "'Syncronisation with the database is taking too long. Retry or contact the system administrator.' better ? It's hard to write a helpful message since I don't know what actually *is* database syncronisation !


> Also, add a `self.lock_screen(false)` line here if the `lock_screen(true)` line is moved

done thanks :)

pos_mail_receipt/static/src/js/screens.js
victor-champonnois: To be sure I understand, do you mean adding *another* message if the state is `connecting`, before waiting for the timeout ? I am affraid that this could trigger too often...

pos_mail_receipt/static/src/js/screens.js
carmenbianca: I like your proposal! Just fix the spelling to 'synchronisation' :)

(Database synchronisation, as I understand it, is the browser syncing its local cache with stuff in the DB. But I'm not 100% on the details either.)

pos_mail_receipt/static/src/js/screens.js
carmenbianca: Ah, GitHub UI is doing what GitHub UI does. I'm looking exclusively at the line `self.lock_screen(true);`. That line should be moved up :)

pos_mail_receipt/static/src/js/screens.js
victor-champonnois: OK, fixed :)

pos_mail_receipt/static/src/js/screens.js
victor-champonnois: Ah I see, it's done :)

victor-champonnois:
Status update : we didn't manage to fix the CI. We are deploying to production nevertheless.

ivantodorovich:
> Status update : we didn't manage to fix the CI. We are deploying to production nevertheless.

Switching CI to github actions here: https://github.com/OCA/pos/pull/813
It might solve the issue

legalsylvain:
> Switching CI to github actions here: https://github.com/OCA/pos/pull/813
> It might solve the issue

Done. Rebase (and solve conflict) required.

ivantodorovich:
FTR, tests are still timing out even after switching to gh actions.
They should probably be re-written

victor-champonnois:
Indeed, tests still timeout, the reason is explained [here](https://github.com/OCA/pos/pull/428#issuecomment-1164544986). But, after some time investigating, with the help of other, we still don't understand what's wrong.
@ivantodorovich 
- I don't see a way to rewrite the tests, since there is only one function to test. Do you have an idea ?
- Could we consider merging this PR without the tests ?

ivantodorovich:
If we can't solve this issue, I would be in favour of merging without the tests.
But the situation is far from ideal, I would hate to see this issue arise in the future because some other tests trigger this method and fail. It would be a pain to debug.

I've just skimed through the code and I wonder, do we really need to call `_run_wkhtmltopdf` directly? Couldn't you just create an actual report for the `pos.order` model, and then generate+send the report when the order is saved?

remytms (approved):
LGTM

remytms:
@ivantodorovich Test has been changed by @huguesdk . It's now green. :)

pos_mail_receipt/static/src/css/print.css
ivantodorovich: This comment is still pending AFAICS.

@legalsylvain has a point here, this css is not scoped so it'll be applied for any report and that's not good.

ivantodorovich (changes requested):
Thanks!! Indeed, tests are green now 🙏🏻 
Could you take a look at this comment?

pos_mail_receipt/models/pos_order.py
victor-champonnois: Done

pos_mail_receipt/static/src/css/print.css
victor-champonnois: Done, I've encapsulated the CSS

victor-champonnois:
@ivantodorovich @legalsylvain I replied to the comments, could you re-review ?

ivantodorovich (changes requested):
Looks good to me 👍🏻 thanks a lot

Missing:
- CI is red, possibly a typo somewhere
- Squash your commits

victor-champonnois:
@ivantodorovich done :)

ivantodorovich (approved):
Thanks!! 🙏🏻

cesarlr:
> @ivantodorovich The merge process could not be finalized, because command `twine upload --repository-url https://upload.pypi.org/legacy/ -u __token__ odoo12_addon_pos_mail_receipt-12.0.1.0.0.99.dev7-py3-none-any.whl` failed with output:
> 
> ```
> Uploading distributions to https://upload.pypi.org/legacy/
> Uploading odoo12_addon_pos_mail_receipt-12.0.1.0.0.99.dev7-py3-none-any.whl
> �[?25l
> �[2K�[35m  0%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m0.0/36.6 kB�[0m • �[36m--:--�[0m • �[31m?�[0m
> �[2K�[35m  0%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m0.0/36.6 kB�[0m • �[36m--:--�[0m • �[31m?�[0m
> �[2K�[35m  0%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m0.0/36.6 kB�[0m • �[36m--:--�[0m • �[31m?�[0m
> �[2K�[35m  0%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m0.0/36.6 kB�[0m • �[36m--:--�[0m • �[31m?�[0m
> �[2K�[35m100%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m36.6/36.6 kB�[0m • �[33m00:00�[0m • �[31m70.9 MB/s�[0m
> �[2K�[35m100%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m36.6/36.6 kB�[0m • �[33m00:00�[0m • �[31m70.9 MB/s�[0m
> �[2K�[35m100%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m36.6/36.6 kB�[0m • �[33m00:00�[0m • �[31m70.9 MB/s�[0m
> �[2K�[35m100%�[0m �[90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━�[0m �[32m36.6/36.6 kB�[0m • �[33m00:00�[0m • �[31m70.9 MB/s�[0m
> �[?25h�[33mWARNING �[0m Error during upload. Retry with the --verbose option for more details. 
> �[31mERROR   �[0m HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/          
>          The user 'OCA' isn't allowed to upload to project                      
>          'odoo12-addon-pos-mail-receipt'. See                                   
>          https://pypi.org/help/#project-name for more information.              
> ```

Hi! we own this module in pypi, I will try to transfer it to OCA

cesarlr:
I have already sent an invitation to OCA to own the project.

ivantodorovich:
I summon @sbidoul 🙏🏻 

We have an issue here. @OCA-git-bot said the merge process failed, but [it did merge it](https://github.com/OCA/pos/tree/12.0/pos_mail_receipt) 😓 

And now it seems the module is not pushed to pypi due to this conflict.

Is the solution offered by @cesarlr enough to solve it? (inviting OCA to the pypi project)
Alternatively, what could we do here to fix this?

sbidoul:
Is this a module that is transferred from CoopIT Easy to OCA?

victor-champonnois:
@sbidoul Yes the PR is made from Coop IT Easy's fork

sbidoul:
Ok, I got the PyPI invite to the OCA account. Let me see if I can accept it.

sbidoul:
I accepted the transfer to OCA and removed coopdevs from that project on PyPI.
Please do a dummy PR to force a publish (or wait until tomorrow that the main branch bot does it).

The fact that the bot did merge anyway is tracked in https://github.com/OCA/oca-github-bot/issues/221. If anyone want's to investigate that it's welcome :)

PR #438, opened 2020-01-10
ADD module pos_default_partner

bealdav:
- [ ] add test to check that add partner in create() have no change on price


legalsylvain (commented):
Hi @bealdav. Thanks for this module.

- I think that in the next step, set the partner should be set in JS part. Something like : in each new order, the customer is set by default. (and the user can change it if he want). It will avoid incorrect computation. (price, fiscal position, etc...) If you have no time for that, you can set it in the ROADMAP.rst section.

- About configuration, I was asking to me if it was more modular to set the partner on the ``pos.config`` and not on the ``res.company``. I see it more generic and maybe more intutitive. (all the PoS configuration is in pos.config, (almost) nothing on res.company)  
What do you think ?

Otherwise, LGTM. (Code review / no test).

And happy new year.

bealdav:
thanks for the review @legalsylvain 

The goal of the module is to not deal with frontend just backend to reduce complexity. I have to describe it in readme. The choosen partner here must not be taken account for price computing.

About config, I think there are a lot config to maintain in all pos.config then for a real default partner, I think it can be sufficient on company.
What are the use cases where you need to have a fake/anonymous and different customers in your pos.config ? In my use case it'll be the same for all companies.

legalsylvain:
> About config, I think there are a lot config to maintain in all pos.config then for a real default partner, I think it can be sufficient on company.
> What are the use cases where you need to have a fake/anonymous and different customers in your pos.config ? In my use case it'll be the same for all companies.

i don't have any use, case. It's for consistency reason (all the conf is in pos.config, nothing on res.company) and for access right reason. The same occurs for pricelist, etc... I never saw 2 pos config with 2 differents pricelist for the time being. 
But it's just a proposal. Not a blocking point. It can be done later, if wanted by some people.

> The goal of the module is to not deal with frontend just backend to reduce complexity. I have to describe it in readme. The choosen partner here must not be taken account for price computing.

Well, I'm not asking for a change. But please, set this evolution from python code to JS code in a roadmap section, as it will be more consistent (price computation), more user friendly (the user will see that this is the anonymous partner that is selected), the ticket with customer name will be the same, if reprint, etc...

thanks !

robinkeunen (approved):
I agree with Sylvain about putting the defaults in `pos.config`, otherwise LGTM.

bealdav:
About config, pratical example:
If you have a company with one hundred Pos config (it could be possible in the future), you have to set one hundred times defaut_partner_id field instead once in pos general config screen.
In my customer case there a lot of config fields in pos.config (not only OCA modules) and big scroll on the page. If I have only those which are different accross each pos it's easier to manage, I think.
Each configuration has it's own use (about pricelist an example with a different value could be found, not for now).
Moreover it make me think that if we could a module like https://github.com/OCA/product-attribute/tree/10.0/product_profile, a generic one for any models, it could be easier to manage a lot of pos configs. But that's another story

legalsylvain (changes requested):
> If you have a company with one hundred Pos config

If you have on hundred PoS, you'll the same problem for all the other field of pos.config.
-> it's not a valid argument to move config that should be in pos.config in another place.

For your specific problem of your hundred PoS config, you have OCA modules that perfectly works like mass_editing :

https://github.com/OCA/server-ux/tree/12.0/mass_editing

kind regards.

hparfr (approved):
>  But it's just a proposal. Not a blocking point. It can be done later, if wanted by some people.

Why request a change then ?

legalsylvain:
I think that the the two following points should be written in the roadmap section /  know issues section : 
- implement code in Javascript code
- moving configuration in pos.config model. (as asked also by @robinkeunen)
I think that it is a Beta module with some limitations. No problem to merge this PR, but should be written in the roadmap.

thanks.

bealdav:
> -> it's not a valid argument to move config that should be in pos.config in another place.

You mean my argument didn't convinced you, not "It's not a valid argument" from an absolute point of view, I hope.

I think Pos app has not a wild card to merge all settings to pos.config.
Think about manufacturing app and mrp.workcenter model which embeds configs/settings even if res.config.settings include a manufacturing part. It can be applied to a lot of models.

I'm ok to add js comment in readme, but I think it's an error for moving general settings to pos.config. What is the gain ?

Sorry if failed to convince you but it doesn't mean it's not a valid argument.

But as PSC here, you can refuse and ask to not put this module to oca, no offense for me.

flotho:
Let's give it a troll ;-)
What about having a default setting for the company AND for the pos.
each time you create a pos, teh dafult config is updated with the company's one and when you sepcifically want to change it for a pos, you can do this easily .

bealdav:
Ahah, you'll finally reconcile us !? 

But my customer no more require this feature for its project, then I don't want go on this PR.

This PR is quite empty and Sylvain would add features (js, etc.).

Then I let you decide if I close it or let somebody go on inside ?

jarroyomorales:
@bealdav Hi! I'm interested in this module, I can supersede it if you want.

Regarding where should the `default_partner` be defined, I agree with @legalsylvain that it is more modular to have it on `pos.config`.

PR #439, opened 2020-01-17
[10.0] [FIX] pos_timeout: effective timeout should depend on the number of o…

nilshamerlinck:
…rders


legalsylvain (commented):
Hi @nilshamerlinck. Thanks a lot for this improvment. 
LGTM for me, regarding the code.
Regarding documentation, I think that you should change string and help of the field pos_order_timeout because, you are changing the meaning of the field. (also in the readme file)

what do you think ?

https://github.com/nilshamerlinck/pos/blob/10.0_pos_timeout/pos_timeout/models/pos_config.py#L13

nilshamerlinck:
Hi @legalsylvain you're right.

Improved documentation and took the occasion to move it to the new `oca-gen-addon-readme` format.

legalsylvain:
Thanks @nilshamerlinck !

/ocabot merge

PR #301, opened 2018-10-27
11.0 mig pos payment terminal

cvinh:

PR #440, opened 2020-01-18
12.0 mig pos payment terminal

legalsylvain:
from 10.0

CC : @quentinDupont


pos_payment_terminal/models/account_journal.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import models, fields
+
+
+class AccountJournal(models.Model):
+    _inherit = 'account.journal'
+
+    payment_mode = fields.Selection(
+        [('card', 'Card'), ('check', 'Check')], 'Payment Mode',
+        help="Select the payment mode sent to the payment terminal")
```
ivantodorovich, line 14: There were some interesting comments by @chienandalu and @rousseldenis in https://github.com/OCA/pos/pull/301 about either renaming this to `pos_terminal_payment_mode` or integrating it with `payment.mode` model

https://github.com/OCA/pos/pull/301#discussion_r277226621

Maybe it's worth exploring it? What do you think?

pos_payment_terminal/static/src/js/devices.js
```diff
@@ -0,0 +1,70 @@
...
+                    line = lines[i];
+                }
+            }
+
+            var data = {'amount' : line.get_amount(),
+                        'currency_iso' : currency_iso,
+                        'currency_decimals' : currency_decimals,
+                        'payment_mode' : line.cashregister.journal.payment_mode,
+                        'order_id': order.uid};
+            //console.log(JSON.stringify(data));
```
ivantodorovich:
```suggestion
```

ivantodorovich (commented):
Hi @legalsylvain ! 
Thanks for this PR !! :)

I left a comment I took from https://github.com/OCA/pos/pull/301, I think it might be worth exploring. I'm approving anyways cause even if we don't improve that, this looks good :+1: !

You should probably squash your 2 migration commits into a single one though

Considering there's not much activity in https://github.com/OCA/pos/pull/301, I guess it should be pretty easy to backport this to 11.0 after it's merged, don't you think?

ivantodorovich (approved):

pos_payment_terminal/models/account_journal.py
legalsylvain: Hi. 
thanks for pointing this comments. 
- for the time being, I'm in favor to rename the field as proposed by @chienandalu.

- The proposal of @rousseldenis is not valid I think, because this field has nothing to do with payment mode of the module ``account_payment_mode`` module. (except the name ;-)). The value of the selection fields only depends on the hardware you have. (credit cart terminal, or check printer).

ivantodorovich: Sounds good to me :+1:

ivantodorovich:
Tests are failing. I think there's another a "payment_mode" reference there, to be renamed

ivantodorovich:
This one too, I think:
https://github.com/legalsylvain/pos/blob/12.0-mig-pos_payment_terminal/pos_payment_terminal/tests/test_transactions.py#L14

legalsylvain:
on va y arriver ;-)

legalsylvain:
Victoire !

/ocabot merge

ivantodorovich:
Since we're at it, I take the opportunity to ask a question about migrations.

In cases like these, where a field name was changed during a migration. Should we add a migration script somewhere? Here, on this module, or maybe in openupgrade repository?

PR #442, opened 2020-01-20
Normalize accented characters in product search

Fkawala:
This PR provides normalization for the POS product search string. The normalization allows user to type indistinctively "café" or "cafe". The normalization applies to all the [Diacritical marks](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks) in range \u0300-\u036f and _œ_.


hbrunn:
this kind of thing should be done globally with [`unaccent`](https://github.com/odoo/odoo/blob/13.0/odoo/osv/expression.py#L59)

Fkawala:
@hbrunn thank you for the comment.

Your reference suggests doing backend configuration of the pgSQL server for it to include an unaccent primitive. This is super interesting and would alleviate the problem we have in inventory product search. 

However, I believe that the product search in POS is done "offline" (ie. synch-ed once at page loading time, see https://github.com/odoo/odoo/blob/9.0/addons/point_of_sale/static/src/js/db.js#L384), hence the pgSQL won't do much here.

Despite that, some users can install addons but are not sysadmin of the machine running their odoo instance and won't be able to update/configure pgSQL. Do you have the same reading of the situation?

hbrunn:
right, thanks for the elaboration. I must admit I only read the title. Just looked through this part and it looks good to me.

Fkawala:
This PR contains a commit irrelevant to the problem it solves, I opened a new, cleaner, PR #444

robinkeunen:
I'll review #444, but you could also have used the following sequence and dropped the irrelevant commit :-)

```
git fetch --all
git rebase -i oca/9.0
# opens an editor, replace "pick" by "drop" on the irrelevant commit and close
git push --force 
```

PR #445, opened 2020-01-21
[12.0] FIX pos_partner_firstname allowing non admin users to get partner_names_order from POS frontend

eLBati:
Otherwise they would get access error on ir.config_parameter

Steps:

 - Create a user POS manager without administration rights
 - Using that user, open POS frontend


mrcast (approved):
Functional test ok.
Thanks

legalsylvain (approved):
thanks for the patch.

/ocabot merge minor

eLBati:
I think you need to run `/ocabot merge minor` in a separate message.
Thanks

PR #436, opened 2020-01-08
[12.0] [ADD] pos_tare

legalsylvain:
propose a new module ``pos_tare`` that allow to enter a tare, when weighting products in the point of sale. The UI will so compute a net weight, depending on the tare and the gross weight.

pos_tare

Thanks for your review.

CC : @quentinDupont

PR #447, opened 2020-01-28
[9.0] [ADD] sell loose goods using tare barcode label

Fkawala:
This add-on enables POS to read and print tare barcodes. We print a barcode tare label to sell loose goods in a Bring Your Own pot (BYOC) scheme.

The BYOC scheme has five steps:
1. The cashier weighs the pot and sticks the tare barcode onto the customer's pot.
2. The customer go and put loose goods into the labeled pot.
3. The cashier weighs the pot with loose goods inside. POS computes the price including the pot.
4. The cashier scans the tare barcode. POS get the pot weight from the barcode. POS subtracts the pot weight from the weight of the latest product. POS sets the billable price for the loose goods.
5. The customer pays.

This add-on adds a news screen to POS to print (web) the tare barcode labels. This add-on enables POS to read a tare barcode. Reading a barcode makes POS adjust the weight of the latest article in the order list. The new weight is equal to the total weight minus the tare weight. The price is updated accordingly to the weight change.


Fkawala:
Hello @robinkeunen, hello @legalsylvain, would you please have time to give a look at this PR? Thx!

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,199 @@
...
+            var weight_in_gram = weight * 10e2;
+            var weight_with_padding = '0'.repeat(padding_size) + weight_in_gram;
+            var padded_weight = weight_with_padding.substr(
+                weight_with_padding.length - padding_size);
+            var barcode_data = default_weight_prefix_id.concat(void_product_id,
+                padded_weight);
+            var checksum = this.ean13_checksum(barcode_data);
+            var barcode = barcode_data.concat(checksum);
+
+            console.log(barcode);
```
legalsylvain: hehe ;-) to remove.

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,199 @@
...
+                counterCheckSum += 1;
+                result += parseInt(s.charAt(counter), 10) * counterCheckSum;
+            }
+            var checksum = 10;
+            checksum -= result % 10;
+            return checksum % 10;
+        },
+        barcode_data: function (weight) {
+            var padding_size = 5;
+            var default_weight_prefix_id = "21";
```
legalsylvain: I think that it should be done with a ``barcode.nomenclature``, and not hardcoded.
Note : can be done in a second time, but should be written in the "Roadmap / Know issue" section.

pos_barcode_tare/models/pos_config.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+from openerp import models, fields
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+    tare_label_button = fields.Boolean(
```
legalsylvain: Generally, all the PoS options names are prefixed by iface.
Think that it should be applied here. (not a blocking point, just a Odoo convention)

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,199 @@
+odoo.define('barcode_tare', function (require) {
```
legalsylvain: the convention is : module_name.type (type = screens / models / etc...)

``Here  odoo.define('pos_barcode_tare.screens', function (require) {``

and you can rename the file pos_barcode_tare.js into screens.js.

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,199 @@
...
+                var self = this;
+                this._super();
+                this.pos.barcode_reader.set_action_callback(
+                    'weight',
+                    _.bind(self.barcode_weight_action, self));
+            },
+        });
+});
+
+odoo.define('tare-screen-button.button', function (require) {
```
legalsylvain: no need to redefine other namespace. (or maybe I missed something ? )

pos_barcode_tare/static/src/xml/pos_barcode_tare.xml
```diff
@@ -0,0 +1,47 @@
...
+            </div>
+        </div>
+    </t>
+
+    <t t-name="PosTareLabel">
+        <div class="pos-center-align">
+            <div class="pos-tare-paper">
+                <div class="pos-tare-label">
+                    <img t-att-src="'/report/barcode/EAN13/' + widget.get_barcode_data()" />
+                    <span class="caption">tare = <t t-esc="widget.get_weight()" />kg</span>
```
legalsylvain, line 32: maybe set the barcode also. (I saw other sytem where the barcode was written)

something like : 
                     <span class="caption"><t t-esc="widget.get_barcode_data()" /></span>

pos_barcode_tare/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,12 @@
...
+The BYOC scheme has four steps:
+    1. The cashier weights the container and sticks the tare bar code onto the customer's container.
+    2. The customer takes the desired quantity of the good s-he wants to buy.
+    3. The cashier weights the filled container and good, POS gives the corresponding price.
+    4. The cashier scans the tare bar code, POS removes the container's weight from the latest product of the order.
+
+This add-on adds a news screen to POS to (web) print the tare barcode labels. This add-on enables POS to read a tare
+barecode. When such a barcode is read, POS adjusts the weight of the latest article in the order list.
+The new weight is equal to the total weight minus the tare weight. The price is updated accordingly to the weight
+change.
```
legalsylvain: could you add a screenshot ? 

something like : 

    ..figure: ../static/description/point_of_sale_tare_screen.png

legalsylvain (changes requested):
Hi @Fkawala. 

thanks a lot for sharing this module.

some comments inline. no important blockings points. 

FYI, I wrote a module for V12, named ``pos_tare``. Complementary features. 
maybe we could imagine to merge that feature in a second step.

could you take a look, and make a review ? 
Thanks ! 
https://github.com/OCA/pos/pull/436

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: héhé indeed (^^) good catch.

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: I'll change that!

pos_barcode_tare/models/pos_config.py
Fkawala: I'll change that. I'd like to stick to conventions as much as possible.

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: Thank you for the namespace explanation! For the file name, I tried to follow the [OCA guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#17complete-structure) but I might have missed something.

robinkeunen:
@vvrossem @houssine78 fyi

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: Agreed! I'll try to fix that using the nomenclature in this version.

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: Yep, newbie mistake, I'll merge all into the screens namespace.

pos_barcode_tare/static/src/xml/pos_barcode_tare.xml
Fkawala: Not sure about this. To my understanding, the rationale for having the barcode data written is to enable the operator to manually input the barcode if the scanner fails to decode it. Do you think this applies to POS? If so I'm happy to print the barcode. 

Let me know what you think!

pos_barcode_tare/readme/DESCRIPTION.rst
Fkawala: Will do :)

pos_barcode_tare/static/src/xml/pos_barcode_tare.xml
Fkawala: Hello @legalsylvain, 

Thank you for the thoughtful review!

I'll (try to) do the same for #436. I agree with you #447 and #436 are complementary and should merge at some point! 

Thank you,
François

> Hi @Fkawala.
> 
> thanks a lot for sharing this module.
> 
> some comments inline. no important blockings points.
> 
> FYI, I wrote a module for V12, named `pos_tare`. Complementary features.
> maybe we could imagine to merge that feature in a second step.
> 
> could you take a look, and make a review ?
> Thanks !
> #436

Fkawala:
Hello @legalsylvain, 

Thank you for the thoughtful review! I'll (try to) do the same for #436. 

I agree with you #447 and #436 are complementary and should merge at some point! 

Thank you,
François

> Hi @Fkawala.
> 
> thanks a lot for sharing this module.
> 
> some comments inline. no important blockings points.
> 
> FYI, I wrote a module for V12, named `pos_tare`. Complementary features.
> maybe we could imagine to merge that feature in a second step.
> 
> could you take a look, and make a review ?
> Thanks !
> #436

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: This is solved, not sure I used the best possible way.

Fkawala:
Howdy @robinkeunen @vvrossem @houssine78 a review would be much appreciated.

houssine78:
Hi @Fkawala Thanks for this very nice contribution. I don't know this POS part enough to give relevant comment. However the whole module seems to be good. I didn't went through the js code sorry. @vvrossem will probably give you a better review than mine.

Fkawala:
Hi @legalsylvain, would you have time to take a look? I updated the code using the UOM as you do in #436. You'll need to define a "Kg" UOM if you plan to test that using the debug module. The UOM matching is case sensitive and the debug panel sets the weight to "Kg" whereas it should be "kg" (see https://github.com/odoo/odoo/blob/9.0/addons/point_of_sale/static/src/js/devices.js#L380).

Fkawala:
Hi @robinkeunen! I would love to get that addon running in our shop. This addon would massively speed up the checkout process. This is one of our highest pain-point right now. Would you mind to check if someone from coopITEasycould can review this? Thank you in advance!

vvrossem:
> Hi @legalsylvain, would you have time to take a look? I updated the code using the UOM as you do in #436. You'll need to define a "Kg" UOM if you plan to test that using the debug module. The UOM matching is case sensitive and the debug panel sets the weight to "Kg" whereas it should be "kg" (see https://github.com/odoo/odoo/blob/9.0/addons/point_of_sale/static/src/js/devices.js#L380).

Hi @Fkawala, could you please explain me what happens if you put 'kg' instead of 'Kg' ?
Many thanks

Fkawala:
@vvrossem, thank you for looking into this PR. Please find a more detail explanation below. 

### How the Addon works?
The add-on saves weight on the tare label using the "kg" unit of measure (UOM). This UOM should exist for the add-on to work. The UOM is fetched from the POS and matched by its name (case sensitive). See [here](https://github.com/OCA/pos/pull/447/files#diff-235382a6bcecc79a91e2a4576a7f059fR18) and [there](https://github.com/OCA/pos/pull/447/files#diff-235382a6bcecc79a91e2a4576a7f059fR65%20) for the code responsible to find the UOM based on UOM name.

When the electronic scale sends the weight, it comes with the unit name, for instance, "kg". The add-on fetches the UOM corresponding to the unit name. The UOM that matches the unit name sent by the scale is the "__scale UOM__". The add-on converts the scale weight from the **scale UOM** to the "kg" UOM. At this point, we're ready to create a tare label. When we read the tare label, we read a tare weight saved in "kg" UOM. We convert that tare weight into the **product UOM** so that we can compute the net weight in the product UOM.

In summary, we have three UOM:

- Product UOM (fetch from the last product in order list)
- Tare label UOM (which is always equal to the default "kg" reference UOM)
- Electronic scale UOM (matched from the unit name sent by the electronic scale)

Today all the electronic scales that are supported (ADAM / Mettler-Toledo) use the "kg" UOM. However, this architecture would allow us to use an electronic scale that use another UOM, as long as this UOM is defined.

### What can be the problem in debug?
My note on the "**K**g" UOM is there because of the particular behavior of the "debug panel". As you can see [here](https://github.com/odoo/odoo/blob/9.0/addons/point_of_sale/static/src/js/devices.js#L380) the debug pannel forges weights with a non-standard unit name "**K**g". If you do run this code using the debug panel to moke the electronic scale, then the add-on will search for the "**K**g" UOM. This UOM does not exist by default. Hence the add-on would fail to create a tare label.

I hope this is clear enough, if not, please let me know!

pos_barcode_tare/models/pos_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+    iface_tare_label = fields.Boolean(
+        'Show tare label button',
+        help="Print tare labels with this POS"
+    )
+
+    iface_tare_barcode_sequence_id = fields.Integer(
```
legalsylvain: Hi. I don't think it is the good way to do. (mainly because you hardcode 36 ID in the code !)

I think that a more modular approach is (mainly) : 

* to create a new **type** of barcode rule (ref : https://github.com/odoo/odoo/blob/12.0/addons/barcodes/models/barcodes.py#L184) named ``tare``
* create a new function barcode_tare_action like (per exemple) ``barcode_cashier_action`` (https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/static/src/js/screens.js#L76)
* add a new action of BarcodeReader (ref) https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/static/src/js/devices.js#L501

Note : regarding your data, I don't think that 21 is a good prefix as it is reserved for other rule in Odoo "weightable products". See : https://github.com/odoo/odoo/blob/12.0/addons/stock/data/default_barcode_patterns.xml#L10
This rule should be used for weightable product, when .... = little internal product code and NNDDD is the weight.

for that purpose, I think that a new type like this is the good solution : 
```
    <data noupdate="1">
         <record id="barcode_rule_tare model="barcode.rule">
            <field name="name">Tare</field>
            <field name="barcode_nomenclature_id" ref="barcodes.default_barcode_nomenclature"/>
            <field name="sequence">DONT_CARE</field>
            <field name="type">tare</field>
            <field name="encoding">ean13</field>
            <field name="pattern">XX00000{NNDDD}</field>
        </record>
    </data>
```
Where XX should not be in existing values.

kind regards.

legalsylvain (changes requested):
see comment regarding barcode nomenclature / rule.

pos_barcode_tare/models/pos_config.py
Fkawala: Thank you for the detailed explanation! I'll do that.
Cheers.

Fkawala: @legalsylvain this is done (re-using some of your code from `pos_price_to_weight`) as this is developed for v9.0.

pos_barcode_tare/views/pos_config_view.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+
+<odoo noupdate="1">
```
legalsylvain: - the view view_pos_config_form should be in a classical ``odoo`` tag. (without no update) because view should allways updated.

- the barcode rule should be in a file like /data/barcode_rule.xml in  ``odoo noupdate=1`` because this data should be customizable by the user. (changing pattern, etc...)

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,273 @@
...
+                this.lock_screen(false);
+            }
+        },
+        get_weight: function () {
+            if (typeof this.weight_in_kg === 'undefined') {
+                return this.default_tare_value;
+            }
+            return this.weight_in_kg;
+        },
+        ean13_checksum: function (s) {
```
legalsylvain: In Odoo core, there are some functions that could to be used, instead of rewriting it. For exemple, in barcodes modules : 
https://github.com/odoo/odoo/blob/13.0/addons/barcodes/static/src/js/barcode_parser.js#L54
odoo ``ean_checksum`` could replace your ``ean13_checksum``. Don't you think ? 
(I'm not sure however, because it's not the same algorithm.)

legalsylvain (approved):
- A little think to fix in a data files. (view + barcode.rule)
- A factorization suggestion. (not sure)

LGTM otherwise. (Code review / not test)

Thanks a lot for sharing this module.

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: Good idea, my function looks redundant. I'll fix this.

pos_barcode_tare/views/pos_config_view.xml
Fkawala: Thank you for the explanation, I will do so.

Fkawala: Done!

Fkawala:
@legalsylvain your guidance on this is very much appreciated. I learned a lot. thank you! I'll try to find time to port this addon for odoo 12.0.

legalsylvain:
> @legalsylvain your guidance on this is very much appreciated

your welcome, thanks !

>  I'll try to find time to port this addon for odoo 12.0.

OK ! Please ping me before beginning the port in V12. I'd like to talk with you about how to merge (or converge) features from my module ``pos_tare`` and your ``pos_barcode tare``.
Especially, I'd like to add gross weight and or net weight and or tare weight on each pos order line, to have the possibility to display the information on the bill, for the customer.

legalsylvain:
Hi @Fkawala : could you fix the conflicts ? 

tahnks !

Fkawala:
Hello @legalsylvain, which conflict are you thinking about? Regards, François.

Fkawala:
Howdy @legalsylvain! I worked out a way to display gross and net weight on each Orderline. I'll refactorize a little bit and then add this to this PR.

Fkawala:
Hello @legalsylvain, I committed code to get the tare weight displayed in the orderline.

legalsylvain:
> Hello @legalsylvain, which conflict are you thinking about? Regards, François.

I'm talking about that.

image

I think that locally, you should to something like : 
git pull --rebase origin 9.0

(origin should be the name of the OCA remote)

this should fix the problem (maybe with manual operation you'll have to do)

About new features : 

1) when I click on Print a tare label : and I set a weight, i have this error : don't you have ?

TypeError: from_unit is undefined
http://3408593-447-6ebf47.runbot1-2.odoo-community.org/pos_barcode_tare/static/src/js/pos_barcode_tare.js:31

2) otherwise : 
- very nice display : 

image

- good to have thought to border use case : 

image

Fkawala:
Hello @legalsylvain, git is not complaining about conflict, I do not understand why we do not have the same here.
image

About the `TypeError: from_unit is undefined` problem. Did you define the `kg` UOM? It is necessary to define this UOM because the debug forged scale measures are sent in `kg` UOM (at not `Kg` UOM).

Thanks for the test and feed backs.

legalsylvain:
Hi. 
-> Well, about conflicts, it's quite weird. No problem, we'll see when we'll merge the PR.

-> about the error, I just tested on runbot without any custom settings. If setting is required, I think that we should add demo data, with correct settings, to makes the module working in a demo mode, without any configuration. don't you think ?

kind regards.

Fkawala:
@legalsylvain 

I'll add the demo data (ie. missing UOM) that's a good idea. 

How should we proceed to merge both our modules?

Would you be willing to do the migration work to get the result working on odoo 12.0? 

Thank you!

legalsylvain:
> I'll add the demo data (ie. missing UOM) that's a good idea.

nice ! 

> How should we proceed to merge both our modules? 
> Would you be willing to do the migration work to get the result working on odoo 12.0?


i propose to : 
1) merge this PR in 9.0, if you consider it as finished. 9.0 serie is quite old (unsupported recently)

2) integrate on https://github.com/OCA/pos/pull/436 the feature present in this module. do you have a deadline ? I'd like to do it, but I'm currently quite busy in the next weeks.

kind regards.

Fkawala:
It works for me! The plan looks fine, thank you. 

We're running odoo 9.0. It won't change in the coming months. Hence, there is no rush to get that
working on 12.0 (for us at least). 

Would you happen to know someone that could do the second review and approve the merge of this PR?

legalsylvain:
@PierrickBrun could you take a time to make a little functional review on that PR ?
Thanks !

Fkawala:
Hello @legalsylvain and @PierrickBrun. This is ready to be reviewed and hopefully merged. Thank you in advance!

PierrickBrun (approved):

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,359 @@
...
+
+    // This configures read action for tare barcode. A tare barcode contains a
+    // fake product ID and the weight to be subtracted from the product in the
+    // latest order line.
+    screens.ScreenWidget.include(
+        {
+            barcode_tare_action: function (code) {
+                try {
+                    var order = this.pos.get_order();
+                    var last_order_line = order.get_last_orderline();
```
vvrossem, line 70: I suggest to replace `order.get_last_orderline()` with `order.get_selected_orderline()`. It allows the cashier to scan the tare of a pot after another orderline has been added :
order get_selected_orderline

Fkawala:
Thanks for the review @PierrickBrun! Could we proceed forward and merge this? Best regards

pos_barcode_tare/static/src/js/pos_barcode_tare.js
```diff
@@ -0,0 +1,359 @@
...
+            this.weight_barcode_prefix = this.get_barcode_prefix();
+            // Setup the proxy
+            var queue = this.pos.proxy_queue;
+            // The pooling of the scale starts here.
+            var self = this;
+            queue.schedule(function () {
+                return self.pos.proxy.scale_read().then(function (weight) {
+                    try {
+                        self.set_weight(weight);
+                    } catch (error) {
```
vvrossem, line 129: could the syntax `scale_read().then(function (onFulfilled){...}, function (onRejected){...})` be an alternative for UOM management? `scale_read` in `point_of_sale.devices` uses it

pos_barcode_tare/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+Install this add-on and configure the point of sale where you want to be able to use the tare barecode. The label printing is done using web print. To streamline the label printing it is advised to use the silent printing mode (firefox) or the kiosk printing (chrome).
+
+The command line to start a chrome base browser in kiosk mode with silent printing looks like:
+
+``chromium-browser --use-system-default-printer --kiosk --kiosk-printing http://localhost:8069/``
```
vvrossem, line 5: I would also add a _Configuration_ section explaining to check the _show tare label button_ checkbox in _Loose good options_ for the functional users

pos_barcode_tare/static/src/xml/pos_barcode_tare.xml
```diff
@@ -0,0 +1,86 @@
...
+                            <t t-esc="line.get_discount_str()" />%
+                        </em>
+                        discount
+                    </li>
+                </t>
+                <t t-if="line.get_tare_str() !== '0'">
+                    <li class="info">
+                        tare = <t t-esc="line.get_tare_str_with_unit()" />
+                    </li>
+                </t>
```
vvrossem, line 81: Since the tare and net weight is specified on the orderline, I wonder if the gross weight should be included as well for legal reasons

Fkawala:
Thank you @legalsylvain !

vvrossem (approved):
Hi @Fkawala ! First of all, thank you for your work !
I've tested and reviewed the module. I have minor suggestions that can be added in the roadmap if the community agrees :)

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: I agree this would be quite more flexible!

pos_barcode_tare/readme/USAGE.rst
Fkawala: I agree this can help and it is fairly straightforward.

pos_barcode_tare/static/src/xml/pos_barcode_tare.xml
Fkawala: We can add it, it does not cost much. But I think this would be less readable. I'm not too worried about legal compliance since the whole POS scaling hasn't been certified (at least not in UE).

pos_barcode_tare/static/src/js/pos_barcode_tare.js
Fkawala: I do not get the rationale here, could you please explain it to me? Thank you.

vvrossem: On a first opinion, I was thinking it would be better to move the UOM management in an overridden `scale_read` function, but my second opinion is that it's better to handle this on the `TareScreenWidget` and not in the `ProxyDevice`

Fkawala:
@vvrossem What would the procedure to ship the changes you suggested to do (now that this PR has been merged)? Thanks!

legalsylvain:
@vvrossem could you make a new PR agains 9.0 with all the nice improvments you propose ?
Thanks

PR #451, opened 2020-02-07
[12.0][ADD] pos_order_return_traceability

dalonsod:
This module adds full return traceability to POS frontend

-------------

Current issues:
- prevent order list caching for orders that already changed (because of becoming non-returnable during the POS session). It only works when changing to a new session.
- Zero lines can be generated when selecting an order with non-returnable lines.
- ~~Improve test support~~


legalsylvain:
1) thanks for the general design of the modules. 

2) I tested on runbot and i have an error. Step to reproduce : 
- make an order and confirm
- select the order to make a refund
- change the quantity to return. for exemple -1 --> -5.

I have the following error : 

```
TypeError: self.gui is undefined
http://3407959-451-b5fb36.runbot2-2.odoo-community.org/pos_order_return_traceability/static/src/js/widgets.js:70
Traceback:
```

legalsylvain (changes requested):
see comments

dalonsod:
> 
> 
> see comments

Shit! I've force-pushed before looking at these coments, and they're no longer available, I'm sorry :(
Anyway, I fixed the error described above and improved test coverage.

legalsylvain (approved):
Hi @dalonsod. 
first, thanks for the Changes. Code review & test : LGTM ! Thanks ! 

Work like a charm.

image


Just a remarks : Could you have, in the ``pos_order_return module`` a ``ROADMAP.rst`` file in the ``readme`` folder, containing something like : 

    When migrating the module ``pos_order_return`` in version > 12.0 please merge both modules
    ``pos_order_return`` and  ``pos_order_return_tracaeability`` into a single module.

As I think that it will be useless to have two modules. (the back office one became quite useless in recent Odoo version).

kind regards.

dalonsod:
> Just a remarks : Could you have, in the `pos_order_return module` a `ROADMAP.rst` file in the `readme` folder, containing something like :
> 
> ```
> When migrating the module ``pos_order_return`` in version > 12.0 please merge both modules
> ``pos_order_return`` and  ``pos_order_return_tracaeability`` into a single module.
> ```

Done!

lromero-solvos (approved):
LGTM

PR #453, opened 2020-02-11
[MIG][12.0] pos_empty_home

ivantodorovich:


pos_empty_home/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,13 @@
...
+is selected. (when root category is selected)
+
+This module is interesting to avoid to display useless products, when user did
+not yet realized a search. (Default displayed products are just the first
+eighty products)
+
+It can limits errors to avoid users to select undesired product.
+
+
+.. image:: /pos_empty_home/static/description/pos_empty_home.png
```
legalsylvain: Replace by ../static/description/pos_empty_home.png

This path will work in all cases.

pos_empty_home/i18n/fr.po
```diff
@@ -0,0 +1,28 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* pos_empty_home
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 9.0c\n"
```
legalsylvain: Could you update this fr po file for 12.0 serie ?

pos_empty_home/static/src/js/db.js
```diff
@@ -0,0 +1,26 @@
...
+
+
+odoo.define('pos_empty_home.db', function (require) {
+    "use strict";
+
+    var PosDB = require('point_of_sale.DB');
+
+    PosDB.include({
+        get_product_by_category: function(category_id){
+            console.log(this.config);
```
legalsylvain: Hehe, the most reliable tools to dev Odoo PoS module ! Could you remove ?

legalsylvain (approved):
Some minors points. The rest looks good. Thanks ! 
(Code review / no test)

pos_empty_home/i18n/fr.po
ivantodorovich: Isn't this generated by travis? :thinking:

ivantodorovich:
done! :)

legalsylvain (changes requested):
Just tested on runbot.
problem 1) when no search is done, a scroll bar is displayed. I think a div is too big. (minor problem)
problem 2) when a search is done, the text "To select a product, please scan a barcode or search products by category or by name" is not hidden. (tested on Chromium + firefox)

could you check ?

ivantodorovich:
Re 1: you mean this, right?
scrollbar

(fixed here: https://github.com/OCA/pos/pull/453/commits/69ebb2ea7c360e7dbbe70104f36757af459558e1) 

------


Re 2: I'm not sure I understand. I tried to search and it works fine IMO:
product_not_hidden

legalsylvain:
problem 1 : solved. Thanks.

problem 2 : I'm talking about that. The text is not hidden :
image

context : 
- last runbot build / Ubuntu 18.04 with debug=assets to load last js / css files.
- fail on chromium 79.0.3945.130 
- fail on Firefox 72.0.2 (64 bits)

that's weird that you don't have this bug.

ivantodorovich:
It's because of this module: https://github.com/OCA/pos/blob/2fece129090177bb0d63e4cb4672098cfb05c13d/pos_fix_search_limit/static/src/js/pos_fix_search_limit.js#L19

It's overloading the method without calling super().

I've uninstalled that module on runbot, and now it works.

ivantodorovich:
> It's because of this module:
> 
> https://github.com/OCA/pos/blob/2fece129090177bb0d63e4cb4672098cfb05c13d/pos_fix_search_limit/static/src/js/pos_fix_search_limit.js#L19
> 
> It's overloading the method without calling super().
> 
> I've uninstalled that module on runbot, and now it works.

I think this should be fixed in `pos_fix_search_limit`.
Since it's completely overloading the method, the assets view that loads that js file should be given a higher priority, so that it always loads first right after odoo

hparfr:
Hi, 
Not calling super in pos_fix_search_limit is intended and there is not a better place to do it.

About what you are going to achieve here, you can do the following : 
```js
 set_product_list: function(product_list){
        this._super(product_list);
        if (product_list.length)
           $(this.el.querySelector('.product-list-empty-home')).show()
        else
            $(this.el.querySelector('.product-list-empty-home')).hide()
    },
```

ivantodorovich:
> Hi,
> Not calling super in pos_fix_search_limit is intended and there is not a better place to do it.
> 
> About what you are going to achieve here, you can do the following :
> 
> ```js
>  set_product_list: function(product_list){
>         this._super(product_list);
>         if (product_list.length)
>            $(this.el.querySelector('.product-list-empty-home')).show()
>         else
>             $(this.el.querySelector('.product-list-empty-home')).hide()
>     },
> ```

:+1: nice! Thanks!

It's implemented in https://github.com/OCA/pos/pull/453/commits/2b02aa05f3279113f1f387bb1b15c377bb839dcc

legalsylvain (approved):
Thanks @hparfr for your help.

LGTM. code review. no test.

dalonsod (approved):
Tested on runbot (with `pos_fix_search_limit` uninstalled, of course), 👍

ivantodorovich:
> Tested on runbot (with `pos_fix_search_limit` uninstalled, of course),

actually since last commit, it works even with `pos_fix_search_limit` installed :+1: 
thanks to @hparfr suggestion: https://github.com/OCA/pos/pull/453#issuecomment-588816731

PR #454, opened 2020-02-14
12.0 mig pos quick logout

ajayjaviya:
Migrate pos quick logout to version 12


legalsylvain:
Why closing ?
Regards.

PR #430, opened 2019-12-20
[12.0] pos_partner_firstname

eLBati:
From https://github.com/OCA/pos/pull/427

PR #459, opened 2020-02-18
[12.0][ADD] pos_ticket_partner_firstname

ivantodorovich:
On the POS ticket, prints only the salesman's firstname.

receipt


pos_ticket_partner_firstname/static/src/xml/pos.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<templates id="template" xml:space="preserve">
+
+    <t t-extend="PosTicket">
```
legalsylvain: until Odoo 13 (I think) there are two ticket format. ``PosTicket`` and the ``XmlReceipt``. 
could you handle both cases in your module? 

https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/static/src/xml/pos.xml#L736

if it's not in the scope of your developpement, please add a know issues in the roadmap file to mention that this module does'nt work in a non proxy mode.

thanks !

legalsylvain (commented):
Hi @ivantodorovich thanks for your contribution.

a remarks, regarding the ticket. otherwise LGTM.
I was asking to myself if if was not better to add this option in a existing glue module https://github.com/OCA/pos/tree/12.0/pos_partner_firstname introduced by https://github.com/OCA/pos/pull/430


what do you think @eLBati ?

kind regards.

eLBati:
@legalsylvain `pos_partner_firstname` handles first name of customers, the module of this PR handles first name of salesman. I think they have different scopes.

I would rather suggest to rename this module to `pos_ticket_salesman_firstname`

legalsylvain:
OK @eLBati. Thanks for your point of view !

ivantodorovich:
Thanks @eLBati !
I agree, `pos_ticket_salesman_firstname` is more clear :+1: 
I've renamed the module

pos_ticket_partner_firstname/static/src/xml/pos.xml
ivantodorovich: It's already handled, by overloading `export_for_printing` on the js file :)
`XmlReceipt` gets the cashier.name from that function

legalsylvain: thanks !

legalsylvain (approved):
@elbati, is it now OK for you ? 

@ivantodorovich : thanks. LGTM. (code review / no Test)

eLBati:
@ivantodorovich could you squash the 2 commits into 1?

eLBati (changes requested):
Testing on runbot with `Print only the salesman's first name` set, the name of the user is not displayed:

image

@ivantodorovich do you know why?

ivantodorovich:
> Testing on runbot with `Print only the salesman's first name` set, the name of the user is not displayed:
> 
> image
> 
> @ivantodorovich do you know why?

My bad. Forgot to change the assets view after renaming the module.
It should be okay now

ivantodorovich:
> @ivantodorovich could you squash the 2 commits into 1?

done :+1:

eLBati (approved):
Thanks

PR #460, opened 2020-02-18
[12.0][FIX] pos_fix_search_limit: priority asset view change

dalonsod:
Related to https://github.com/OCA/pos/pull/453#issuecomment-585172722

cc @legalsylvain @ivantodorovich

### DIFF
--- a/pos_fix_search_limit/README.rst
+++ b/pos_fix_search_limit/README.rst
@@ -84,6 +84,7 @@ Contributors
 ~~~~~~~~~~~~
 
 * Raphaël Reverdy <raphael.reverdy@akretion.com> (https://www.akretion.com)
+* David Alonso <david.alonso@solvos.es>
 
 Other credits
 ~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


legalsylvain (approved):
LGTM but I have not a deep view of this module, and why the call of super() was not present. 

@hparfr : could you take a look ? 

thanks !

ivantodorovich:
Calling super will have this module re-render the product list.
First with the full list of products (`_super()`), and then with the limit.

It looks ok but in terms of performance it kinda beats the point of limiting the number of products. It would be even worst than not having this module installed

I think we don't need to call `_super`, we cause we do want to completely override this function.
But we need to make sure this happens before any other module tries to do it. So that when they do, and they call `_super`, they get to ours.

I believe this can be achieved by changing the view priority in assets.xml, to control the order js files are loaded.

hparfr (changes requested):
No, not calling super is intended because this module change how the rending should be done.

dalonsod:
@hparfr OK... So, are you agree with @ivantodorovich , should the assets view priority be changed? To a lower, or a higher one?

ivantodorovich:
Setting priority=1 will have this script load before the others, and making it possible for other modules to extend this method.

~~I'm 99% sure about this~~. 
I just tested it locally. 100% sure

dalonsod:
> Setting priority=1 will have this script load before the others, and making it possible for other modules to extend this method.

Done!

ivantodorovich (approved):

legalsylvain:
@hparfr : now ok for you ? 
thanks !

hparfr (approved):

legalsylvain:
thanks all !

PR #461, opened 2020-02-21
Add _t var

NathanQj:
Missing var called below in the err msgs - throws error messages without it


legalsylvain (approved):
Thanks a lot @NathanQj. 

Fast merging bugfix.

PR #463, opened 2020-03-02
[12.0] ADD pos_supplierinfo_barcode: Search products by supplier barcode

eLBati:
Depends on https://github.com/OCA/stock-logistics-barcode/pull/257

### DIFF
--- a/pos_supplierinfo_barcode/__manifest__.py
+++ b/pos_supplierinfo_barcode/__manifest__.py
@@ -1,0 +1,26 @@
+# Copyright 2020 Lorenzo Battistini @ TAKOBI
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+{
+    "name": "Point of sale - Supplier barcodes",
+    "summary": "Search products by supplier barcode",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Point Of Sale",
+    "website": "https://github.com/OCA/pos",
+    "author": "TAKOBI, Odoo Community Association (OCA)",
+    "maintainers": ["eLBati"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "auto_install": True,
+    "preloadable": True,
+    "depends": [
+        "product_supplierinfo_barcode",
+        "point_of_sale",
+    ],
+    "data": [
+        "views/assets.xml",
+    ],
+    "demo": [
+    ],
+}
--- a/pos_supplierinfo_barcode/models/product.py
+++ b/pos_supplierinfo_barcode/models/product.py
@@ -1,0 +1,17 @@
+import json
+from odoo import models, fields, api
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    # technical field used in POS frontend
+    supplier_barcode_json = fields.Char(
+        "Supplier barcode list", readonly=True,
+        compute="_compute_supplier_barcode_json")
+
+    @api.multi
+    def _compute_supplier_barcode_json(self):
+        for t in self:
+            supplier_barcode_json = [x for x in t.mapped('seller_ids.barcode') if x]
+            t.supplier_barcode_json = json.dumps(supplier_barcode_json)
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,2 +1,3 @@
 queue
 partner-contact
+stock-logistics-barcode
--- a/pos_supplierinfo_barcode/__init__.py
+++ b/pos_supplierinfo_barcode/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/pos_supplierinfo_barcode/models/__init__.py
+++ b/pos_supplierinfo_barcode/models/__init__.py
@@ -1,0 +1,1 @@
+from . import product
--- a/pos_supplierinfo_barcode/readme/CONFIGURE.rst
+++ b/pos_supplierinfo_barcode/readme/CONFIGURE.rst
@@ -1,0 +1,1 @@
+Open product form and click purchase tab. In Vendors lines, set Barcode field.
--- a/pos_supplierinfo_barcode/readme/CONTRIBUTORS.rst
+++ b/pos_supplierinfo_barcode/readme/CONTRIBUTORS.rst
@@ -1,0 +1,1 @@
+* Lorenzo Battistini (https://takobi.online)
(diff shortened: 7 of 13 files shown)


pos_supplierinfo_barcode/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "summary": "Search products by supplier barcode",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Point Of Sale",
+    "website": "https://github.com/OCA/pos",
+    "author": "TAKOBI, Odoo Community Association (OCA)",
+    "maintainers": ["eLBati"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
```
legalsylvain, line 14: I think that such module should be auto_installable to True.
Don't you think ?

pos_supplierinfo_barcode/models/product.py
```diff
@@ -0,0 +1,20 @@
+import json
+from odoo import models, fields, api
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    # technical field used in POS frontend
+    supplier_barcode_list = fields.Char(
```
legalsylvain: As it is not a python list, i think that a name like supplier_barcodes_json could be more explicit. Don't you think ?

pos_supplierinfo_barcode/models/product.py
```diff
@@ -0,0 +1,20 @@
...
+        "Supplier barcode list", readonly=True,
+        compute="_compute_supplier_barcode_list")
+
+    @api.multi
+    def _compute_supplier_barcode_list(self):
+        for t in self:
+            supplier_barcode_list = []
+            for s in t.seller_ids:
```
legalsylvain: [ x for x in t.mapped('seller_ids.barcode') if x] could be more synthetic, and quick to process. Don't you think ?

legalsylvain:
Some remarks inline. 
Thnaks otherwise for this contributions !

Side question : what is the use case when a same product has different barcodes, depending on the supplier ? I never have this situation in retail context.

eLBati:
> Some remarks inline

I think I did not receive them.

> what is the use case when a same product has different barcodes, depending on the supplier ? I never have this situation in retail context

@legalsylvain retailer may have several suppliers for the same product, buying it from one or another according to some criteria.
Then, assuming the user will not modify the barcode used by the supplier, the product can be added to POS orders just scanning the barcode on the product (provided by the supplier).

legalsylvain (commented):
Sorry. Inline comments stayed in draft status.

mrcast (approved):
I approved the wrong PR I wanted to approve https://github.com/OCA/stock-logistics-barcode/pull/257

pos_supplierinfo_barcode/__manifest__.py
eLBati: Done, thanks

pos_supplierinfo_barcode/models/product.py
eLBati: Done, thanks

pos_supplierinfo_barcode/models/product.py
eLBati: Done, thanks

legalsylvain (approved):
thanks !

GSLabIt:
> @legalsylvain retailer may have several suppliers for the same product, buying it from one or another according to some criteria.
> Then, assuming the user will not modify the barcode used by the supplier, the product can be added to POS orders just scanning the barcode on the product (provided by the supplier).

@eLBati not getting the point, couldn't be better having different barcode per product instead of having unique barcode per supplier? What about if different supplier using same barcode for product? This is quite common situation.

eLBati:
> couldn't be better having different barcode per product

I don't see advantages with this approach, while I see the disadvantage of missing the information "which barcode is used by which supplier"

> What about if different supplier using same barcode for product

We could a (optional?) unique check on barcodes

GSLabIt:
> I don't see advantages with this approach, while I see the disadvantage of missing the information "which barcode is used by which supplier"

Thing is, barcode is related to product, not to supplier. Technically barcode is related to the manufacturer, so having barcode per supplier, imo, is not a good approach.


> We could a (optional?) unique check on barcodes


I don't think it is a good idea, because you could have different product with same barcode.

eLBati:
> Technically barcode is related to the manufacturer

@GSLabIt that is EAN, not generic barcode.
Barcodes can be generated by a company and used only internally and/or with their customers (that is the case of this PR).
This is more frequent in small companies.

legalsylvain:
I agree with @elbati on that point. Most of ean13 are bought and are international. But some small producers doesn't buy ean13. (Good or bad, it's the reality) and that design can introduce conflict between products, BTW.

In that use case, this module (and the stock barcode one) can be interesting and "fix" the reality. :-)

If you're not agree, better to talk on the other PR that introduce barcode on product supplier info. This PR just introduce a glue module for PoS.

Kind regards.

GSLabIt:
@legalsylvain well ..i don't agree. international barcodes (bought by GS1 authority) or local one (see weighted product starting with nomenclature 2.....{MMMDD}) are always per product, not per supplier,  but feel free to merge it

eLBati:
> are always per product, not per supplier

@GSLabIt see https://github.com/OCA/pos/pull/463#issuecomment-594436182
In that case, if my suppliers generates barcodes by themselves and I want to use their barcodes to track and sell the products I buy from them, what should I do?

GSLabIt:
@eLBati 2 cases:
1. you supplier has unique product, you don't need to have barcode in supplier info as it is a different product.

2. your supplier sell a product, exactly same product, you can buy from another supplier like tomatoes, you will decide what barcode will have the tomatoes in your store, not the supplier. So you can use multi barcode per product. 

Like i wrote before, barcodes are related to product not to supplier. Sorry, my fault, I really not understanding the use case.

eLBati:
> your supplier sell a product, exactly same product, you can buy from another supplier like tomatoes, you will decide what barcode will have the tomatoes in your store, not the supplier. So you can use multi barcode per product

You should think about companies which are just resellers of packaged products: their main activity is to buy products and retail them.

The use case is described here https://github.com/OCA/pos/pull/463#issuecomment-594436182

Changing barcode for every product in order to not use the suppliers' ones would be an additional cost and they have no reason to do it

eLBati:
Green travis now

PR #465, opened 2020-03-03
[12.0][MIG] pos_price_to_weight

ivantodorovich:


legalsylvain (approved):
Code review. No test.
Thanks.

pos_price_to_weight/static/src/js/models.js
```diff
@@ -0,0 +1,38 @@
...
+            var product = this.db.get_product_by_barcode(parsed_code.base_code);
+            if(!product){
+                return false;
+            }
+            var quantity = 0;
+            var price = parseFloat(parsed_code.value) || 0;
+            if (price !== 0 && product.price !== 0){
+                quantity = price / product.price;
+            }
+            selectedOrder.add_product(product, {quantity:  quantity, merge: false});
```
Fkawala, line 33: We could have a product with a price zero, shouldn't we raise a warning in such a case?

Fkawala (approved):
LGTM (see minor comment). I Did a test on runbot but wasn't able to forge a product and barcode that worked (most likely I'm missing something obvious).

pos_price_to_weight/static/src/js/models.js
ivantodorovich: In that case quantity will be 0, so I think it's what you'd expect

legalsylvain: Indeed, this module is usefull with a system that generate barcodes, generaly with a scale and a printer. Null weight does'nt seems to be a valid use case and should be handled upwind in any case.

Fkawala: Works for me! 
Thank you for the clarification and migration work.

legalsylvain:
> LGTM (see minor comment). I Did a test on runbot but wasn't able to forge a product and barcode that worked (most likely I'm missing something obvious).

Well you can take the demo product which barcode is ``0212345000007`` with pattern "``02.....{NNNDD}``. So : 
``02`` is prefix
``12345`` is the small product id
Then come the price {NNNDD}for exemple 3,00€ is 00300
Then generate the checksum (here 8), with [such website](https://barcode-coder.com/en/ean-13-specification-102.html) 
so the barcode of 3€ of apple is ``0212345003008``

In PoS, the when scanning this product, Odoo will recover the price (3,00€), get the Unit price (1,5€ / kg) and so add 2 kg in a new pos order line.

Is it more clear for you ? Could you improve the description to make it more understandable ? 

kind regards.

Fkawala:
> > LGTM (see minor comment). I Did a test on runbot but wasn't able to forge a product and barcode that worked (most likely I'm missing something obvious).
> 
> Well you can take the demo product which barcode is `0212345000007` with pattern "`02.....{NNNDD}`. So :
> `02` is prefix
> `12345` is the small product id
> Then come the price {NNNDD}for exemple 3,00€ is 00300
> Then generate the checksum (here 8), with [such website](https://barcode-coder.com/en/ean-13-specification-102.html)
> so the barcode of 3€ of apple is `0212345003008`
> 
> In PoS, the when scanning this product, Odoo will recover the price (3,00€), get the Unit price (1,5€ / kg) and so add 2 kg in a new pos order line.
> 
> Is it more clear for you ? Could you improve the description to make it more understandable ?
> 
> kind regards.

Hi Sylvain,

Thank you for the detailed explanation. No need to change the description, this is crystal clear already.

My problem to get a working example of the barcode was that I failed to find the right product ID. The only additional thing we could do here would be to add a product with ID 12345 in the demo data. Doing so would allow one to test the example barcode provided in the documentation. I don't know if it is possible to load a demo product with a preset product ID. 

Cheers,
François

legalsylvain:
No, well the product id i said is not the odoo one. It's another code.

PR #467, opened 2020-03-12
[IMP] Set lot/serial number when returning orders

ivantodorovich:
When returning an order, it'll copy the lot/serial number of the original order lines and set it by default on the refund order

### DIFF
--- a/pos_order_mgmt/static/src/js/widgets.js
+++ b/pos_order_mgmt/static/src/js/widgets.js
@@ -12,7 +12,7 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
     var screens = require('point_of_sale.screens');
     var gui = require('point_of_sale.gui');
     var chrome = require('point_of_sale.chrome');
-    var pos = require('point_of_sale.models');
+    var models = require('point_of_sale.models');
 
     var QWeb = core.qweb;
     var ScreenWidget = screens.ScreenWidget;
@@ -210,12 +210,24 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
             order.trigger('change');
             this.pos.get('orders').add(order);
             this.pos.set('selectedOrder', order);
+            order.get_orderlines().forEach(function (orderline) {
+                if (orderline.pack_lot_lines)
+                {
+                    _.each(orderline.return_pack_lot_names, function(lot_name) {
+                        orderline.pack_lot_lines.add(new models.Packlotline(
+                            {'lot_name': lot_name}, {'order_line': orderline}
+                        ));
+                    })
+                    order.save_to_db();
+                    orderline.trigger('change', orderline);
+                }
+            });
             return order;
         },
 
         _prepare_order_from_order_data: function (order_data, action) {
             var self = this;
-            var order = new pos.Order({}, {
+            var order = new models.Order({}, {
                 pos: this.pos,
             });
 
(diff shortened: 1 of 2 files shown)


pos_order_mgmt/static/src/js/models.js
```diff
@@ -27,4 +27,5 @@ odoo.define('pos_order_mgmt.models', function (require) {
             return res;
         },
     });
+
```
pedrobaeza: Why adding this meaningless empty line?

ivantodorovich: That's a question I don't have an answer for :+1: 
Removed it!

flotho (approved):
LGTM

Thanks for the job

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -210,12 +210,24 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
             order.trigger('change');
             this.pos.get('orders').add(order);
             this.pos.set('selectedOrder', order);
+            order.get_orderlines().forEach(function (orderline) {
```
chienandalu: I feel like this isn't the right method to do it. Why not to put it in `_prepare_order_from_order_data` or in `_prepare_orderlines_from_order_data`?

pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -210,12 +210,24 @@ odoo.define('pos_order_mgmt.widgets', function (require) {
...
+                    orderline.trigger('change', orderline);
+                }
+            });
             return order;
         },
 
         _prepare_order_from_order_data: function (order_data, action) {
             var self = this;
-            var order = new pos.Order({}, {
+            var order = new models.Order({}, {
```
chienandalu, line 218: Not a big deal, but this kind of pointless changes misslead the target of the PR, don't you think :)

pos_order_mgmt/static/src/js/widgets.js
ivantodorovich: Because in that case it would affect other actions like `copy`.
IMO we only need this in this specific action. If you duplicate an order you expect to fill new lots.

pos_order_mgmt/static/src/js/widgets.js
ivantodorovich: Well.. I needed to reference `models.Packlotline` and this way it's much more readable than `pos.Packlotline`. Specially because `pos` is used everywhere.
IMO, it should've always been `models`, so I don't consider it pointless

pos_order_mgmt/static/src/js/widgets.js
ivantodorovich: @chienandalu I moved it here: https://github.com/OCA/pos/pull/467/commits/ff281e4811f614b8a41c158fd0edc7ff789a255e

I agree it's better here ^^

pos_order_mgmt/static/src/js/widgets.js
legalsylvain: In favor to change. pos is not the good name, and is error prone.

legalsylvain (approved):
Lgtm thanks !

/ocabot merge minor

ivantodorovich:
Thanks! :)

ivantodorovich:
@OCA-git-bot are you alive? :thinking: :laughing:

PR #468, opened 2020-03-12
[ADD] pos_invoice_send_mail

ivantodorovich:
Allows to send the invoice by email, from the POS.

invoice-mail-config

invoice-mail-screen


pos_invoice_send_mail/models/pos_config.py
```diff
@@ -0,0 +1,21 @@
...
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    iface_invoice_mail = fields.Boolean('Send Invoice by Mail')
+
+    invoice_mail_template_id = fields.Many2one(
+        'mail.template',
+        string='Invoice Email Template',
+        domain=[('model', '=', 'account.invoice')],
+        context=lambda self: {
```
legalsylvain, line 17: Why this line. Is not default value enough ?

pos_invoice_send_mail/models/pos_order.py
```diff
@@ -0,0 +1,31 @@
...
+        order_ids = super().create_from_ui(orders)
+        refs_to_send_mail = [
+            o['data']['name'] for o in orders
+            if o['data'].get('to_send_mail')
+        ]
+        # Identify orders to send email
+        for order in self.browse(order_ids):
+            if (
+                order.config_id.iface_invoice_mail
+                and order.config_id.invoice_mail_template_id
```
legalsylvain: You could add a constraint (if iface is checked, template is required) and so remove this check.

pos_invoice_send_mail/static/src/js/screens.js
```diff
@@ -0,0 +1,60 @@
...
+    click_send_mail: function() {
+        var order = this.pos.get_order();
+        order.set_to_send_mail(!order.is_to_send_mail());
+        if (order.is_to_send_mail()) {
+            this.$('.js_send_mail').addClass('highlight');
+        } else {
+            this.$('.js_send_mail').removeClass('highlight');
+        }
+        if (order.is_to_send_mail() && !order.is_to_invoice()) {
+            this.click_invoice();
```
legalsylvain, line 28: Question : it's not better to hide the button (send by email) if invoice button is not checked ?

legalsylvain (commented):
Some minor remarks and questions. LGTM otherwise.
Thanks for this great module !

flotho (approved):
Great job,

thanks

pos_invoice_send_mail/models/pos_config.py
ivantodorovich: It's only for usability.
If the user tries to create a new email template using "Create and edit..", it'll have the correct model set by default.

pos_invoice_send_mail/static/src/js/screens.js
ivantodorovich: Hmm, I thought about it but it's much faster for the cashier to do it like this

pos_invoice_send_mail/models/pos_order.py
ivantodorovich: Implemented here: https://github.com/OCA/pos/pull/468/commits/f1a372a95662d3f659c54e469bb6839e65c32e32

legalsylvain (approved):

PR #471, opened 2020-03-20
[MIG] 12.0 mig hw telium payment terminal 

fmdl:
The 8.0 version doesn't work with PosBOX 19.07.

@alexis-via


fmdl:
cc @sla-subteno-it

fmdl:
cc 
@azmeuk
@vco-subteno-it

hw_telium_payment_terminal/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- encoding: utf-8 -*-
```
azmeuk: Odoo 12 requires python 3.5 so the `encoding: utf-8` header is not needed.

azmeuk (approved):
LGTM.

alexis-via (approved):

PR #473, opened 2020-03-25
[IMP] Add margins in res.config

ivantodorovich:
Small UX improvement, for consistency with other fields


legalsylvain:
thanks. Fast tracking trivial patch.

/ocabot merge

PR #474, opened 2020-03-25
[12.0][add] pos_jsprintmanager - Print without an IoT Box

JordiBForgeFlow:
This module allows to print Tickets from the Point of Sale to the computer's
normal or ESC/POS printer without the need to use an IoT Box.

It is based on the JS Print Manager https://www.neodynamic.com/products/printing/js-print-manager/

***Attention***
JS Print Manager is a commercial software. It offers a Trial version which is
fully-functional and does not expire. The Trial version will display a pop-up
window when the receipt is printed. When you purchase the license this pop-up
is not going to appear.

Configuration
==========
In order to use JS Print Manager you must first download and install the client app from
https://www.neodynamic.com/downloads/jspm/

Once you install this module in Odoo:

* Go to *Point of Sale / Configuration / Point of Sale* and flag the checkbox '​Use Jsprintmanager'.

* Complete the field 'Default Printer for Receipts" with the name of the actual
  printer, as it appears in your computer. If you don't set this field
  the application will use the computer's default printer.

* Complete the field 'Output Format' (Normal os ESC/POS) according to the printer that you
  are going to use. The ESC/POS is the special format used by thermal receipt
  printers. See https://reference.epson-biz.com/modules/ref_escpos/index.php?content_id=2

If you have purchased the license:

* Go to *Settings* and activate the developer mode.

* Got o *Settings / Technical / Parameters / System Parameters* and create the
  following parameters:

Key: jsprintmanager.license_owner
Value: as provided by the software provider.

Key: jsprintmanager.license_key
Value: as provided by the software provider.

The client app will then check the validity of your license by connecting to
<your odoo url>/jspm. If you have set the above parameters the page will
return information to allow the app to check the validity.

Usage
=====
#. Open a new PoS session.
#. Make an order and validate it.
#. The receipt should be inmediately printed on your computer's printer.


pedrobaeza:
I don't really understand the need of this... We are printing directly using the browser option of bypassing the print dialog.

JordiBForgeFlow:
@pedrobaeza this module also prints to ESC/POS format, which is something you can't achieve unless you have an IoT Box

pedrobaeza:
OK, I don't know about that part, but you can still do it without IoTBox (with hw_* modules) AFAIK. Any way, this can be a more direct solution, but the commercial part is not very attractive and having an OCA module "promoting" it is not the best IMO, but nevermind, it's my opinion.

legalsylvain (commented):
Hi @JordiBForgeFlow. 

I also don't understand the interest of this PR, sorry.

All is functional :
- with Odoo iotBox out of the box
- with custom local installation of odoo + hw_* modules. (from Odoo & OCA/pos)
- alternatively, you can use the light library [pywebdriver](https://github.com/akretion/pywebdriver) that is quite mature, and from the time being, used by Akretion, Acsone, GRAP. (set under OCA umbrella is discussed).

What is the interest to use this module, instead of more standard and community solutions ? 

thanks !

PR #476, opened 2020-04-14
[12.0][ADD] POS Multiple Control

quentinDupont:
Migration from 8.0 of [this module](https://github.com/grap/grap-odoo-incubator/tree/8.0/pos_multiple_control).
Depends on pos_move_reason, PR [here](https://github.com/OCA/pos/pull/384)

### Description
This module extends the functionality of the point of sale by allowing a better control at the closing of the session with **checking multiple journals**.

- Show differences for all statements chosen (see config part)
- Improves the ergonomics of balances
- Allow user to control each statement. (not only the cash statement, by default) and change his starting and ending balance
change_starting_balance

- Quickly solve differences (we can set an absolute limit), thanks to [pos_move_reason](https://github.com/OCA/pos/pull/384)
See this gif where where we set ending balance with difference, solve it and close session :
end_session_balance_automatic_solve

As the verification is more complete, allow the user to reopen a new session, if the first one is in a closed state.
image

Extra checks are done, to prevent user errors:
-  It is not possible to click on the button 'Close Session' if there are some draft orders

#GRAPOCA

@legalsylvain


legalsylvain:
mais non !

quentinDupont:
> mais non !

and it's green for runbot and travis :heavy_check_mark:

PR #485, opened 2020-05-04
[12.0][IMP] pos_price_to_weight: choose price field

mclaeysb:
`pos_price_to_weight` recomputes the price as `price` (= scanned through barcode) / `list_price`. This PR makes the latter configurable for a POS.

For the review, I am unsure of a couple of things:
- First time working with POS config, please review carefully.
- Feel free to test the additions to the demo data. They are required for making the record available in the POS.
- It would be much more efficient if `product.product` `pos_price_to_weight_price` would be a stored field, but since we don't know which field its compute function should depend on, this seems impossible?

I tried to implement the approach in `product_product.py` in `models.js` first, but couldn't make this work. I'm pitching the idea here -  *sorry for the clunky code*. 
I tried to consult the pos config to find out which field needed to be pushed (stored in `pos_price_to_weight_price_field_id`) and then pushing that field with `product_model.fields.push()`. To acchieve this, I was thinking to use this code in `models.js`:
```js
var pos_price_to_weight_price_field_id_but_then_the_actual_field_name
initialize: function (session, attributes) {
        var self = this;
        _super_PosModel.initialize.apply(this, arguments);

        this.ready.then(function () {
            if (self.config.pos_price_to_weight_price_field_id) {
                pos_price_to_weight_price_field_id_but_then_the_actual_field_name = some_function_to_get_actual_field_name(self.config.pos_price_to_weight_price_field_id)
                });
            }
        });
        var product_model = _.find(this.models, function(model){
            return model.model === 'product.product';
        });
        product_model.fields.push('pos_price_to_weight_price_field_id_but_then_the_actual_field_name');
},
scan_product: function(parsed_code) {
    ...
        quantity = price / product[pos_price_to_weight_price_field_id_but_then_the_actual_field_name]
    ...
}
```
I ran into two problems:
1) In the first step `self.config.pos_price_to_weight_price_field_id` only returned a database ID and string name (e.g. 'Total with VAT') not the actual field name (e.g. `total_with_vat`). Is there a function/way to retrieve the latter?
2) I didn't know in JS how to then call the field I pushed on `product` (to replace `product.list_price`) since the field I pushed was defined by the code. Can this be done with `product[pos_price_to_weight_price_field_id_but_then_the_actual_field_name]`?

Would this be a more interesting approach? Other ideas?


mclaeysb:
@legalsylvain Could you take a look at this?

legalsylvain:
Hi @mclaeysb . I make some changes on your PR. see : https://github.com/coopiteasy/pos/pull/33
It works locally. Could you test with your fields ? 

kind regards.

legalsylvain:
Hi @mclaeysb. I don't understand why runbot is red. Could you rebase and squash the commit ?

mclaeysb:
I rebased, let's see how it turns out. There seem to be a lot of linting errors, most of them pre-existing. Thanks so much for the help!

legalsylvain (approved):
thanks !

remytms (approved):
LGTM.

chienandalu (approved):

PR #488, opened 2020-05-13
[12.0][MIG] pos_accented_search

Fkawala:
This is a copy of the v9 code. It has been tested on V12. Readme has been updated.


ivantodorovich:
Hi @Fkawala 

We need to keep the commit history when migrating a module.
You can follow the steps described here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

Fkawala:
> Hi @Fkawala
> 
> We need to keep the commit history when migrating a module.
> You can follow the steps described here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

Hey @ivantodorovich thanks for the heads up, this is fixed now.

legalsylvain:
Cc : @quentinDupont

pos_accented_search/__manifest__.py
```diff
@@ -1,9 +1,7 @@
-# -*- coding: utf-8 -*-
-# @author: François Kawala
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': "Point of Sale - Accented Product Search",
-    'version': '9.0.0.0.2',
+    'version': '12.0.0.0.2',
```
ivantodorovich:
```suggestion
    'version': '12.0.1.0.0',
```

ivantodorovich (approved):
Minor comment, otherwise LGTM :+1:

pos_accented_search/static/src/js/db.js
```diff
@@ -0,0 +1,25 @@
+odoo.define('pos_accented_search', function (require) {
+
+    "use strict";
+    var db = require("point_of_sale.DB");
+    db.include({
+
+        normalize_characters: function (product) {
+            return product.normalize("NFD")
+                .replace(/[\u0300-\u036f]/g, "")
```
legalsylvain, line 13: maybe you could add a little comment to explain what are there codes ?

legalsylvain (changes requested):
Hi. thanks for this module ! 

Tested on runbot. 

I created two products : 

image

if I write the following text, here is the result : 
- ``café`` : two matches : OK
- ``cafe`` : two matches : OK
- ``càfe`` : two matches : OK
- ``câfe`` : one match : KO
- ``çafé`` : two matches : OK

Do you think you could fix that ? (if not, could you add it in the roadmap / know issues section)

kind regards.

ivantodorovich:
It looks like odoo implemented this in core in `13.0`. So this will probably the last version of this module.

Maybe we can backport this function, seems to cover way more symbols and it's probably fixing @legalsylvain issue

https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/core/utils.js#L669

it has a very extended diacritics map: https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/core/utils.js#L15

https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/static/src/js/db.js#L183

legalsylvain:
Sorry, I did a mistake, when I tested. it seems that when you tip ``^`` then ``a`` the search function is not raised at the second keypress. I don't know if it's an error introduced by this module, or if it's a native bug. so ``câfe`` is OK.

pos_accented_search/static/src/js/db.js
Fkawala: Done! see the latest commit.

Fkawala:
> It looks like odoo implemented this in core in `13.0`. So this will probably the last version of this module.
> 
> Maybe we can backport this function, seems to cover way more symbols and it's probably fixing @legalsylvain issue
> 
> https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/core/utils.js#L669
> 
> it has a very extended diacritics map: https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/core/utils.js#L15
> 
> https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/static/src/js/db.js#L183

I'm not sure I want to backport V13 code, the normalization + diacritics solution covers most (if not all) cases I see in our day to day work at le nid.

pos_accented_search/__manifest__.py
Fkawala: Done thx!

ivantodorovich (approved):

remytms (commented):
LGTM

legalsylvain:
@remytms is it an approval ? if yes, could you do it via the github UI ?

thanks !

remytms:
@legalsylvain How sorry. Seams that I clicked to fast.

remytms (approved):
It looks good to me. :)

PR #332, opened 2019-03-12
12.0 pos_product_qty_available

Ommo73:

PR #490, opened 2020-05-13
[12.0][ADD] pos_require_product_quantity

mclaeysb:
A popup is raised if product quantity is set to 0 for one of the order lines.

Port to v12 of https://github.com/OCA/pos/pull/328 by @robinkeunen


pos_require_product_quantity/models/pos_config.py
```diff
@@ -0,0 +1,10 @@
+from odoo import models, fields
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    require_product_quantity = fields.Boolean(
```
legalsylvain, line 7: If this module is installed, we could imagine that the default behaviour is to forbid empty lines. So i think default= true is more intuitive. Don't you think ?

legalsylvain (approved):
One minor remark. LGTM otherwise.

pos_require_product_quantity/models/pos_config.py
chienandalu: In general my preference is to go with defaults that behave as the standard avoiding possible future weird interactions with other modules tests.

ivantodorovich (approved):

pos_require_product_quantity/models/pos_config.py
legalsylvain: Indeed even if i don't see which test could be broken, enabling that feature. Do you have some ecemple ?

But well, to be sure we can :
Set default = true here
Set the value to false in the demo data.

chienandalu: It's just a rule of thumb so you don't have to worry thinking about wich cases could break in the future :sweat_smile: (for expample, a Tour changed upstream that suddenly breaks the branch CI)

flotho:
Additionnaly adding a control that prevent pos.order to have either positive and negative quantity on stockable products could be a really usefull check.

Regards

chienandalu:
@flotho https://github.com/OCA/pos/pull/332

pos_require_product_quantity/models/pos_config.py
legalsylvain: OK. 
but what about my proposal ?

> Set default = true here
> Set the value to false in the demo data.

chienandalu: Ok, fine :)

mclaeysb: I have bad experience with setting default to true (it is not recognisez and stays false) and setting config values in the demo data (unable to update the module) so I'd prefer to keep this like this. I also agree with @chienandalu.

pos_require_product_quantity/static/src/xml/templates.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <template id="assets_backend" name="pos_require_product_quantity_extension_assets" inherit_id="web.assets_backend">
```
PierrickBrun:
```suggestion
        <template id="assets" name="pos_require_product_quantity_extension_assets" inherit_id="point_of_sale.assets">
```

mclaeysb (commented):
👍

Issue #274, opened 2018-05-26
pos_pricelist - Uncaught TypeError: this.product.get_price is not a function


crsilveira:
Odoo 10.0 
 
Uncaught TypeError: this.product.get_price is not a function
http://127.0.0.1:8069/pos_pricelist/static/src/js/models.js:230
Traceback:
TypeError: this.product.get_price is not a function
    at child.models.Orderline.set_quantity (http://127.0.0.1:8069/pos_pricelist/static/src/js/models.js:230:46)
    at child.initialize (http://127.0.0.1:8069/point_of_sale/static/src/js/models.js:1076:14)
    at child.models.Orderline.initialize (http://127.0.0.1:8069/pos_pricelist/static/src/js/models.js:207:31)
    at child.Backbone.Model (http://127.0.0.1:8069/web/static/lib/backbone/backbone.js:256:21)
    at child [as constructor] (http://127.0.0.1:8069/web/static/lib/backbone/backbone.js:1541:41)
    at new child (http://127.0.0.1:8069/web/static/lib/backbone/backbone.js:1541:41)
    at child.add_product (http://127.0.0.1:8069/point_of_sale/static/src/js/models.js:1987:20)
    at Class.click_product (http://127.0.0.1:8069/point_of_sale/static/src/js/screens.js:1236:33)
    at Object.click_product_action (http://127.0.0.1:8069/point_of_sale/static/src/js/screens.js:1207:59)
    at HTMLSpanElement.click_product_handler (http://127.0.0.1:8069/point_of_sale/static/src/js/screens.js:1078:21)

models.js:230 Uncaught TypeError: this.product.get_price is not a function
    at child.models.Orderline.set_quantity (models.js:230)
    at child.initialize (models.js:1076)
    at child.models.Orderline.initialize (models.js:207)
    at child.Backbone.Model (backbone.js:256)
    at child [as constructor] (backbone.js:1541)
    at new child (backbone.js:1541)
    at child.add_product (models.js:1987)
    at Class.click_product (screens.js:1236)
    at Object.click_product_action (screens.js:1207)
    at HTMLSpanElement.click_product_handler (screens.js:1078)

pedrobaeza:
Please try to reproduce the problem in runbot (https://runbot.odoo-community.org/runbot/repo/github-com-oca-pos-184), once the 10.0 build is done

rodrig92:
In my case, this error is because I have installed pos_cache module and it seems that they are incompatible...
[...]

PR #492, opened 2020-05-18
[10.0][FIX] - handle the case where product is a dict

sbejaoui:
when `pos_cache` is installed, the product is read from the cache and became a `dict` instance instead of `OdooClass` object instance.

This PR, assures that `get_price` is always processed with an `OdooClass` object.

fixes: #274

### DIFF
--- a/pos_pricelist/static/src/js/models.js
+++ b/pos_pricelist/static/src/js/models.js
@@ -206,12 +206,13 @@ odoo.define("pos_pricelist.models", function (require) {
     models.Orderline.prototype.initialize = function (attr, options) {
         _Orderline_initialize.apply(this, arguments);
         if (options.product) {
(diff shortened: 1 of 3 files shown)


sbejaoui:
cc/ @Cedric-Pigeon

Cedric-Pigeon:
@sbejaoui I just tested. Now the session loads without any error. But, when I try to add a product in the basket, it fails:
image

sbejaoui:
> @sbejaoui I just tested. Now the session loads without any error. But, when I try to add a product in the basket, it fails:

My bad, I didn't make a full test, now it's fixed.

Cedric-Pigeon (approved):

legalsylvain:
/ocabot merge patch
thanks.

PR #500, opened 2020-06-03
12.0 fix pos order return

ajayjaviya:
Issue 1:Pos order partial return gives wrong total
When create a partial pos order return amount total on return order is always copy from the original order

pos_order_wrong_total

Issue 2: pos order return module is not working with `l10n_fr_pos_cert`

If `l10n_fr_pos_cert`  is installed user can not return pos order if order is Invoiced or Posted AND the company's country code falls in any of ['FR', 'MF', 'MQ', 'NC', 'PF', 'RE', 'GF', 'GP', 'TF'].
pos_order_return

### DIFF
--- a/pos_order_return/models/pos_order.py
+++ b/pos_order_return/models/pos_order.py
@@ -93,6 +93,8 @@ class PosOrder(models.Model):
                     'returned_line_id': wizard_line.pos_order_line_id.id,
                     'qty': qty,
                 })
+                copy_line._onchange_amount_line_all()
+        new_order._onchange_amount_all()
         return res
 
     def action_pos_order_paid(self):
--- a/pos_order_return/models/pos_order.py
+++ b/pos_order_return/models/pos_order.py
@@ -72,12 +72,13 @@ class PosOrder(models.Model):
         for line in self.lines:
             qty = - line.max_returnable_qty([])
             if qty != 0:
-                copy_line = line.copy()
-                copy_line.write({
-                    'order_id': new_order.id,
-                    'returned_line_id': line.id,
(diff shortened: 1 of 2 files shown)


ajayjaviya:
@ivantodorovich : can you please review this PR

legalsylvain (approved):
Hi @ajayjaviya. Thanks for this patch. 
regarding the first point (wrong total). LGTM
regarding the second point, I don't understand why your last commit is fixing something. (not a blocking point, just to understand).

thanks !

ivantodorovich (approved):

ivantodorovich:
@legalsylvain to answer on point 2:

When calling `copy()` without setting a new value for `order_id`, it triggers this constraint:
https://github.com/odoo/odoo/blob/12.0/addons/l10n_fr_pos_cert/models/pos.py#L195

legalsylvain:
> When calling copy() without setting a new value for order_id, it triggers this constraint:
https://github.com/odoo/odoo/blob/12.0/addons/l10n_fr_pos_cert/models/pos.py#L195

Copying an object make a write on the copied object ? I don't understand why.

OK. then. Thanks for the answer !

ivantodorovich:
Well.. now that you mention this, no.
In fact what's happening is that the following write triggers the constraint with the previous order_id value, because super() is called at the very end https://github.com/odoo/odoo/blob/12.0/addons/l10n_fr_pos_cert/models/pos.py#L196

ivantodorovich:
@legalsylvain can we merge this one? :)

legalsylvain:
Sure !
/ocabot merge patch

PR #503, opened 2020-06-10
[12.0][MIG] pos_lot_selection

KolushovAlexandr:
Migration of pos_lot_selection from v11 to v12.


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0. If the jump is between several versions, you have to modify the source branch to accommodate to this circumstance.

PR #511, opened 2020-06-25
[12.0] ADD pos_multi_ean

eLBati:
In point of sale interface, allow to search products by multiple EAN (defined by product_multi_ean module) while scanning product barcode to add it to order.


pos_multi_ean/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "development_status": "Beta",
+    "category": "Point Of Sale",
+    "website": "https://github.com/OCA/pos",
+    "author": "TAKOBI, Odoo Community Association (OCA)",
+    "maintainers": ["eLBati"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "auto_install": True,
+    "preloadable": True,
```
legalsylvain: why do you define the module as preloadable ?

legalsylvain (approved):
- code review - no test.
- One question. 

LGTM otherwise. 

Thanks !

pos_multi_ean/__manifest__.py
eLBati: :thinking: I copied from the template. It's more clear without it. Removed

ivantodorovich (commented):
LGTM !

Regarding the module name, isn't `pos_product_multi_ean` more convenient, given that this is basically a glue module between `product_multi_ean` and `point_of_sale` ?

eLBati:
@ivantodorovich I followed the same naming as `pos_supplierinfo_barcode`, considering that, in general, the module prefix should indicate the funcional area: prefix  `product` became `pos`.

No problem to change to `pos_product_multi_ean` if you guys think it's better

ivantodorovich (approved):

ivantodorovich:
Makes sense ! Thanks for the clarification

mrcast (approved):
Functional test ok!
Thanks

PR #515, opened 2020-07-02
[ADD] default_partner_pos

jarroyomorales:
Superseeds https://github.com/OCA/pos/pull/438

Now the default_partner is defined in `pos.config` but it is open for discussion.

@bealdav @legalsylvain @ivantodorovich @etobella


legalsylvain (commented):
Hi. thanks for finishing this work. 
regarding this comment, could you add a Roadmap section, mentionning that the code should be moved from Python to Javascript code ? 

https://github.com/OCA/pos/pull/438#pullrequestreview-341698479

otherwise, LGTM.

thanks.

pos_default_partner/views/pos_config.xml
```diff
@@ -0,0 +1,33 @@
...
+        <field name="name">pos.config.form (in pos_default_partner)</field>
+        <field name="model">pos.config</field>
+        <field name="inherit_id" ref="point_of_sale.pos_config_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='inventory_location']" position="after">
+                <h2>Customers</h2>
+                <div class="row mt16 o_settings_container" id="customers_config">
+                    <div class="col-12 col-lg-6 o_setting_box">
+                        <div class="o_setting_right_pane">
+                            <label for="picking_type_id" string="Default Customer"/>
```
ivantodorovich:
```suggestion
                            <label for="default_partner_id"/>
```

pos_default_partner/models/pos_config.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+
+    _inherit = 'pos.config'
+
+    default_partner_id = fields.Many2one(
+        comodel_name="res.partner"
```
ivantodorovich:
```suggestion
        comodel_name="res.partner",
        string="Default Customer",
```

ivantodorovich (approved):
Minor comment, otherwise lgtm :+1:

etobella:
@legalsylvain I think that the code needs to be replicated, not moved. Orders might be created using other systems and we don't want to lose the functionality

bealdav (approved):
OK for me with previous comments of @ivantodorovich

I don't understand why we should add js in the stack while it works as it. It could require more maintenance and less stability.

js implementation'll change heavily in 14.

@legalsylvain I you want the same behavior in js, you could do another implementation in an other module. Just an idea, not mandatory.

But you're free to make all changes you wish here. #justmyopinion

Thanks @jarroyomorales

jarroyomorales:
@ivantodorovich Comments attended! Thanks for the review.

@bealdav @legalsylvain I added it on the roadmap. In my opinion it makes a lot of sense to have it also on the UI even though the pos order will be generated correctly anyways and I think it can go in this module too. I dont know how hard can this be to implement...

@etobella Completely agree with you, we shouldnt remove it from python

flotho (approved):
LGTM

legalsylvain:
> I don't understand why we should add js in the stack while it works as it.

Because a lot of things will not work with the current implementation.
- If the partner has a specific pricelist, the prices will not be computed correctly.
- if the partner has a specific fiscal position, taxes computation will be wrong.
- the name of the partner will not be displayed on the ticket, at the first print. But reprinting the bill with pos_order_mgmt will display the name. (It is attended that the original ticket and the new ticket are identical).
- Regarding the User Experience, it is confusing to create a pos order without customer, and finally having a pos order saved with a customer. 
- etc.

Again, not mandatory for a beta implementation of the module. But writing a Roadmap section avoids forgetting the problems related to the current implementation.

> @bealdav @legalsylvain I added it on the roadmap.

@jarroyomorales thanks !

kind regards.

etobella (approved):
LGTM :+1:

bealdav (approved):
@legalsylvain, from my point of view:
If it's a default partner there is no reason to have specific settings like specific pricelist, specific fiscal position, specific of any kind (then it's better if only 1 default partner by company in this case but the debate is over now)
If this default customer is named `Anonymous` no problem to be printed on ticket or not (including for accounting manager or fiscal administration)
About user experience, again if it's named `anonymous` it's self explained.

Ok for me.

PR #517, opened 2020-07-25
[12.0-MIG] pos order load

cvinh:


legalsylvain (changes requested):
Hi @cvinh. 
thanks for this work ! 

Just tested on runbot.
1) the save draft order works correctly, and create in the backoffice, an draft pos order.
2) the load draft order doesn't work. when clicked, it displays the list of confirmed orders. (conflict with pos_mgmt in my opinion).

kind regards

PR #521, opened 2020-08-04
[12.0][ADD] pos_session_closing_stock_error

ivantodorovich:
Prevent closing PoS Sessions that have stock.picking errors.

These errors usually happen when selecting a wrong serial/lot number in the POS.

session_with_errors

error_msg

Configure
--------

In any Point of Sale configuration (pos.config), disable "Allow closing sessions with stock errors"
to prevent closing sessions with stock errors.


pos_session_closing_stock_error/models/pos_config.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models, fields
+
+
+class PosConfig(models.Model):
+    _inherit = 'pos.config'
+
+    allow_session_closing_with_stock_errors = fields.Boolean(
```
legalsylvain, line 7: Hi. I think that the field should be set to default=False. Otherwise, the module has no effect, once installed. It is attended that it enables the feature, I think.
It is also more secure. What do you think ?

ivantodorovich: > "It is attended that it enables the feature"

Yea.. sometimes I find myself in a dilemma because of this.
But I agree, most modules in this repo enable the feature upon install, so this one should do it too. :+1:

legalsylvain (approved):
Thanks !

flotho (approved):
The error could also come from the fact that people could use postive and negative quantities in the same pos.order and odoo can't deal with 2 picking one for return the other for delivery.
Very useful module.

PR #525, opened 2020-08-26
[FIX] pos_order_mgmt: test pricelist

chienandalu:
In the test, the only available pricelist should be the one we prepared

cc @Tecnativa TT25280


pedrobaeza (approved):
Agh, we missed that!

chienandalu:
Yeah, sorry, that was the crucial part :sweat_smile:

PR #526, opened 2020-08-27
[FIX] pos_order_mgmt test

chienandalu:
If we `copy()` the config the currency error raises before prevent it.
Doing a `copy()` is meaningless anyway.

Sorry about the noise. Now tested for sure before PR :sweat_smile: 

cc @Tecnativa TT25280

### DIFF
--- a/pos_order_mgmt/tests/test_module.py
+++ b/pos_order_mgmt/tests/test_module.py
@@ -34,8 +34,7 @@ class TestModule(TransactionCase):
         })
 
         # Create a new pos config and open it
-        self.pos_config = self.env.ref('point_of_sale.pos_config_main').copy()
-        self.pos_config.write({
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't get the problem copying, but this still depends on existing data, so why not creating a new config from scratch? You can also pass a dictionary to copy for using that values in the copy instead of writing after.

chienandalu:
There's no need to copy the config IMO.

> why not creating a new config from scratch? 

I think the problem there is the amount of demo data that is needed to be created. I try it anyway

pedrobaeza:
Anyway, if you think the copy is valuable, just pass the overriden vals in the copy call.

chienandalu:
Ok. Changed and tested in our CI

pedrobaeza (approved):

PR #491, opened 2020-05-17
[12.0][MIG] pos_tare w/ barcode

Fkawala:
Following up our discussion with @legalsylvain here #436 and there #447 Sylvain and I decided to split the work related to tare in two parts.

- The first part handles manual tare input (as described in #436) and barcode base tare input (as described in #447). 
- The second part, to be released afterwards, handles the tare barcode label printing from the POS UI (as described in #447).

This PR covers the first part. The second part will be delivered later next week.

PR #527, opened 2020-08-27
[12.0][MIG] pos_tare_generate_barcode

Fkawala:
This PR contains the refactorized migration of the tare barcode generator. This work follows on PR #491 as decided by @legalsylvain and myself here #436 and there #447. This new add-on depends on `pos_tare` (PR #491). 

This PR adds a new screen to the POS. The new screen enables to web-print barcode labels of tare values. The tare value is measured using the POS scale. A capture of the new screen below (no change since v9.0 / PR #447).

ready_to_print

cc. @houssine78 @vvrossem


pos_tare/i18n/fr_CH.po
```diff
@@ -0,0 +1,290 @@
+# Translation of Odoo Server.
```
legalsylvain, line 1: Hi @Fkawala. 

regarding this file : 
-> it's seems to be out of the scope of this PR.
-> but overall, it will not work. regarding translation, if you add a new module on github, you can add po & pot files. But once the PR merged, you have to use weblate. https://translation.odoo-community.org/projects/pos-12-0/pos-12-0-pos_tare/ Otherwise, I think your work will be erased.

legalsylvain (changes requested):
Hi @Fkawala. Just tested on runbot.

when I click on the button, and if the setting is not correctly set, you have loop on the error, and cannot go out of the screen.

image

(quick review only.)

Fkawala:
> Hi @Fkawala. Just tested on runbot.
> 
> when I click on the button, and if the setting is not correctly set, you have loop on the error, and cannot go out of the screen.
> 
> image
> 
> (quick review only.)

Hi @legalsylvain,

This is a known problem/bug of the POS debug tool. The fake scale sends weight measures in an unexisting UOM "Kg", whereas the POS weight reference UOM is "kg" in **lowercase**. I'll fix the error loop problem.

Best regards,
François

ps. thank you for your note regarding the translation!

Fkawala:
Hey, @remytms would you have 5 minutes to review that? Sylvain was OK with this PR, aside from the looping error that is now fixed. Cheers, François

houssine78:
@Fkawala you probably wanted to ping @vvrossem on this PR for review.

pos_tare_generate_barcode/static/src/xml/pos_tare_generate_barcode.xml
```diff
@@ -0,0 +1,47 @@
...
+                     <span class='button back'>
+                        <i class='fa fa-angle-double-left'></i>
+                        Back
+                    </span>
+                    <h1>Print a tare label</h1>
+                </div>
+                <div class="centered-content">
+                    <div class="pos-tare-label-container"></div>
+                    <div class="pos-directions-for-user">
+                        Set the pot on the scale and check the weight above
```
houssine78, line 16: I would rather use container instead of pot.

pos_tare_generate_barcode/static/src/xml/pos_tare_generate_barcode.xml
```diff
@@ -0,0 +1,47 @@
...
+            </div>
+        </div>
+    </t>
+
+    <t t-name="PosTareLabel">
+        <div class="pos-center-align">
+            <div class="pos-tare-paper">
+                <div class="pos-tare-label">
+                    <img t-att-src="'/report/barcode/EAN13/' + widget.get_barcode_data()" />
+                    <span class="caption">tare = <t t-esc="widget.get_tare_weight()" />kg</span>
```
houssine78: kg is hard coded.

Potential issue would be that the tare is on g and not kg. Any way to have uom if the tare configured somewhere ?

second point here is that it limits the use of this module on the kg metric system.

houssine78: I see that my proposition could to some uom conversion between the tare of the container and the uom of the contained product (not sure of that).

I agree to go ahead without change if you add a section explaining the limitation in the issue/limitation of the readme

pos_tare_generate_barcode/static/src/xml/pos_tare_generate_barcode.xml
Fkawala: I can change and I do not really care, "container" is more generic whereas "pot" is more specific to food/cooking.

pos_tare_generate_barcode/static/src/xml/pos_tare_generate_barcode.xml
Fkawala: Good point, I'll fix, thank you.

houssine78: ok no issue with conversion, I just read the js code... and saw that you use the uom returned by the scale which is gr for our metric system. only limitation here is it doesn't support UK/US metric system. I'm fine with it !

houssine78 (approved):

pos_tare_generate_barcode/static/src/js/screens.js
```diff
@@ -0,0 +1,200 @@
...
+                    return r.type === "tare";
+                })[0];
+            return rule.pattern;
+        },
+        get_barcode_rules: function () {
+            return this.pos.barcode_reader.barcode_parser.nomenclature.rules;
+        },
+        set_tare_weight: function (scale_measure) {
+            var weight = scale_measure.weight;
+            var measure_unit = this.pos.units.filter(
```
Fkawala, line 86: @houssine78 We pick up the scale UOM from there. A scale using the Imperial scale would not be a problem there as long as the UOM exists.

pos_tare_generate_barcode/static/src/js/screens.js
```diff
@@ -0,0 +1,200 @@
...
+                     "measure (UOM) in not found in the point of sale. You " +
+                     "may need to create a new UOM named %s. The UOM name is "+
+                     "case sensitive."), scale_measure.unit,
+                    scale_measure.unit));
+            }
+
+            if (weight > 0) {
+                var tare_uom = this.pos.config.iface_tare_uom_id[0];
+                var tare_unit = this.pos.units_by_id[tare_uom];
+                this.weight_in_tare_unit = convert_mass(weight,
```
Fkawala, line 115: @houssine78 The tare UOM is a parameter of the POS. Here we convert the weight expressed in the scale's UOM into the parameterized UOM.

pos_tare_generate_barcode/static/src/js/screens.js
```diff
@@ -0,0 +1,200 @@
...
+                this.lock_screen(false);
+            }
+        },
+        get_tare_weight: function () {
+            if (typeof this.weight_in_tare_unit === 'undefined') {
+                return this.default_tare_value;
+            }
+            return this.weight_in_tare_unit;
+        },
+        barcode_data: function (weight) {
```
Fkawala, line 135: @houssine78 Here we might have a problem b/c we only consider weight to be in the international standard. We can not store the UOM in the label. We do not want the old labels to become obsolete if the user changes the UOM configuration. Hence we should stick to international standard mass units but fix the display to be in the configured UOM. The current solution would probably break with the Imperial scale. I'll try to propose a fix.

Fkawala:
@houssine78 this solves the point you raised about UOM used in the barcode label. @vvrossem Could you please have a look? Thank you!

Fkawala:
@legalsylvain Besides fixing the tare barcode UOM problem, I introduced an additional parameter in the `pos_tare` module. This parameter enables users to configure if they want to display the tare input field on the scale screen.

The rationale supporting this parameter is as follows. In our shop, we use Adam scales. Those scales do not feature a reset signal, hence odoo needs to reset the weight after `<x>` seconds to avoid having the operator ordering twice the same weight for two different products. A consequence of odoo resetting the weight after `<x>` seconds is that our cooperators do not have enough time to use the tare input field on the scale screen before reset happens. One could argue that we should increase `<x>` but that'll increase the risk of ordering with the wrong weight.

Therefore, to reduce the cognitive load of the operator using the scale screen, we need to hide the input tare field (until we have better scales)

I hope all that is clear. Let me know if not.

pos_tare/models/pos_config.py
```diff
@@ -4,6 +4,13 @@
 class PosConfig(models.Model):
     _inherit = "pos.config"
 
+    iface_tare_on_scale_screen_method = fields.Boolean(
+        string='Show tare input on scale screen',
+        help="Adds an input text field to set the tare on the scale screen.",
+        default=True,
+        required=True
```
legalsylvain: I don't think a required value is ok for a boolean field.

pos_tare/static/src/js/screens.js
```diff
@@ -18,7 +20,21 @@ odoo.define('pos_tare.screens', function (require) {
                 var order = this.pos.get_order();
                 var selected_order_line = order.get_selected_orderline();
                 var tare_weight = code.value;
-                selected_order_line.set_tare(tare_weight, true);
+                var tare_uom = this.pos.config.iface_tare_uom_id[0];
+                var tare_unit = this.pos.units_by_id[tare_uom];
+                var kilogram_uom = this.pos.units.filter(
+                    function (u) {
+                        return u.name === "kg";
```
legalsylvain, line 27: sorry. I still don't understand why there are hardcoded text regarding uom in this module. there is no way to have something more modular ? if it's not possible, do you think you could write a section in the ROADMAP section, to mention to do it (maybe when odoo PoS will be refactored...)

legalsylvain (commented):
partial review regarding pos_tare module.

pos_tare/models/pos_config.py
Fkawala: Ok, I'll fix then, thanks for the heads up.

pos_tare/static/src/js/screens.js
Fkawala: Assume you do not hard code this UOM (the one that is used to encode the tare barcode labels). Let's say this UOM is configured to be kilogram. Now print a tare label with 1 unit of UOM, that's one kilogram. Assume the user changes the configuration for this UOM and selects "metric ton". Afterwards one scans the old tare barcode label, she'll get a tare of one metric ton instead of the correct weight of the container that is one kilogram. If you have a better solution than hard coding, I'm happy to change that. Hope this is clear enough.

Fkawala: Hope this is clear enough...

Fkawala: @legalsylvain What's your opinion on that? Are we all good?

Fkawala:
Hello, @legalsylvain Could you please approve this PR?

Fkawala:
@legalsylvain do you think we can merge that PR? If there is something blocking could you please let me know what it is? This PR is waiting to be merged for months. Thank you!

pos_tare/static/src/js/screens.js
legalsylvain: Sorry. Back on that point. I was fully busy the last monthes. 

I think _" Assume the user changes the configuration for this UOM and selects "metric ton"_ can not occure.

I tested your Use case with demo data in V12 : 
- select the product with kilogram unit (Cable Management Box) (``stock.product_cable_management_box``)
- change the quantity (70kg in hand).
- select the uom kg (``uom.product_uom_kgm``)
- try to change uom.

You'll have this errors.

image

My opinion is that odoo doesn't handle that use case. (that is quite hard to handle, I think). so, you should not have to handle that use case in a OCA module, if Odoo core module doesn't allow it.
It will make the code simpler.
Dont you think ?

pos_tare_generate_barcode/static/src/js/screens.js
```diff
@@ -0,0 +1,220 @@
...
+                    try {
+                        self.set_tare_weight(weight);
+                    } catch (error) {
+                        var title = _t("Failed to read weight from scale.");
+                        var popup = {title: title, body: error.message};
+                        self.gui.show_popup('error', popup);
+                        self.pos.proxy_queue.clear();
+                    }
+                });
+            }, {duration:150, repeat: true});
```
vvrossem, line 61: > [There  must  be  at  least  a  200-ms  delay  between  commands  to  allow  for  processing  data  response  time  at  the  scale.](https://www.manualslib.com/manual/861274/Mettler-Toledo-Viva.html?page=51#manual)

`150 ms`  might be too low if, on the scale software-side, no `protocol.commandDelay` is used between weight readings

pos_tare_generate_barcode/static/src/js/screens.js
```diff
@@ -0,0 +1,220 @@
...
+            this.$('.back').click(function () {
+                self.click_back();
+            });
+            this.$('.print-label').click(function () {
+                if (!self._locked) {
+                    self.print();
+                }
+            });
+        },
+        render_receipt: function () {
```
vvrossem, line 207: a detail, but if you render `PosTareLabel`, you can name your function `render_label`

pos_tare_generate_barcode/static/src/xml/pos_tare_generate_barcode.xml
```diff
@@ -0,0 +1,47 @@
...
+                </div>
+            </div>
+        </div>
+    </t>
+
+    <t t-name="PosTareLabel">
+        <div class="pos-center-align">
+            <div class="pos-tare-paper">
+                <div class="pos-tare-label">
+                    <img t-att-src="'/report/barcode/EAN13/' + widget.get_barcode_data()" />
```
vvrossem, line 31: TIL you can generate barcode that way :tada:

vvrossem (changes requested):
Thank your work on this module :)
My apologies for the very late review. Here are some remarks.

1. the `duration` of your `proxy_queue` might create issues on the scale software-side

2. In standard POS, Odoo implements `set_weight()` (in `ScaleScreenWidget`) as:
```javascript
    set_weight: function(weight){
        this.weight = weight;
        this.$('.weight').text(this.get_product_weight_string());
        this.$('.computed-price').text(this.get_computed_price_string());
    },
```
where  `get_product_weight_string()` returns a `defaultstr` (e.g. `0.000 Kg`) [in specific cases](https://github.com/OCA/OCB/blob/12.0/addons/point_of_sale/static/src/js/screens.js#L402)

  If not already done, I think `set_tare_weight()` could use the same idea: 
- `weight` should be what the scale returns
- `get_tare_weight_string()` should return a default value in case of missing UoM

PR #528, opened 2020-08-28
[13.0][MIG] pos_timeout

acsonefho:
Migration from `10.0` to `13.0`.

The default timeout set by Odoo is 30 seconds (instead of 7.5 for v`10.0`).


chienandalu (commented):
I personally prefer `pos_picking_delayed` (`stock.picking` is the operation that takes most of the time) than blindly raising order sync timeouts.

ivantodorovich:
> I personally prefer `pos_picking_delayed` (`stock.picking` is the operation that takes most of the time) than blindly raising order sync timeouts.

Yeah, but sometimes it’s not enough, specially with very a customised POS

acsonefho:
> pos_picking_delayed

Sorry but the purpose of this PR is to migrate the module. So I prefer to let original field name.
That avoid migration script from previous version etc.
But feel free to create a new MR with your new field if necessary.

ivantodorovich:
> > pos_picking_delayed
> 
> Sorry but the purpose of this PR is to migrate the module. So I prefer to let original field name.
> That avoid migration script from previous version etc.
> But feel free to create a new MR with your new field if necessary.

https://github.com/OCA/pos/tree/12.0/pos_picking_delayed Is another module that more or less tackles the same problematic

chienandalu:
> > I personally prefer `pos_picking_delayed` (`stock.picking` is the operation that takes most of the time) than blindly raising order sync timeouts.
> 
> Yeah, but sometimes it’s not enough, specially with very a customised POS

Ok, ok :)

ThomasBinsfeld (approved):

ivantodorovich (approved):

PR #529, opened 2020-08-28
[13.0][MIG] pos_restricted_customer_list (from 9.0)

acsonefho:


pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
```diff
@@ -0,0 +1,57 @@
...
+                    // eslint-disable-next-line handle-callback-err
+                    function(err, event) {
+                        event.preventDefault();
+                        def.reject();
+                    }
+                );
+            return def;
+        },
+    });
+});
```
ivantodorovich, line 61: I think this is not ok.

I can't find a reason to overload `load_new_partners`:
https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/static/src/js/models.js#L622

Also I think the purpose of that function is to get **new** partners (after the pos has loaded): 
see: https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/static/src/js/models.js#L617

I do think you need to overload `prepare_new_partners_domain` to add the leaf to filter by `available_in_pos`, but you need to keep the original leafs by calling _super(), the `write_date` leaf is what gets the job done.. we shouldn't remove it https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/static/src/js/models.js#L616


That's to have the `load_new_partner` function work with this module.. But we also need to restrict the partners loaded when the pos is starting, which is actually the main purpose of this module.

To do that I think it's better to add a domain property to the `res.partner` model object: https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/static/src/js/models.js#L200

I haven't tested but I think something like this should work:

```js
var PosModels = require('point_of_sale.models');
var PosModel = PosModels.PosModel;

function model_load_domain(model_name, domain) {
    var models = PosModel.prototype.models;
    for (var i = 0; i < models.length; i++) {
        var model = models[i];
        if (model.model === model_name) {
            if ((model.domain instanceof Array) && model.domain.length > 0) {
                model.domain = model.fields.concat(domain || []);
            } else {
                model.domain = domain;
            }
        }
    }
}

model_load_domain("res.partner", [['available_in_pos', '=', true]]);
```

pos_restricted_customer_list/static/src/js/pos_restricted_customer_list.js
```diff
@@ -0,0 +1,57 @@
+odoo.define("pos_restricted_customer_list.point_of_sale.models", function(require) {
```
ivantodorovich, line 1:
```suggestion
odoo.define("pos_restricted_customer_list.models", function(require) {
```

Also this file should be renamed to `models.js`, for convenience.

it's also good to return the PosModels object at the end. I know it's unlikely some other module will extend it, but it's nice ;)

pos_restricted_customer_list/views/assets_backend.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2017-2019 Therp BV <https://therp.nl>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template
+        id="assets_backend"
+        name="pos_restricted_customer_list assets"
+        inherit_id="web.assets_backend"
+    >
```
ivantodorovich, line 9:
```suggestion
    <template id="assets" inherit_id="point_of_sale.assets">
```

ivantodorovich (changes requested):

ThomasBinsfeld (approved):

chrisdec (approved):

PR #531, opened 2020-09-02
[FIX][12.0] pos_cash_move_reason : date of statement line depends now on the context timezone

### LINKED ISSUE #507: cash in/out journal entry uses odoo server time and not user settings time
Hello community,

I was testing the pos_cash_move_reason module in odoo 12 and works fine for the exception that the movements created in the journals takes place using the odoo server time and not the user settings time.
This creates confusion with the end user and also leads to problems when creating reports because the point of sale closing balance and sales are all created using the user settings time.
This was tested in runbot (please see attached image). I set the timezome of the test user to Argentina (UTC-3) and the date the test took place is the 06-16-2020.
The sales are registered correctly in the date 06-16-2020 (which is correct by using the user timezone) but the cash-out movement was registered one day after because of the difference between server time and user time.
pos_cash_move_reason wrong dateTime

Thank you!
Emanuel

legalsylvain:
Closes #507

Trivial patch. for the time being, the bank statement line generated by this module has a bad date. (doesn't take into account timezone). 
Applying same code as per odoo pos order payment. ref : https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/models/pos_order.py#L948

CC : @emavarela, @ecogit, @chienandalu

thanks.

### DIFF
--- a/pos_cash_move_reason/wizard/wizard_pos_move_reason.py
+++ b/pos_cash_move_reason/wizard/wizard_pos_move_reason.py
@@ -90,7 +90,7 @@ class WizardPosMoveReason(models.TransientModel):
             account = self.move_reason_id.expense_account_id
             amount = - self.amount
         return {
-            'date': fields.Datetime.now(),
+            'date': fields.Date.context_today(self),
             'statement_id': self.statement_id.id,
             'journal_id': self.journal_id.id,
             'amount': amount,


chienandalu (approved):

ivantodorovich (approved):

legalsylvain:
/ocabot merge patch

Thanks for your review.

PR #535, opened 2020-09-09
[12.0] [IMP] pos_picking_load: returning widgets so they can be inherited

Du-ma:
All new widgets should be returned.
Because right now there isn't really a way to inherit the widget and add some functionality to it.

### DIFF
--- a/pos_picking_load/static/src/js/widget.js
+++ b/pos_picking_load/static/src/js/widget.js
@@ -317,4 +317,9 @@ odoo.define('pos_picking_load.widget', function (require) {
         },
     });
 
+    return {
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Hi @Du-ma. 

thanks for your contribution. LGTM.
Is it for custom needs ? or do you will to developp an extra OCA module ? Curious to know what are your needs.

Fast tracking trivial patch.

/ocabot merge patch

Du-ma:
Hey @legalsylvain , cheers for the fast reaction.

I'm making a custom need module for a client. And some of the needs include spicing the javascript part up a bit :grin:

ivantodorovich (approved):

PR #536, opened 2020-09-09
[ADD][12.0] pos_hide_banknote_button

legalsylvain:
Hi all,

Adding a trivial module that hides PoS payment button "+10", "+20", "+50", that are quite useless and can be error prone.

## Before

before

## After 

after


legalsylvain:
Note : i didn't make the "Backspace" button bigger, because it was requiring overwriting XML structure. Prefered light CSS changes to avoid incompatibility with other modules.

ivantodorovich (approved):
:+1: 

Not blocking, but what about "disabling" the buttons instead, in a way that doesn't mess with the visual layout of the pad?

SandieFavre (approved):
LGTM (Functional test / no code review)

legalsylvain:
> Not blocking, but what about "disabling" the buttons instead, in a way that doesn't mess with the visual layout of the pad?

Other approach indeed. the reality is that my users asked me the purpose of there buttons. (that is quite useless indeed). 
- there are not all the bill (not +5 for exemple, in €)
- there are not button for pieces.

If we disable the buttons, my users still continue to ask me the objective of there buttons -;)

pos_hide_banknote_button/README.rst
```diff
@@ -0,0 +1,70 @@
...
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-legalsylvain%2Fpos-lightgray.png?logo=github
+    :target: https://github.com/legalsylvain/pos/tree/12.0-pos_disable_change_cashier/pos_disable_change_cashier
+    :alt: legalsylvain/pos
+
+|badge1| |badge2| |badge3| 
+
+This module extends Odoo Point Of Sale features, allowing to disable
+the button that allow to switch cashiers.
```
chienandalu: Mmm. Can you regenerate the README :)

legalsylvain: removed. let @OCA-git-bot generates a correct one. Thanks.

PR #538, opened 2020-09-15
[IMP] the font size of the displayed product name depends on the length of the product name

legalsylvain:
Hi all,

proposing a trivial patch that is changing the font size, depending on the length of the product name. It so improve the visibility of the products that have a large name.

## Before

image

## after

image


pedrobaeza:
Usability indicates that a fixed font size (larger or smaller) and a fixed position are advisable for better recognition, so I don't see this change as good (neither the previous situation), but it's only my opinion and I'm not using this module at all.

pos_default_empty_image/static/src/xml/pos_default_empty_image.xml
```diff
@@ -14,7 +22,7 @@
                     </span>
                 </t>
             </div>
-            <div class="product-name-without-image" t-attf-id="article_product_#{product.id}">
+            <div class="product-name-without-image" t-attf-id="article_product_#{product.id}" t-attf-style="font-size: #{font_size}px !important;">
```
gaikaz: To avoid large spacings between lines, I would suggest experimenting with `line-height`, something like so:
```suggestion
            <div class="product-name-without-image" t-attf-id="article_product_#{product.id}" t-attf-style="font-size: #{font_size}px !important; line-height: #{font_size}px !important;">
```
I didn't check this, just wrote this suggestion.

gaikaz (commented):
I would agree with @pedrobaeza that this is kind of bad practice in UX terms, but in case of what this module is for (replacement of "no image" image), any end-user would be much more happy to see more of the product name instead of first few words.
This would be very useful to have when there are many similarly named variants of products.

legalsylvain:
Thanks pedro ! Didn't thought about that point. I have to think about it !
Regards.

legalsylvain:
@gaikaz : I applied your suggestion. thanks ! It works better.

image

@pedrobaeza :Here is the situation : 
- In the default PoS, font-size if very small (11 px) (regarding back-office tree font size, 13px) and in a lot of implementation there are no image in the product. So, 90% of the space of the square of the product is useless. -> not a good UX.
- So removing the image placeholder let a big blank space and very small text -> not a good UX.
- Adaptating size to length of the product is a good solution to optimize the space, and avoid to annoy users with very little text. 

In the other hand, you're right : fixed font size seems to be better. 

The real solution could be to have a tree view in the PoS, instead of the kanban view. (I personally find better products in tree than in kanban, because in tree view there is only one dimension). I don't have the time to provide such customization, and I think that all the PoS will change with owl, so I prefer to wait. 

Regarding your comment, I introduced a ``iface_fixed_font_size`` configuration in the ``pos.config``. if not set, the font-size will depends on the length of the name. Otherwise, it will be fixed.

pos_config_form


11 px
image

14 px
image

What do you think ?

pos_default_empty_image/static/src/xml/pos_default_empty_image.xml
legalsylvain: done. thanks.

gaikaz (approved):

pedrobaeza:
Good balance found, @legalsylvain 

About the tree view, we are going somehow in the opposite direction, creating a sort of kanban view in sales orders (PR coming to OCA/web soon), but counting with most of the products with image, and thinking in mobile screens (which is the target with other revolutionary changes that we will have soon - a hint with the module name: `web_pwa_cache` :wink: ).

WIP gif:

legalsylvain:
wow ! A little out of the scope of the current topic, but looks very nice ! please ping me when it's ready to review.

regarding tree vs kanban view, I find the first screenshot is better to search a product than a second one, specially when there is no image.

image

image

don't you think ? don't know if there are analysis regarding that point, as there are regarding lower / upper case.

(ref : https://www.sciencedirect.com/science/article/pii/S0042698907002830)

pedrobaeza:
Sorry for the off-topic, but it was something that want to share with you related somehow with the kanban view thing.

About the Z scans for the result, I don't know literature right now, but Odoo proved paradigms are:

- Iterative filtering for getting few results to pick up from instead of long scans.
- Visual scan instead of text scan. Maybe you can use the now so-well technique in mails of putting initials as avatar of a recipient.
- Kanban priority over list, as it has several advantages:

  - Better eyes ergonomy, having the information condensed in an area instead of spread over full monitor width.
  - More information, as the density is higher, not constrained to maximum column width.
  - Richer information, being able to use formats, images, etc. This is being compensated somehow in v14 giving richer format also to lists (but still constrained by row height).

pedrobaeza:
Some readings about Z scanning (not directly related, but interesting about design patterns):

https://uxplanet.org/z-shaped-pattern-for-reading-web-content-ce1135f92f1c
https://medium.com/@lineindesign/be-a-designer-who-can-also-help-with-writing-copy-2f4ea02a5646

SandieFavre (commented):
LGTM (Functional test / no code review)
Perhaps the best solution would have been to limit the number of characters in the name of the product, but in fact, when users will be confronted with entitled that are too long and therefore written in small print, they will probably think of rewriting it.

quentinDupont (approved):

PR #540, opened 2020-09-25
[FIX] pos_customer_display: compatibility with pos_restaurant

legalsylvain:
similar to #539 

don't crash if ``get_order()`` is null, that occures if ``pos_restaurant`` is enabled.


hparfr (approved):

legalsylvain:
thanks for your review.

fast tracking trivial patch.

/ocabot merge patch

PR #541, opened 2020-09-28
[12.0] ADD pos_product_multi_price: Make pricelists based on product_multi_price work in point of sale

eLBati:
See ``product_multi_price``'s documentation about how to configure and use it.

### DIFF
--- a/pos_product_multi_price/models/product.py
+++ b/pos_product_multi_price/models/product.py
@@ -1,0 +1,24 @@
+import json
+from odoo import models, fields, api
+
+
+class Product(models.Model):
+    _inherit = 'product.product'
+
+    # technical field used in POS frontend
+    price_ids_json = fields.Char(
+        "Multi price data dict", readonly=True,
+        compute="_compute_price_ids_json")
+
+    @api.multi
+    def _compute_price_ids_json(self):
+        for p in self:
+            res = []
+            for price in p.price_ids:
+                res.append({
+                    'id': price.id,
+                    'price_name': price.name.name or '',
+                    'price_id': price.name.id or '',
+                    'price': price.price,
+                })
+            p.price_ids_json = json.dumps(res)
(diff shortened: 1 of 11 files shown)


mrcast (approved):
Functional test ok!

eLBati:
@OCA/pos-maintainers what do you think?

eLBati:
Conflict fixed

pos_product_multi_price/static/src/js/models.js
```diff
@@ -0,0 +1,64 @@
...
+                        }
+                        if (rule.price_max_margin) {
+                            price = Math.min(price, price_limit + rule.price_max_margin);
+                        }
+                        return true;
+                    }
+                });
+            }
+            return false
+        });
```
qgroulard, line 59: Here you ignore possible high-priority rules not based on multi_price.
Price set by those high-priority rules will be overridden by the multi_price.

eLBati: Doesn't the same happen with `product_multi_price`?

qgroulard: No it doesn't.
If you have a rule like "50% of standard price on product variant A" and a second rule "Use multi_price X for all products", when you create a sale order for product A, the price will be 50% of the standard price.
The second rule won't override the first one.

pos_product_multi_price/models/product.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo import models, fields, api
+
+
+class Product(models.Model):
+    _inherit = 'product.product'
+
+    # technical field used in POS frontend
+    price_ids_json = fields.Char(
+        "Multi price data dict", readonly=True,
+        compute="_compute_price_ids_json")
```
ivantodorovich, line 11:
```suggestion
    price_ids_json = fields.Char(
        "Multi price data dict",
        help="Technical field used in POS frontend",
        compute="_compute_price_ids_json",
    )
```

- computed fields are already readonly by default.
- you can move the comment to the help attribute, it's more helpful this way

ivantodorovich (changes requested):
also check @qgroulard 's comment :)

Joaco1980 (approved):
Works correctly also on V13

eLBati:
Hi @legalsylvain any chance for this PR?

PR #544, opened 2020-10-06
[IMP] pos_payment_terminal : improve display of the loading button

legalsylvain:
improve display of the loading button in the pos_payment_terminal module.

context : 
when communicating with the terminal, for the time being, the validate button changes.
so, if the communication fails, the button never goes back to the original display. So adding a new button.

(also improve code, before inserting html by javascript is not a good practice). 

## Before

image

## After

image

CC : @alexis-via, @ivantodorovich, @SandieFavre, @quentinDupont #GRAPOCA

Thanks.

### DIFF
--- a/pos_payment_terminal/static/src/js/screens.js
+++ b/pos_payment_terminal/static/src/js/screens.js
@@ -32,9 +32,9 @@ odoo.define('pos_payment_terminal.screens', function (require) {
             if (!order) {
                 return;
             } else if (order.in_transaction) {
-                self.$('.next').html('<img src="/web/static/src/img/spin.png" style="animation: fa-spin 1s infinite steps(12);width: 20px;height: auto;vertical-align: middle;">');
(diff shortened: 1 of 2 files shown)


alexis-via (approved):

alexis-via:
Great initiative to improve this module. I'm not good enough in JS to understand the details, but it's great to improve this stuff. Another thing that would be great : make the button "Start transaction" bigger. The current button is small and therefore it's not easy to click on it on a tactile screen.

ivantodorovich (approved):
I'm not using this module in 12.0 yet (it's in my todo list though)

I agree with @alexis-via, it'd be good to increase the size of that button, and probably style it better.

But regardless of that, this improvement looks good 👍🏻 !

legalsylvain:
/ocabot merge patch

i'll work on the other point in another PR.

PR #549, opened 2020-10-16
[12.0] ADD pos_session_close_by_tax

eLBati:
This module modifies the computation of journal items generated while closing point of sale session.

The new computation considers the untaxed amount grouped by tax and gets tax amount based on it.

This computation is required by some regulations (like the Italian one).

### DIFF
--- a/pos_session_close_by_tax/__manifest__.py
+++ b/pos_session_close_by_tax/__manifest__.py
@@ -1,0 +1,18 @@
+# Copyright 2020 Lorenzo Battistini @ TAKOBI
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+{
+    "name": "Point of sale: closure by tax",
+    "summary": "Group journal items by tax when closing session",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Point Of Sale",
+    "website": "https://github.com/OCA/pos",
+    "author": "TAKOBI, Odoo Community Association (OCA)",
+    "maintainers": ["eLBati"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "point_of_sale",
+    ],
+}
--- a/pos_session_close_by_tax/models/pos_order.py
+++ b/pos_session_close_by_tax/models/pos_order.py
@@ -1,0 +1,90 @@
+from odoo import models, _
+from odoo.tools import float_round
+
+
+class Order(models.Model):
+    _inherit = 'pos.order'
+
+    def _create_account_move_line(self, session=None, move=None):
+        res = super(Order, self)._create_account_move_line(session, move)
+        if move:
+            self._group_lines_by_tax(move)
+        return res
+
+    @staticmethod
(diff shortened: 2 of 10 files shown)


pos_session_close_by_tax/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "author": "TAKOBI, Odoo Community Association (OCA)",
+    "maintainers": ["eLBati"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "point_of_sale",
+    ],
+    "data": [
+    ],
```
MeetKD: Can we please remove empty key ?

MeetKD (changes requested):
Minor Change.

pos_session_close_by_tax/__manifest__.py
eLBati: Done

matteoopenf:
@MeetKD  is it possible  merge the pr?

matteoopenf (approved):

GSLabIt (approved):

pos_session_close_by_tax/models/pos_order.py
```diff
@@ -0,0 +1,90 @@
...
+    _inherit = 'pos.order'
+
+    def _create_account_move_line(self, session=None, move=None):
+        res = super(Order, self)._create_account_move_line(session, move)
+        if move:
+            self._group_lines_by_tax(move)
+        return res
+
+    @staticmethod
+    def _check_grouping_applicability(lines):
```
ivantodorovich, line 15:
```suggestion
    @api.model
    def _check_grouping_applicability(self, lines):
```

Same for others.
Besides it being "the standard" in odoo code, it makes it easier for other possible modules extending this one to do things like

```py
    def _check_grouping_applicability(self, lines):
        if not self.env.context.get("skip_check"):
            return super()._check_grouping_applicability(lines)
```

Suggestion also applied to the other methods with `@staticmethod`

pos_session_close_by_tax/models/pos_order.py
```diff
@@ -0,0 +1,90 @@
...
+        list_of_date_maturity = lines.mapped('date_maturity')
+        if not all(elem == list_of_date_maturity[0] for elem in list_of_date_maturity):
+            return False
+        if len(lines.filtered(lambda x: x.tax_line_id).mapped('account_id')) > 1:
```
ivantodorovich, line 19:
```suggestion
        if len(lines.filtered('tax_line_id').mapped('account_id')) > 1:
```
If you wanna save some characters ;) not blocking of course
same goes for the line below

ivantodorovich (changes requested):
Thanks!

Check my comment regarding `staticmethod`.

Besides that, it looks good technically.

Functionally speaking I'm not exactly sure why it's needed, besides it being required in italian accounting. If you feel like it, improving the readme would help more people understand it too.. for example you could provide a sample account.move before and after this module in the readme. This is up to you of course, not blocking :+1:

Issue #556, opened 2020-11-05
[BUG][12.0] pos_payment_change: Missing access right - Cannot update payment


r-sierra:
module: pos_payment_change
version: 12.0

**Steps to reproduce**
1 - In Point of Sale configuration:
  * Set **Payment Change Policy** to _Update Payments_
  * Set at least two payment methods

With a user that has _Point of Sale: User_ accesses
2 - Open a new session and make a sale
3 - Logout from POS session
4 - Find the sale and try to change payment

**Current behavior**
Odoo server error is displayed when trying to change the payment:
> Sorry, you are not allowed to delete this document. Only users with the following access level are currently allowed to do that:
> - Technical Settings/Show Full Accounting Features
> - Point of Sale/Manager
> 
> (Document model: account.bank.statement.line) - (Operation: unlink, User: 6) 

**Expected behavior**
Payment should be updated successfully

To solve this issue the _Point of Sale: User_ access right on account.bank.statement.line should be changed to allow delete access

PR #557, opened 2020-11-05
[12.0][FIX] pos_payment_change Add missing unlink access for Point Of Sale:User

r-sierra:
See https://github.com/OCA/pos/issues/556

### DIFF
--- a/pos_payment_change/models/pos_order.py
+++ b/pos_payment_change/models/pos_order.py
@@ -45,7 +45,7 @@ class PosOrder(models.Model):
         )
 
         if self.config_id.payment_change_policy == "update":
-            self.statement_ids.with_context().unlink()
+            self.sudo().statement_ids.with_context().unlink()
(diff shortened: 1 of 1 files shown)


ivantodorovich:
Not sure it's a good idea to add this permission. It will have much more impact than you want.

Instead, what about using `sudo()` to remove the statement?

r-sierra:
@ivantodorovich  Branch has been updated to use `sudo`, thanks for the suggestion.

ivantodorovich:
Awesome ☺️ thanks!
One last thing, could you squash the commits into a single one with a meaningful commit message?

r-sierra:
@ivantodorovich Commits squashed and ammended

ivantodorovich (approved):

ivantodorovich:
fast-tracking simple fix

/ocabot merge minor

PR #560, opened 2020-11-10
 [FIX] pos_customer_display : do not raise error if text === False

legalsylvain:
trivial patch. if text to send to the customer display is null (that occures if pos.config is not set) an error is raised.

CC : @alexis-via, @quentinDupont #GRAPOCA


ivantodorovich (approved):
LGTM :+1:

alexis-via (approved):

quentinDupont (approved):

flotho (approved):
LGTM

PR #566, opened 2020-11-16
 [FIX] pos_tare : raise an error if the gross weight is not correct

legalsylvain:
Raise an error if the gross weight is not correct.

Exemple : 

image

**before the patch**

- a pos order line is added silently, with a qty to 0.

**After the patch**

- an error is raised.

image

(Also fix an bad title for another warning).

CC : 
- @grap, @quentinDupont, @sandiefavre (Ref : #/board/144/card/1080)
- @Fkawala

### DIFF
--- a/pos_tare/static/src/js/screens.js
+++ b/pos_tare/static/src/js/screens.js
@@ -36,6 +36,8 @@ odoo.define('pos_tare.screens', function (require) {
         },
     });
 
+    var _super_ScaleScreenWidget_order_product = screens.ScaleScreenWidget.prototype.order_product;
+
     screens.ScaleScreenWidget.include({
 
         // /////////////////////////////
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):
Code review LGTM :+1:

Fkawala (approved):

quentinDupont (approved):

Fkawala:
I see much activity here, hence I'll shamelessly try to reroute a little bit of it to a related PR #527 that has been waiting to get approved and merged for a while. Sorry + thank you.

legalsylvain:
Note : also add another fix, regarding a bad overload of the function ``validate_order`` that is not overloadable by design. (*) (see similar problem for ``pos_meal_voucher`` [here](https://github.com/OCA/pos/pull/565).

the confirm action is moved on the scale screen, when user is weighing its product.


image



(*) : With the curent design, the other core checks are skipped. for exemple, with the non patched current design of ``pos_tare``, the check regarding huge amount is not done.

image

CC : @quentinDupont

houssine78 (approved):

houssine78:
@legalsylvain I think that this can be merged

legalsylvain:
Thanks. indeed, using this patch in production since a month without problem.

/ocabot merge patch

PR #574, opened 2020-12-18
14.0 mig pos backend communication rebased

sebastienbeau:


hparfr (approved):
LGTM

pos_backend_communication/static/src/js/back.js
```diff
@@ -0,0 +1,47 @@
+odoo.define("pos_backend_communication.back", function (require) {
+    "use strict";
+    var tools = require("pos_backend_communication.tools");
```
ivantodorovich:
```suggestion
    const tools = require("pos_backend_communication.tools");
```

Non blocking, but I'd use `const` for all the `var`s

ivantodorovich (changes requested):
Can you squash all your Migration commits into a single "[MIG] pos_backend_communication: Migration to 14.0" commit?

ivantodorovich:
It'd also be good to take the opportunity to squash some useless commits to keep a clean history:

image

sebastienbeau:
@ivantodorovich done

ivantodorovich (approved):
Thanks! ☺️

sebastienbeau:
thanks

PR #577, opened 2020-12-22
[ADD] pos_escpos_status module

PierrickBrun:
With the new printing mechanism in Odoo PoS, huge work would be required to fetch 'status' correctly from ESC/PoS printers in pywebdriver.

ref: https://github.com/odoo/odoo/commit/03b2f7b77dc6189bb485e0b834dba5f6d3d4da2c#diff-78bcd46ec9b20eca4256aa02d1d54549c4fa893b64302a799630dc42cca8ccf3 

This module allows to use the old mechanism to fetch the printer's status and thus make pywebdriver work easily. (Printing with pywebdriver does not require it, only for displaying the printer's status)


legalsylvain (changes requested):
Hi @PierrickBrun Thanks a lot for this module. 
However, I think that setting pywebdriver under OCA umbrella should be done before accepting this module.

PierrickBrun:
> Hi @PierrickBrun Thanks a lot for this module.
> However, I think that setting pywebdriver under OCA umbrella should be done before accepting this module.

That is indeed needed. I proposed it here because I wanted to make sure that anyone needing this would find it easily.

For reference, here is the issue about pywebdriver's OCA inclusion: https://github.com/akretion/pywebdriver/issues/57

alexis-via (approved):
Tested with a real Epson TM-T20 printer and pywebdriver. Works very well.

hparfr (approved):
LGTM

ivantodorovich (approved):
Even thought this was built for pywebdriver, it seems a generic-enough module :+1: who knows, maybe it can be used in other cases too

PR #582, opened 2020-12-23
 [FIX] pos_user_restriction: disable test that are failing if pos_restaurant module is installed

legalsylvain:
The tests of the module ``pos_user_restriction`` is failing if ``pos_restaurant`` is installed.

Don't know exactly how to fix it. proposing this patch for the time being, if nobody can fix it quickly.

CC : @eLBati, @danimv5

The error is blocking all new pos module that depends on ``pos_restaurant``, like : https://github.com/OCA/pos/pull/579

kind regards.


eLBati (approved):
OK for now

PR #583, opened 2020-12-27
[14.0][ADD] pos_require_customer

hparfr:
Block orders with no customer set

### DIFF
--- a/pos_require_customer/static/src/js/models.js
+++ b/pos_require_customer/static/src/js/models.js
@@ -1,0 +1,25 @@
+/* Copyright 2020 Akretion (https://www.akretion.com)
+ * @author Raphaël Reverdy <raphael.reverdy@akretion.com>
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+odoo.define("pos_require_customer.models", function (require) {
+    "use strict";
+
+    var models = require("point_of_sale.models");
+
+    models.PosModel.prototype.models.some(function (model) {
+        if (model.model !== "product.pricelist") {
+            return false;
+        }
+        model.fields.push("pos_require_customer");
+        return true; // Exit early the iteration of this.models
+    });
+
+    models.Order = models.Order.extend({
+        is_customer_required: function () {
+            var order = this;
+            var pricelist = order.pricelist;
+            return pricelist.pos_require_customer;
+        },
+    });
+});
--- a/pos_require_customer/__init__.py
+++ b/pos_require_customer/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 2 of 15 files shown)


legalsylvain (commented):
hi @hpar, thanks for your work. 
A question, do you know the existing V12 module ``pos_customer_required`` ? 
https://github.com/OCA/pos/tree/12.0/pos_customer_required

It does similar things. (block / allow to sell without customer defined).

Should be great to extend the existing module instead of having a new one. Don't you think ? 

The existing one defines settings on ``pos.config``, you are defining settings on ``product.pricelist``. But the rest of the javascript is common.

CC : @ivantodorovich

ivantodorovich:
Indeed both modules are doing pretty much the same thing.
I agree with @legalsylvain , it's better to migrate the existing module and improve it instead

hparfr:
Thanks for pointing me out this module. I completely forget this module exists before proposing this pr.

I did a quick look at it, the goal is identical but the behavior is quite different. 
I will take some time to think about it, but I'm not sure it's worth it to merge these modules.

ivantodorovich:
Hi @hparfr ,

Indeed it's slightly different..

Coincidently enough I've just got a request for a similar feature, only we need to configure the customer_required per product (not per pricelist, neither per pos config).

I can't imagine how yet but I think there's probably a good way to integrate the existing features + all of this into a single module. I just mention this for the sake of discussion, I don't have an idea myself :sweat_smile: 

For the time being I'm not sure it's a good idea to have several "customer required" modules in OCA. I think it's better to find a way to integrate all into 1 module or to treat these requests as custom developments

pos_require_customer/static/src/js/models.js
```diff
@@ -0,0 +1,25 @@
...
+
+    var models = require("point_of_sale.models");
+
+    models.PosModel.prototype.models.some(function (model) {
+        if (model.model !== "product.pricelist") {
+            return false;
+        }
+        model.fields.push("pos_require_customer");
+        return true; // Exit early the iteration of this.models
+    });
```
ivantodorovich, line 16: Use `load_fields` instead:

https://github.com/odoo/odoo/blob/c0295d5844e2924e684312245e56e82be92458c1/addons/point_of_sale/static/src/js/models.js#L1435-L1439


```suggestion
    models.load_fields("product.pricelist", ["pos_require_customer"]);
```

pos_require_customer/static/src/js/models.js
```diff
@@ -0,0 +1,25 @@
...
+
+    models.Order = models.Order.extend({
+        is_customer_required: function () {
+            var order = this;
+            var pricelist = order.pricelist;
+            return pricelist.pos_require_customer;
```
ivantodorovich, line 22:
```suggestion
            return this.pricelist.pos_require_customer;
```

nitpicking here but the one-liner is more elegant ;)

ivantodorovich (changes requested):

ivantodorovich:
Given that there may be more than 1 approach to require customers in the PoS, and that they're all valid and doesn't make sense to include them in the same module (at least for now), may I suggest renaming this module to `pos_pricelist_require_customer` ?

flotho:
hi there,

superseeded by #748  ?

PR #585, opened 2021-01-11
[12.0] IMP pos_partner_firstname allowing to use tab to move from firstname to lastname

eLBati:


eLBati:
@robyf70 what do you think?

robyf70 (approved):
That LGTM

eLBati:
@robyf70 you should be allowed to run `ocabot merge patch` if you want

PR #586, opened 2021-01-11
 [FIX] pos_payment_change: use original date_order for the new orders and the payments

legalsylvain:
### in "update" mode

**Before** 
- the payment lines have the date of the correction.

**After**
- the payments lines have the date of the original order.

### in "refund" mode 

**Before** 
- the refund order doesn't have the correct session and have the date of the correction

** After**
- the refund order doesn't have the same session and the same date as the original order.

CC : @quentinDupont  (board/144/card/1243)

### DIFF
--- a/pos_payment_change/models/pos_order.py
+++ b/pos_payment_change/models/pos_order.py
@@ -5,7 +5,7 @@
 
 from datetime import datetime
 
-from odoo import _, api, fields, models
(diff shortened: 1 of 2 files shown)


legalsylvain:
Patch used in production since 6 monthes.
/ocabot merge patch

PR #589, opened 2021-01-18
[MIG] 14.0 pos report session summary

PierrickBrun:


ivantodorovich (approved):
Lgtm! Thanks!

It'd be great if you can squash translation and administrative commits

alexis-via:
The pos session report works. But I think it should be re-designed because, in v14, o.statement_ids only has the cash journal, not the other journals (card, checks). So this report only display cash information, which is very partial.
I think it would be interesting to have some info about the pos orders (number of pos orders, total with tax, total without tax, total tax, ...)

alexis-via:
In v14, you can call self.env['report.point_of_sale.report_saledetails'].get_sale_details(self, date_start=False, date_stop=False, config_ids=False, session_ids=session_ids)
The session_ids arg didn't exist before.
With that, you can get all the info you need.

legalsylvain:
Hi @PierrickBrun . any news regarding @alexis-via remarks ?

elvise:
@ivantodorovich any chance for this PR ?

legalsylvain:
@elvise this PR needs fixing

PierrickBrun:
done in another PR

PR #596, opened 2021-02-02
[12.0][FIX][pos_report_order_payment] Avoid possible division by zero

ivantodorovich:
This fixes a possible division by zero due to some orders having amount_total = 0

### DIFF
--- a/pos_report_order_payment/report/pos_order_payment_report.py
+++ b/pos_report_order_payment/report/pos_order_payment_report.py
@@ -148,9 +148,10 @@ class PosOrderPaymentReport(models.Model):
                 pol.id AS pos_line_id,
                 sl.id AS statement_line_id,
                 st.journal_id AS journal_id,
-                (
-                    sl.amount / po.amount_total * pol.price_subtotal_incl
-                ) AS amount
+                CASE po.amount_total
+                    WHEN 0 THEN 0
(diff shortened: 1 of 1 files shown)


pos_report_order_payment/report/pos_order_payment_report.py
```diff
@@ -148,9 +148,10 @@ def _payment_query(self):
                 pol.id AS pos_line_id,
                 sl.id AS statement_line_id,
                 st.journal_id AS journal_id,
-                (
-                    sl.amount / po.amount_total * pol.price_subtotal_incl
-                ) AS amount
+                CASE po.amount_total
+                    WHEN 0 THEN 0
+                    ELSE sl.amount / po.amount_total * pol.price_subtotal_incl
```
legalsylvain, line 153: And if price_subtotal_incl is 0. Can occures with a discount of 100%. Dont'you think ?

Otherwise. Lgtm. Thanks a lot !

ivantodorovich: That's actually evaluated as `(sl.amount / po.amount_total) * pol.price_subtotal_incl`, so it's not an issue

legalsylvain: Oh ! You're right.
Thanks.

legalsylvain:
Fast track.

/ocabot merge patch

PR #597, opened 2021-02-03
[14.0] Add module pos_no_cash_bank_statement

alexis-via:
About this new module:

Up to Odoo 12.0, POS sessions generated bank statements for each payment method and therefore closing a POS session would generate payment journal entries for all payment methods.

In Odoo 13.0 and upper, only the cash payment method generate a cash register/bank statement and therefore the payment journal entries. The non-cash payment methods don't generate any payment journal entries.

With this module, you have a new option **Generate Bank Statement** on non-cash POS payment method. If you enable that option, Odoo will generate a bank statement and will generate the payment journal entries upon POS session closing. For example, if your POS accepts payment by check, you certainly want to enable that option for the *Check* POS payment method.

### DIFF
--- a/pos_no_cash_bank_statement/models/pos_payment_method.py
+++ b/pos_no_cash_bank_statement/models/pos_payment_method.py
@@ -1,0 +1,76 @@
+# Copyright 2021 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class PosPaymentMethod(models.Model):
+    _inherit = "pos.payment.method"
+
+    cash_journal_id = fields.Many2one(
+        domain=[("type", "in", ("bank", "cash"))],
+        string="Bank/Cash Journal",
+    )
+    journal_type_domain = fields.Char(compute="_compute_journal_type_domain")
+    bank_statement = fields.Boolean(
+        string="Generate Bank Statement",
+        help="By default, Odoo will generate a cash register/bank statement "
+        "upon POS session closing only for cash payment methods. If you enable "
+        "this option for a non-cash payment method, Odoo will generate a bank "
+        "statement upon session closing, which will generate accounting entries "
+        "in the bank journal, which will be automatically reconciled with the "
+        "sale journal entry.",
+    )
+
+    @api.constrains("is_cash_count", "cash_journal_id", "bank_statement")
+    def _check_journal_config(self):
+        for method in self:
+            if method.is_cash_count:
+                if not method.cash_journal_id:
+                    raise ValidationError(
+                        _("Missing cash journal on cash payment method '%s'.")
+                        % method.display_name
+                    )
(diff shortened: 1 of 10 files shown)


pos_no_cash_bank_statement/models/pos_payment_method.py
```diff
@@ -0,0 +1,74 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class PosPaymentMethod(models.Model):
+    _inherit = "pos.payment.method"
+    _check_company_auto = True
+
+    cash_journal_id = fields.Many2one(
+        check_company=True,
```
legalsylvain: Why do you add this attribute ?

legalsylvain (commented):
Light code review. Lgtm. Thanks.

pos_no_cash_bank_statement/models/pos_payment_method.py
alexis-via: Because I feel that it is lacking in the point_of_sale module. In fact, this kind of code is not required for this module, I used it as an opportunity to improve the native code. If can remove it from this module if you prefer.
BTW, _check_company_auto =True and check_company=True are still largely ignored by Odoo/OCA devs, and I feel it's really important since odoo v13 which introduced the possibility to have access to several companies at the same time, which adds a risk of selecting a journal from another company than the pos.payment.method. _check_company_auto =True and check_company=True will auto-add a constraint to have an explicit error when saving (cf the ORM code).

legalsylvain: Hi ! Thanks for your answer.

1) I think you're right regarding the lack of use of ``check_company``.
2) I think it's not the good place to add it here.

If ``check_company`` are missing in many places, I think it's better to have some very simple modules that adds that. 

Idea by module : 
- ``account_check_company``
- ``pos_check_company``
- ``crm_check_company`` ...

or by model: 
- ``account_journal_check_company`` ...

The good place should be in https://github.com/OCA/multi-company/

What do you think ?

alexis-via: I agree with you, it's not the role of this module to fix that kind of issues. So I removed _check_company_auto = True and check_company=True

alexis-via:
I just implemented Pedro's idea about how to replace domain update via onchange (deprecated in v14). It works nicely! No more depreciation warnings in the odoo server logs.

alexis-via:
Arf. I hate these bots! This PR is ready to merge.

ivantodorovich (approved):

ivantodorovich:
It'd be nice to add a test case, but technically looks good :+1:

sbidoul:
The main branch is green and the test failure does not seem to be related to this module so retrying the merge.

/ocabot merge patch

@alexis-via thanks a lot for this module it works like a charm!

Now the only missing bit is to see if we can combine the split transaction mode with having one move on the receivable journal, to avoid many useless customer/bank transactions.

And, more importantly, to see if we can get the electronic payment transaction reference as move line label, to facilitate reconciliation.

sbidoul:
Hm, so this module actually reveals a small bug in a test of `pos_payment_change`, which I address in #743

sbidoul:
So this should pass now.

/ocabot merge patch

PR #598, opened 2021-02-03
[13.0][MIG] pos_payment_restriction

glitchov:


legalsylvain (commented):
Hi @glitchov . I don't understand why the setting is done on the ``pos.config`` and not on the ``account.journal``.

ThomasBinsfeld (approved):

ThomasBinsfeld:
> Hi @glitchov . I don't understand why the setting is done on the `pos.config` and not on the `account.journal`.

Since the `payment_amount_readonly` setting is strongly linked to the 'Prefill Cash Payment' setting, i think it's better to leave it on `pos.config` don't you?

legalsylvain:
> Since the payment_amount_readonly setting is strongly linked to the 'Prefill Cash Payment' setting, i think it's better to leave it on pos.config don't you?

Well, in some case, we can imagine to use this feature for journal (type = 'bank').
- if cashier has a card reader, that return paid amount.
- if cashier receives meal vouchers, so that the amount is automatically set by the barcode of the meal voucher.
- maybe other use case...

for that reason, I think that it should be done in ``account.journal`` and not in ``pos.config``.

pos_payment_restriction/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "development_status": "Beta",
     "maintainers": ["rousseldenis"],
-    "category": "Point of Sale",
+    "category": "Sales/Point Of Sale",
```
legalsylvain, line 12: I'm not aware of conventions regarding categories, and i never know what to write.
Can you tell me where it is? 

Thanks !

rousseldenis: > I'm not aware of conventions regarding categories, and i never know what to write.
> Can you tell me where it is?
> 
> Thanks !

The problem I just faced is during migration if categories changed.
Former one : https://github.com/odoo/odoo/blob/9.0/addons/point_of_sale/__openerp__.py#L7
New one: https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/__manifest__.py#L7

I have a foreign key error on module category, solved changing manifest like this.

For info, OCA shop contains Odoo aggregated categories through core modules : 
image

I see now they are not sync since shop development. Maybe something to do.

To summarize, it's always recommended to fit OCA shop categories in order to be under the good filter.

pos_payment_restriction/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+In some situations, one would have user restrictions on payment lines level.
+
+This includes :
+
+* An option to make payment amount readonly
+* An option that adds a helper to disable the 'Validate' button (e.g.:
+  waiting for an automatic payment transaction to be completed)
```
sbidoul, line 7: @glitchov this part about the validation module is not part of this module anymore, right. It seems to be standard Odoo.

sbidoul:
@legalsylvain 

> for that reason, I think that it should be done in account.journal and not in pos.config

In 13, you probably `pos.payment.method` rather than `account.journal` ?

But I think your use cases are different. This module is meant for forcing the user to pay the whole amount with a single payment method.

ivantodorovich (approved):
LGTM 

Can you squash your last 3 commits into a single one?

etobella:
@glitchov Can you apply the proposed changes?

legalsylvain:
@glitchov friendly reminder.

PR #611, opened 2021-03-18
 [ADD] pos_order_line_no_unlink: new module to prevent unlinking order lines

legalsylvain:
This is a trivial module to prevent to unlink lines of confirmed / invoiced orders. 

CC : @grap, @quentinDupont 
apps/deck/#/board/144/card/1341


houssine78 (approved):

pos_order_line_no_unlink/models/pos_order_line.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo.exceptions import UserError
+
+
+class PosOrderLine(models.Model):
+    _inherit = "pos.order.line"
+
+    @api.multi
+    def unlink(self):
+        if self.filtered(
+            lambda x: x.order_id.state not in ["draft", "cancel"]
```
houssine78, line 15: I'll only allow it for draft.

legalsylvain: Well, it is for consistency reason with the current design.
https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/models/pos_order.py#L700

in fact, 'cancel' orders are not possible in V12. it is kept only for legacy reason, I guess.
see : https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/models/pos_order.py#L785

flotho:
Hi @legalsylvain , additionnaly won't the l10n_fr_pos module prevent the unlink in case of certification?

legalsylvain:
Hi @flotho indeed, in case of certification (french or other) this module should be installed.

flotho (approved):
LGTM

PR #612, opened 2021-03-18
[FIX] Fixed order date tz bug. Fixed payment lines bug.

wpichler:
Two bugs:
 - order date should be in tz of current user - and get formatted according the local datetime format
 - paymentlines do not use cashregister in v13 - use payment method instead

### DIFF
--- a/pos_order_mgmt/models/pos_order.py
+++ b/pos_order_mgmt/models/pos_order.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, fields, models
+import pytz
 
 
 class PosOrder(models.Model):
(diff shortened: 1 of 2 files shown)


pos_order_mgmt/models/pos_order.py
```diff
@@ -110,9 +111,12 @@ def _prepare_done_order_for_pos(self):
         for payment_line in self.payment_ids:
             payment_line = self._prepare_done_order_payment_for_pos(payment_line)
             payment_lines.append(payment_line)
+        tz = self.env.user.tz
```
chienandalu, line 114: Should we follow the same defaults and fallbacks as in core?: https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/models/pos_order.py#L895

I also wonder if we could/should try to guess the real tz of the order as it was originally recorded?

wpichler: we could use the original order user_id timezone instead of the current user timezone. Its unlikely that the same sales person is working on different locations in different timezones. Yes it could be - and according to murphy's law it will be - but i think this is out of scope...

chienandalu (changes requested):
Ok then. Run precommit over your changes to pass CI

chienandalu:
@wpichler AFAIC this only lacked a pre-commit run? Are you planning to fix it?

wpichler:
@chienandalu How to run a pre-commit run ?

chienandalu:
> @chienandalu How to run a pre-commit run ?

Here's the info: https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97

PR #616, opened 2021-04-07
[13.0][ADD] pos_deposit module.

thomaspaulb:
This module allows the use of deposit products for PoS products.

An example of such a deposit product would be a bottle of cola, where the bottle is
being sold for an extra 25 cents, which you get back when you bring the bottle back to
the store intact. Whenever the cola is added to a PoS order, the bottle is added too.
If the cola quantity changes, the bottle quantity will change as well.

Thanks to Open2bizz for the initial 8.0 implementation and their permission to publish the module in OCA.


pos_deposit/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,13 @@
...
+#. Select or add at least one 'deposit product', such as 'Bottle 1L',
+   with a non-zero price and with the checkbox 'Is Deposit' set to True.
+
+#. Select or add at least one product of which 'Use Deposit' is checked,
+   such as '1 liter of cola'. In the accompanying product selection field,
+   select the bottle. The two products are now linked.
+
+#. When someone brings back bottles to the store, of course you want to
+   refund him. For that purpose, create new products of negative price,
+   such as 'Return/refund bottle 1L'. These products don't need to be
```
legalsylvain: That's not the better to handle stock of deposited product.
Why not just refund the bottle. The stock will be ok. Don't you think ?

pos_deposit/static/src/js/pos.js
```diff
@@ -0,0 +1,90 @@
...
+                        var dep_line = new models.Orderline({}, {
+                            pos: this.pos,
+                            order: this,
+                            product: dep_prod
+                        });
+                        this.add_orderline(dep_line);
+                        dep_line.deposit_for = last;
+                        last.deposit_line = dep_line;
+                    } else {
+                        alert ('Product for deposit is not configured properly !');
```
legalsylvain: Better to handle with aconstraint in pyhton. don't you think? It' avoid lazy error.

legalsylvain (commented):
Partial code review. No test.

Thanks for sharing this module !

pos_deposit/static/src/js/pos.js
thomaspaulb: We already have the constraint in the view, but not in Python. IMO it's enough in the view, what do you think

pos_deposit/readme/CONFIGURE.rst
thomaspaulb: True, but typically people bring this in as part of their next order, so the refund of several bottles would need to be simultaneous with the buying of new products. Is that something Odoo 13.0 supports natively to your knowledge? (Excuse my laziness to test it myself)

thomaspaulb: I've tested and we can just sell the bottle with a negative quantity, I guess that's what you meant also. I've removed this comment from the README.

legalsylvain: indeed. But we can imagine that when scanning a product with deposit, it is looking for the according deposited product and make +1 on quantity. Then, if the customer returns some bottles, cashier can set -x on quantity. Do'nt you think ?

pos_deposit/static/src/js/pos.js
thomaspaulb: I've removed this alert in any case.

pos_deposit/readme/CONFIGURE.rst
thomaspaulb: Yes true! So I think this is handled.

gjotten (commented):
@thomaspaulb functionally on Runbot, I configure LED Lamp with 'is deposit' == true, and Office Lamp wtih 'has deposit' == 'LED Lamp'. 

LGTM, as long as I first enter the 'is_deposit' product in the PoS session with its negative amount first if the customer returns it to the store. Afterwards, I can add the 'use_deposit' product with that same 'is_deposit' product linked to all my heart's content, resulting in 3 lines in the session (negative LEDs, positive Lamps, and postive LEDs).

Is that how we mean to implement this? If so, I gather some explanation in the readme would be useful.

pos_deposit/models/product_template.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Sunflower IT
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
```
legalsylvain, line 8: Hi. 
I don't think that it is a ``product.template`` but a ``product.product`` value. Typical Use cas : 
- product template : Cola
- variant 1 : 25cl ; variant 2 : 50 cl. 
-> variant 1 : deposit product = bottle 25cl ; variant 2 : deposit product = 50cl.

don't you think ? 

then we can implement a related field on template (as per other fields like ``barcode`` to make it available on template form, in a single variant context). 

Don't you think ?

pos_deposit/models/product_template.py
```diff
@@ -0,0 +1,14 @@
...
+
+    use_deposit = fields.Boolean("Use Deposit")
```
legalsylvain, line 10: I don't see the interest of this field. If the deposit product is defined, it is using deposit.

pos_deposit/models/product_template.py
```diff
@@ -0,0 +1,14 @@
...
+    is_deposit = fields.Boolean("Is Deposit")
+    select_deposit = fields.Many2one(
```
legalsylvain, line 12: use Odoo convention, renaming to ``deposit_product_id``. (Many2one should end by ``_id``).

pos_deposit/static/src/js/pos.js
```diff
@@ -0,0 +1,90 @@
...
+                var dep_line = new models.Orderline(
+                    {},
+                    {
+                        pos: this.pos,
+                        order: this,
+                        product: dep_prod,
+                    }
+                );
+                this.add_orderline(dep_line);
+                dep_line.deposit_for = last;
```
legalsylvain, line 66: I understand the reason to implement the ``deposit_for`` feature, but I think that it is too complex and unnecessary : 

- it generates the limitation described in the roadmap section.
- it makes more complex the use case : product 1 with Deposit Product A + product B with deposit product A that is a common use case.

I think removing this feature will simplify a lot the JS code. Don't you think ?

legalsylvain (changes requested):
some remarks inline.

Nice to have : could be great to have demo data. 

I made some suggestion on your PR. Do you want I try a PoC ? 

thanks.

PR #618, opened 2021-04-12
[UPD] README file

HaraldPanten:
Incompatibility with pos_cache module from Odoo

### DIFF
--- a/pos_product_sort/readme/ROADMAP.rst
+++ b/pos_product_sort/readme/ROADMAP.rst
@@ -1,1 +1,2 @@
 * add settings on ``pos.config`` to let users choose the field(s) to sort the products.
+* This module is NOT compatible with pos_cache module from Odoo.


legalsylvain:
Hi. @HaraldPanten. thanks for the insight. 
Could you precise the incompatibility ? is the PoS broken, will both module are installed, or just the sort doesn't work ?  

If ``pos_product_sort`` is breaking pos if ``pos_cache`` is installed, we could imagine to add a strong exclusion in the manifest. `` 'excludes': ['pos_cache'],`` 

like here for exemple : https://github.com/OCA/edi/blob/12.0/account_e-invoice_generate/__manifest__.py#L15

Don't you think ? 

kind regards.

HaraldPanten:
Hi @legalsylvain ,

The POS will not be broken, but if you have `pos_cache` installed the functionality of `pos_product_sort` is not working well; I mean... products are not sorted properly regarding `pos_product_sort` module.

I'm not sure what would be better, but IMO I'd put it in the ROADMAP (better explained by my side) instead of excluding the module in the manifest.

I think that this could be solved in a short term.

WDYT?

legalsylvain:
Indeed, if the PoS is not broken, a simple text is enough. 

fast merging.

/ocabot merge patch

PR #619, opened 2021-04-14
Modified the Document of Origin in pos_order.py of module pos_order_return

emavarela:
Modified the document of origin in the module to reproduce the behaviour of the frontend when creating a return out of an invoice with the wizzard.
With these changes the Document of Origin will be the actual document number instead of the Journal movement ID

### DIFF
--- a/pos_order_return/models/pos_order.py
+++ b/pos_order_return/models/pos_order.py
@@ -51,9 +51,9 @@ class PosOrder(models.Model):
         if not self.returned_order_id.invoice_id:
             return res
         res.update({
-            'origin': self.returned_order_id.invoice_id.number,
+            'origin': self.returned_order_id.invoice_id.document_number,
             'name': _(
-                'Return of %s' % self.returned_order_id.invoice_id.number),
+                'Return of %s' % self.returned_order_id.invoice_id.document_number),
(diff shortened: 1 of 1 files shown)


pos_order_return/models/pos_order.py
```diff
@@ -51,9 +51,9 @@ def _prepare_invoice(self):
         if not self.returned_order_id.invoice_id:
             return res
         res.update({
-            'origin': self.returned_order_id.invoice_id.number,
+            'origin': self.returned_order_id.invoice_id.document_number,
```
legalsylvain, line 54: Hi @emavarela. what is this field ? 
Travis is failing on that line : https://travis-ci.com/github/OCA/pos/jobs/498674882#L1886

regards.

chienandalu (changes requested):
I'm afraid your changes attend to a field that is not in the standard Odoo core (maybe it comes from a customization of yours? Please clarify

emavarela:
Hi Community good morning,

Sorry for the delay. Yes indeed the field comes from another custom module. I should have checked that before submiting the request.

Sorry for that, have a nice day.

legalsylvain:
no problem @emavarela. thanks for your reply, and for closing the issue.

have a nice day too.

PR #622, opened 2021-04-15
[12.0][ADD] pos_payment_method & pos_payment_method_adyen

AdriaGForgeFlow:
This MR adds two modules:

* pos_payment_method:

This module enables the addition of different payment methods with its own integration
that need to have a specific behaviour when confirming / canceling a Point of Sale order.

* pos_payment_method_adyen

This module adds an integration between Odoo and Adyen payment terminals.

It is based on Odoo version 13 integration and has some added features such as
Passive and Active Shopper Recognition.

When the payment method is configured in the Point of Sale configuration the user
can send the amount of the order directly to the Adyen terminal and the capturing
of the money will be automatically done in Adyen.

Adyen Documentation: https://docs.adyen.com/point-of-sale

cc ~ @ForgeFlow


chienandalu (commented):
Very interesting. I guess this won't be going outside draft for this version as it's implemented with ES6, isn't it? What's the roadmap for it?

It'd be interesting to remark the differences with `pos_payment_terminal` though.

chienandalu:
@AdriaGForgeFlow are you still working on this?

AdriaGForgeFlow:
> @AdriaGForgeFlow are you still working on this?

We have recently launched this code in production for one of our clients and it seems to be working properly without any major issue. The only problem with this implementation is that you require Odoo up because the communication with Adyen happens from the back end. 
Next feature would be a way of enabling local communications between the PoS and the terminal.

PR #630, opened 2021-04-26
[14.0] [mig] pos_access_right

hkapatel-initos:


pos_access_right/models/pos_config.py
```diff
@@ -57,33 +57,33 @@ class PosConfig(models.Model):
     )
 
     def _compute_group_negative_qty_id(self):
-        for config in self:
+        for _config in self:
```
legalsylvain: Hi. Why this change ?

pos_access_right/static/src/js/pos_access_right.js
```diff
@@ -205,7 +201,6 @@ odoo.define("pos_access_right.pos_access_right", function (require) {
          * Block 'Payment' button if user doesn't belong to the correct group
          */
         renderElement: function () {
-            var self = this;
```
legalsylvain: Hi. I don't understand how the line ``self.gui.get_user_groups();`` can work if you remove this line. Did I missed something ?

pos_access_right/models/pos_config.py
hkapatel-initos: @legalsylvain to avoide Finding  config from system Odoo and avoid this error ( Loop control variable 'config' not used within the loop body.) If this is intended, started the name with an underscore.

pos_access_right/static/src/js/pos_access_right.js
hkapatel-initos: @legalsylvain  hello,
  use this.gui.get_user_groups(); instade of self.gui.get_user_groups();

brendapaniagua (approved):
LGTM! Functional review!

hkapatel-initos:
@ivantodorovich can you please review this PR ?

pos_access_right/models/pos_config.py
ivantodorovich: The real error here is not how you name the loop variable, is that it's not being used.
In computed field methods, self can be a recordset.

I would write these methods like this:

```py
def _compute_group_negative_qty_id(self):
    self.update({
        "group_discount_id": self.env.ref("pos_access_right.group_discount").id,
    })
```

As you don't really need the loop..


In fact there are so many computed methods for the same purpose here... I would also join them all in a single `_compute_groups` method

 ```py
def _compute_groups(self):
    self.update({
        "group_discount_id": self.env.ref("pos_access_right.group_discount").id,
        "group_change_unit_price_id": self.env.ref("pos_access_right.group_change_unit_price").id,
        # ... etc... 
    })
```

File: pos_access_right/static/src/js/ActionpadWidget.js [hunk 3]
```diff
@@ -0,0 +1,33 @@
...
+
+    const PosActionpadWidget = (ActionpadWidget) =>
+        class extends ActionpadWidget {
+            get currentUser() {
+                const cashier =
+                    this.env.pos.get("cashier") || this.env.pos.get_cashier();
+                const user = this.env.pos.users.filter(
+                    (o) => o.id === cashier.user_id[0]
+                );
+                return user[0];
```
ivantodorovich: You can use find instead of filter

```js
return this.env.pos.users.find(user => user.id === cashier.user_id[0]);
```

pos_access_right/static/src/js/ActionpadWidget.js
```diff
@@ -0,0 +1,33 @@
...
+    "use strict";
+
+    const Registries = require("point_of_sale.Registries");
+    const ActionpadWidget = require("point_of_sale.ActionpadWidget");
+
+    const PosActionpadWidget = (ActionpadWidget) =>
+        class extends ActionpadWidget {
+            get currentUser() {
+                const cashier =
+                    this.env.pos.get("cashier") || this.env.pos.get_cashier();
```
ivantodorovich: This should be enough.. no?

```js
const cashier = this.env.pos.get_cashier();
```

pos_access_right/static/src/js/ActionpadWidget.js
```diff
@@ -0,0 +1,33 @@
...
+                    this.currentUser &&
+                    this.currentUser.groups_id.indexOf(
+                        this.env.pos.config.group_payment_id[0]
+                    ) !== -1
+                ) {
+                    return true;
+                }
+                return false;
+            }
+        };
```
ivantodorovich, line 12:
```js
get hasPaymentControlRights() {
    return this.currentUser.groups_id.includes(this.env.pos.config.group_payment_id[0]);
}
```

You can do the same with the others

pos_access_right/static/src/js/NumpadWidget.js
```diff
@@ -0,0 +1,63 @@
...
+    const PosNumpadWidget = (NumpadWidget) =>
+        class extends NumpadWidget {
+            get currentUser() {
+                const cashier =
+                    this.env.pos.get("cashier") || this.env.pos.get_cashier();
+                const user = this.env.pos.users.filter(
+                    (o) => o.id === cashier.user_id[0]
+                );
+                return user[0];
+            }
```
ivantodorovich: I see this is repeated widget after widget. If you're repeating code over and over, it's a good indicator something needs to be refactored.

Consider moving all your `hasSomeGroup` methods to the `PoS` model instead, as that seems to be a better place to hold them. It's possible you want to check if the user has the same group on different widgets, and if you keep the current implementation you would have to copy the whole method again..

Alternatively, I would try removing all these methods entirely and "enrich" the "employee" objects with `hasGroupSomething` properties instead.

See: https://github.com/odoo/odoo/blob/3780fa2af5d5f6cac91e419bcab69a253db280bd/addons/point_of_sale/static/src/js/models.js#L325-L343

If you can extend that method, you can loop through the users in `this.users` and add properties to `user.employee`. Since this "employee" object is the one used in `set_cashier` / `get_cashier`, you'll have the properties everywhere by simply calling `get_cashier`

I think you should be able to do something like this in a `models.js` file:

```js
    const models = require("point_of_sale.models");
    const users = _.findWhere(models.PosModel.prototype.models, {model: "res.users"});
    const _users_loaded_super = users.loaded;
    users.loaded = function(self, users) {
        _users_loaded_super.apply(this, arguments);
        for (const user of self.users) {
            user.employee.hasGroupManualDiscount = user.groups_id.includes(self.config.group_discount_id[0]);
            // etc.. do the same for other groups
        }
    }
```
 
And then you can simply do this when you need to check if the user has a given group. In fact you can do this directly in qweb views so you avoid writing a lot of js to extend widgets..

```js
this.env.pos.get_cashier().hasGroupManualDiscount
```

pos_access_right/static/src/js/pos_access_right.js
```diff
@@ -1,235 +0,0 @@
...
-                if (user.user_id && user.user_id[0] == user_id.id) {
-                    return user_id;
-                }
-            });
-        }
-        if (user_groups) {
-            this.gui.display_access_right(user_groups);
-        }
-        _set_cashier_.call(this, user);
-    };
```
ivantodorovich: I don't see this `set_cashier` method in the new code.
Could it be that you forgot to migrate this feature?

hkapatel-initos: @ivantodorovich  Don't forget I have posted employee group for v14 select employee and switch cashier
                    this.currentUser &&
                    this.currentUser.groups_id.indexOf(
                        this.env.pos.config.group_payment_id[0]
                    ) !== -1

pos_access_right/static/src/js/NumpadWidget.js
hkapatel-initos: @ivantodorovich  Thanks for your suggestions  i'm working on it

pos_access_right/static/src/js/models.js
```diff
@@ -0,0 +1,40 @@
...
+                );
+            pos_cashier.hasGroupMultiOrder =
+                cashier &&
+                cashier.groups_id.includes(this.env.pos.config.group_multi_order_id[0]);
+            pos_cashier.hasGroupDeleteOrder =
+                cashier &&
+                cashier.groups_id.includes(
+                    this.env.pos.config.group_delete_order_id[0]
+                );
+            return pos_cashier;
```
ivantodorovich, line 37: I take it wasn't possible to extend the users.loaded function, to compute this on PoS startup for all users rather than doing it every time `get_cashier` is called?

Unfortunately doing it like this is worse than what you have before :(
Because `get_cashier` is called many many times across the point of sale, and you'd be adding an unnecessary overhead every time :(

I'm not blocking on this as the overhead isn't much, just nitpicking..

I'm curious to know why the proposed solution didn't work though :sweat_smile:

ivantodorovich (approved):
Thanks for taking the time to do this :)

I'm pre approving, just one comment
Besides that, I think you need to update your tests to get travis green

Although in my opinion testing for these computed field values is unnecessary, I'm not even sure it improves coverage as the fields would be computed even without you explicitly testing them, I would just remove them tbh. A good test case for this module would be a javascript tour to actually test the access rights of each user in the point of sale ui, but I understand they're a pain in the ass to do

hkapatel-initos:
> Thanks for taking the time to do this :)
> 
> I'm pre approving, just one comment
> Besides that, I think you need to update your tests to get travis green
> 
> Although in my opinion testing for these computed field values is unnecessary, I'm not even sure it improves coverage as the fields would be computed even without you explicitly testing them, I would just remove them tbh. A good test case for this module would be a javascript tour to actually test the access rights of each user in the point of sale ui, but I understand they're a pain in the ass to do

I updated the test case

pos_access_right/static/src/js/models.js
hkapatel-initos: At least we need to check cashier has proper access rights or not , in odoo we can set employee as cashier also so i have done like that.

ivantodorovich: Yes, we need to check access right but the question is _where_.

My suggestion is to do it during PoS startup, right after the `res.users` are loaded, similarily to how [the core module computes the `role` property](https://github.com/odoo/odoo/blob/3780fa2af5d5f6cac91e419bcab69a253db280bd/addons/point_of_sale/static/src/js/models.js#L325-L343). See my comment there https://github.com/OCA/pos/pull/630#discussion_r637591752.

Alternatively, if my untested proposal doesn't work.. maybe `set_cashier` is a better place to compute this, as it's only called every once in a while

pos_access_right/static/src/css/pos.css
```diff
@@ -0,0 +1,32 @@
...
+    border: solid 1px #ddd;
+    color: white;
+    cursor: not-allowed;
+}
+
+.ticket-screen .pointer {
+    cursor: not-allowed;
+}
+
+.order-row .pointer .delete-button {
```
flotho: the icon is properly displayed.
Yet when Many orders options is disabled, there is no specific icon indicating that you don' have access to this function.
Could it be possible to use the same icon as for the line :
image

flotho (changes requested):
Except the small comment, functionnal test LGTM

pos_access_right/static/src/css/pos.css
hkapatel-initos: @flotho that suggestions are implemented you can re-check it.

PR #641, opened 2021-05-10
[13.0][FIX] pos_fix_search_limit

ivantodorovich:
`.product-list` may not be present after rendering.

This happens when the list is empty, as seen in core code:
https://github.com/odoo/odoo/blob/dcb6a3cea27f32d722d1a8dc87c34a846cf5ac23/addons/point_of_sale/static/src/xml/pos.xml#L257

The element is displayed only if `product_list.length != 0`.

---

Besides it being a rare to see bug (product list has to be empty), it causes a compatibility issue with `pos_empty_home` and when they're both installed an error will prevent the ui from rendering properly:

image

This also affects other modules tours that run in travis ci


pedrobaeza (approved):

legalsylvain (approved):
Makes sense. Thanks !

/ocabot merge patch

ivantodorovich:
can you ask the bot again? I think it doesn't work when the command in a review comment

pedrobaeza:
Indeed

/ocabot merge patch

PR #642, opened 2021-05-10
[IMP][13.0] pos_empty_home : Make it configurable and simplify code

ivantodorovich:
Hello, two commits here

- 1) Simplify code by using qweb expressions, `screens.js` and `.css` files can be removed.
- 2) Make empty home configurable in `pos.config`

The 2nd point is, specially, to be able to run test tours in order modules without this module affecting results, as it completely removes the product list when no category is selected.. which is something most tours would need.

In line with that, "Empty Home" is disabled by default in the demo `pos.config` : `point_of_sale.pos_config_main`, which is the one used by tests normally. It's otherwise enabled by default in all other `pos.config`, as this is what you'd expect when installing the module (also to avoid affecting users that already use this module).


ivantodorovich:
ping @hkapatel-initos, would you please help with a review here :)

hkapatel-initos:
> ping @hkapatel-initos, would you please help with a review here :)
Yes I will review

pos_empty_home/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    iface_empty_home = fields.Boolean(
```
hkapatel-initos, line 10: I've reviewed your PR. it is good that the code has been reduced and LGTM.

hkapatel-initos (approved):

ivantodorovich:
ping @OCA/pos-maintainers , any take on this one?
Should be an easy review and It's kind of blocking to get another PRs tests to pass

pos_empty_home/models/pos_config.py
ivantodorovich: Thanks !

legalsylvain (approved):
Makes sense. thanks for improving the CI.

/ocabot merge minor

ivantodorovich:
Thanks!

Would you please run the ocabot command again?

Unfortunately it's not triggered when written on the review comments

legalsylvain:
Weird.

/ocabot merge minor

PR #643, opened 2021-05-11
[IMP] 12.0 pos_customer_display correct proxy status display if it is the only device

PierrickBrun:
@alexis-via

### DIFF
--- a/pos_customer_display/static/src/js/models.js
+++ b/pos_customer_display/static/src/js/models.js
@@ -125,4 +125,15 @@ odoo.define('pos_customer_display.models', function (require) {
 
     });
 
+    var _config = _.findWhere(
+        models.PosModel.prototype.models,
+        {model: "pos.config"}
+    );
+    var old_loaded = _config.loaded
+    _config.loaded = function (self, configs) {
+        old_loaded(self, configs);
+        if (self.config.iface_customer_display){
+            self.config.use_proxy = true;
+        }
+    }
 });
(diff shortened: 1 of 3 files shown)


pos_customer_display/static/src/js/models.js
```diff
@@ -125,4 +125,15 @@ odoo.define('pos_customer_display.models', function (require) {
...
 
+    var _config = _.findWhere(
+        models.PosModel.prototype.models,
+        {model: "pos.config"}
+    );
+    var old_loaded = _config.loaded
+    _config.loaded = function (self, configs) {
+        old_loaded(self, configs);
+        if (self.config.iface_customer_display){
+            self.config.use_proxy = true;
```
legalsylvain, line 136: That seems very dangerous and qhoud handled in back iffice via constraint, if required

PierrickBrun: This is how Odoo does it here: https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/static/src/js/models.js#L251 

Also, the display will not function without this line (unless another device is used)

PierrickBrun: @legalsylvain Can you update your review on this ?

alexis-via (approved):
Maybe you could add a comment in the JS code to explain the purpose of this new code, because it's not obvious when you read it.

ivantodorovich (approved):

PR #646, opened 2021-05-12
[ADD] pos_hide_empty_category (12.0)

legalsylvain:
This module extends Odoo Point Of Sale features, hidding empty PoS categories.

this module can be usefull to avoid to display categories, if they contains only non saleable products
that can occurs frequently if your shop has seasonal products.

pos_category_tree_view

pos_front_ui

CC : @quentinDupont apps/deck/#/board/144/card/1472


pos_hide_empty_category/static/src/js/db.js
```diff
@@ -0,0 +1,28 @@
...
+            var self = this;
+            var child_ids = this._super(category_id);
+            var not_empty_child_ids = [];
+            _.each(child_ids, function (child_id) {
+                if (self.product_by_category_id[child_id]) {
+                    not_empty_child_ids.push(child_id);
+                }
+            });
+            return not_empty_child_ids;
+        },
```
ivantodorovich:
```suggestion
        get_category_childs_ids: function(category_id) {
            const res = this._super.apply(this, arguments)
            return _.filter(res, (categ_id) => this.product_by_category_id[categ_id])
        },
```
ES6 is already used heavily in latests versions of odoo with owl so I don't see any reason why we shouldn't do it in 12.0 :)
You achieve the same with fewer lines

pos_hide_empty_category/views/assets.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    Copyright (C) 2021 - Today: GRAP (http://www.grap.coop)
+    @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="assets_frontend" inherit_id="point_of_sale.assets">
+        <xpath expr="." position="inside">
```
ivantodorovich:
```suggestion
        <xpath expr="//script[last()]" position="after">
```
So that we OCD people are happy ;D

ivantodorovich (approved):
nitpicking here, not blocking :)

LGTM :+1:  !

pos_hide_empty_category/static/src/js/db.js
legalsylvain: > so I don't see any reason why we shouldn't do it in 12.

Ow ! the only reason is that I don't know ES6 syntax ! But I just took a look on arrow function. Looks interesting ! 

thanks for the improvement.

pos_hide_empty_category/views/assets.xml
legalsylvain: I commited your suggestion, but didn't understood this one. Could you help me understand why it is better ? (And what is OCD ?)

thanks !

pos_hide_empty_category/static/src/js/db.js
legalsylvain: Side question : any reason to not call ``const res = this._super(category_id)`` but ``const res = this._super.apply(this, arguments)``. 
Looks the first version more synthetic, but maybe I missed something.

ivantodorovich: Mostly a habit.. in vanilla js prototype inheritance it's needed to bind `this` to the method, but odoo already does it magically with its `_super` property afaik. In any case I find it's better for inter-module compatibility, because this way you call _super passing all the arguments, regardless of what they are

pos_hide_empty_category/views/assets.xml
ivantodorovich: OCD as in Obsessive compulsive disorder.. 
bad joke about liking orderliness

Technically no benefit afaik, it's just to keep scripts all together, and not mixed with stylesheets.
It's a common practice in web development to load css before js (so that the page "seems" to load faster).. but in odoo I'm not sure it makes any difference

pos_hide_empty_category/static/src/js/db.js
pedrobaeza: I think you shouldn't still use ES6 until v13, as there are supported browsers on this version that are not compatible. Can you confirm, @Tardo ?

legalsylvain: Hi @pedrobaeza. Thanks for your remark. However, I can no find the list of specific browser version supported by Odoo.

More precisely, [the documentation](https://www.odoo.com/documentation/12.0/administration/deployment/deploy.html#supported-browsers) (for V12) says basically : 
* Google Chrome
* Mozilla Firefox
* Microsoft Edge
* Apple Safari

, that all support ES6 in recent version. And the following remark seems to indicate that's OK : 

_Please make sure your browser is up-to-date and still supported by its publisher before filing a bug report._

chienandalu: I'd say the official support comes from this test: https://github.com/odoo/odoo/blob/12.0/odoo/addons/test_lint/tests/test_ecmascript.py#L13

legalsylvain: Hi @chienandalu. Thanks for pointing this code. AFAIK, it is just internal odoo lint, so the ES6 code introduced in this module will not break anything. Right ?

2 options : 

1) keep the ES6, and optionnaly add a text in the module description, to mention that this module does'nt work with old browser. 
2) keep old syntax, and change the code again, when migrating.

Option 1 has my preference. My intuition is that other web modules present in ``web`` or ``geo-spacial``  that include JS librairy will not work properly with old browser. But I'm maybe wrong, and I'm really not a JS expert, so no problem to revert the code, if asked by reviewers.

ivantodorovich: All up-to-date browsers support es6 since several years ago.

https://caniuse.com/?search=es6

Moreover all major browsers have automatic updates in place.

Unlike 5 years ago, this is no longer a debate in web development anymore. Es6 is widely used in websites and in js libraries.

Honestly I think not even a mention is required. It's 2021, 6 years have passed since es6 was "the new thing"

pedrobaeza: I have found why I remember that ES6 is not officially accepted in v12: originally, IE11 was meant to be supported by this version, and IE doesn't have support for ES6, but it was removed from the supported browsers here:

https://github.com/odoo/odoo/commit/47231e921b6f0864954d81720af4e9d233e6adcf

The reason for that is that is "out of official support from their publishers":

https://github.com/odoo/odoo/pull/58547#issuecomment-699916055

so we can go with ES6 indeed.

chienandalu: Ok, neither Odoo supports IE11 anymore officially so let's trash ES5 then :slightly_smiling_face: :+1: https://github.com/odoo/odoo/commit/47231e921b6f0864954d81720af4e9d233e6adcf

wytzejan (approved):
LGTM. Both functional and code review.

ivantodorovich:
This is ready to merge @legalsylvain ☺️

PR #470, opened 2020-03-14
[9.0] [ADD] Self service scale for loose goods and products sold by weight

Fkawala:
This add-on allows you to run a grocery in a Bring Your Own Container (BYOC) schema. With this add-on, your customers will be able to buy loose goods using their own reusable containers. This add-on also enables customers to weight their vegetables and fruits before they checkout. To weight fruit and vegetable before checkout will speed up the checkout and make your customer happier.

The **self-service** BYOC scheme has six steps:
1. The customer weighs her-his pot and sticks the tare barcode onto the pot.
2. The customer puts loose goods into the labeled pot.
3. The customer scans the tare barcode. POS saves the tare value.
4. The customer weighs the pot with loose goods inside. POS computes the price to pay from gross weight and tare weight.
5. The customer gets a price barcode label indicating the price to pay.
6. During checkout, the cashier swiftly scans the price barcode label.

This add-on requires you to have a barcode label printer and an electronic scale. Your customers are expected to use this add-on in a self-service way. This add-on adds multiple news screens that are displayed only on the selected POS.

This add-on displays only the products that are meant to be weighted with a scale.

This addon depends on: 
- point_of_sale
- pos_barcode_tare

You can test with the following tare barcode: 0700000001235 it'll tare for 0.123kg.
Demo data include an article with a weight code bar: 2112345000008.

Below are screens for a POS with this add-on enabled.

The default screen:
default_screen

The product screen:
product_screen

The tare barcode scanning screen:
tare_scan_screen

A price barcode looks like:
barcode

The configuration screen of POS has one new option:
config

PR #655, opened 2021-05-28
[12.0][ADD] pos_self_service_base

vvrossem:
This is the base module for the implementation of POS as a self-service weighing station, inspired by #470  

It depends on [pos_tare](https://github.com/OCA/pos/tree/12.0/pos_tare) to avoid duplicate barcode tare rule.

This base module handles weighing (requires you to use an electronic scale) and EAN13 barcode formatting. It is aimed to be extended by other modules, such as:
- label printing (i.e. from browser, using ZPL #656 )
- barcode scanning
- product selection

Once `iface_self_service` (Is Self-Service) is checked in `pos.config` (POS Settings), the main widget `Chrome` is split in:
- a leftpane containing the `SelfServiceScaleWidget` (the weight returned by the scale) and `self-service-control-buttons`. (`SelfServiceHomeButton`)
- a rightpane containing the `SelfServiceScreenWidget` (the Home screen with user actions such as label printing, barcode scanning, etc.). 

**Screenshots:**
_POS Settings_
pos_self_service_base-iface_self_service

_Startup screen_
pos_self_service_base

_No weight_
pos_self_service_base-no-weight

_Printing enabled_
pos_self_service_base-print-enabled

Credits:
- @leeuwerck 
- @Fkawala


pos_self_service_base/README.rst
```diff
@@ -0,0 +1,101 @@
...
+* Configure the web browser to be in kiosk mode (cf. "Launch Firefox" section)
+
+Those measures aren't sufficient *per se*, but should lower the risk significantly.
+
+Launch Firefox
+~~~~~~~~~~~~~~
+
+This self-service module can be used in kiosk mode. To launch Firefox in kiosk mode, run this command::
+
+    firefox <url> -foreground --kiosk
```
legalsylvain, line 60: what is foreground ? 
I tested without, and it is launching in kiosk mode in the same way.

pos_self_service_base/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,7 @@
+Setup the self-service POS
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
legalsylvain, line 2: nitpicking : I think the the 2 lines are unnecessary.

pos_self_service_base/readme/USAGE.rst
```diff
@@ -0,0 +1,14 @@
+This module should be used from a computer accessible to customers.
+In order to limit the risk of unfortunate/malicious actions, you should:
+
+* **Use a dedicated odoo account with the least possible rights**
+* Configure the web browser to be in kiosk mode (cf. "Launch Firefox" section)
+
```
legalsylvain, line 6: maybe add : ``* hide the keyboard and mouse once the point of sale is launched`` ?

pos_self_service_base/readme/cups
```diff
@@ -0,0 +1,52 @@
+upstream cups-reverse-proxy {
```
legalsylvain, line 1: why this file is in the readme part ?
Don't understand why it should be contained in this module.

pos_self_service_base/static/src/js/screens.js
```diff
@@ -0,0 +1,130 @@
...
+        },
+        click_print: function() {
+            if (this.scale_weight <= 0) {
+                this.gui.show_popup("alert", {
+                    title: _t("No Weight"),
+                    body: _t("Please put your container on the scale"),
+                });
+            }
+        },
+        format_barcode: function(weight) {
```
legalsylvain, line 62: I think that all the algorithm, could be moved in a separate module "pos_generate_barcode" to make this module more light, and because it is a generic feature. (that should be in the core, by the way in the module ``barcodes``, near the ``barcode_parser.js`` file)

I can make a PR for that point if you want.


some thing that could replace line 62 to line 110 by : 

```
get_barcode: function() {
  return new_function_to_do("tare", this.scale_weight * 1e3);
}

```

What do you think ?

legalsylvain (commented):
some remarks inline. (partial review) Thanks for this great work !

legalsylvain:
I wonder if tare printing should not be in a separate module, as it is an optional feature after all. ``pos_self_service_print_tare``
(one could imagine installing only the future module ``pos_self_service_print_product`` without the tare print)

Don't you think?

pos_self_service_base/static/src/js/screens.js
Fkawala: Yep, that'll be great. I can give it a try as well :)

pos_self_service_base/README.rst
vvrossem: `-foreground` used to ["Make this instance the active application."](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#-foreground). I think it is now deprecated and will be removed from `README`

pos_self_service_base/readme/CONFIGURE.rst
vvrossem: points `1.` and `2.`, right?

pos_self_service_base/readme/cups
vvrossem: Indeed, it should be in [`pos_self_service_print_zpl`](https://github.com/OCA/pos/pull/656) (or simply removed since it's in `pos_self_service_print_zpl/readme/CONFIGURE.rst`)

pos_self_service_base/static/src/js/screens.js
vvrossem: That's a great idea and gladly accept your proposition regarding the PR :)

vvrossem:
The goal of this base module is to gather common functionalities shared by modules that would extend it : 
- weighing (and visualize the weight returned by the scale)
- printing a label

I realize the button `Print container tare` is misleading. It should be instead `Print label`.

PR #657, opened 2021-05-28
[13.0][ADD] pos_partner_lang

ivantodorovich:
This module allows to edit the partner's language from the Point of Sale

image


legalsylvain (approved):
Hi. Thanks for this module. (Code review / no test)

I have a  doubt regarding the name. I guess other base field are missing. If yes i suggest to rename into pos_partner_edit_base_fields to allow to add other fields in the same module, if needed, and prevent to add a lot of modules.
Just a suggestion. Not a blocking point. (And maybe not relevant)

ivantodorovich:
> Hi. Thanks for this module. (Code review / no test)
> 
> I have a doubt regarding the name. I guess other base field are missing. If yes i suggest to rename into pos_partner_edit_base_fields to allow to add other fields in the same module, if needed, and prevent to add a lot of modules.
> Just a suggestion. Not a blocking point. (And maybe not relevant)

Hmmm I'm not sure about that.
On the project we're working on this is enough, I wouldn't add other fields apart from this one.

If a single module manages all different fields, then it would need to be configurable.. making it a little bit more complex. I'm not sure what the benefit would be there, as the module is really really small

legalsylvain:
OK. Then :+1:

Could you add the the print screen you made in the static/description folder, and a reference in the description.rst file ?

ivantodorovich:
> OK. Then 
> 
> Could you add the the print screen you made in the static/description folder, and a reference in the description.rst file ?

done ;)

hparfr (approved):

legalsylvain:
/ocabot merge patch
Thanks !

PR #114, opened 2016-08-17
[NEW] Module that links the pos order with it respective picking

sadamo:

PR #658, opened 2021-05-31
[MIG][12.0][NEW] pos_order_picking_link

mileo:
Old threads:
https://github.com/OCA/pos/pull/114
https://github.com/OCA/pos/pull/144


mileo:
@bmessiaz @ygcarvalh @luismalta @gabrielcardoso21 Could you review it?

ivantodorovich (approved):
LG 👍 

I'd add a smart button rather than a page in the picking notebook, but it's ok

mileo:
> LG 
> 
> I'd add a smart button rather than a page in the picking notebook, but it's ok

Done, thanks for the review, can you review again?

pos_order_picking_link/models/stock_picking.py
```diff
@@ -0,0 +1,31 @@
+# © 2016 KMEE INFORMATICA LTDA (https://kmee.com.br)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    pos_order_ids = fields.One2many(
```
chienandalu, line 10: How could there be more than one PoS order linked to a picking?

legalsylvain: Well, Odoo is totally fucked on that point.

One order can generate 2 pickings, so it should be the link between po and sp should be a o2m. but it is a m2o. 
https://github.com/odoo/odoo/blob/12.0/addons/point_of_sale/models/pos_order.py#L624

So, the reverse of a m2o is a o2m, so I think this code is correct. (based on an incorrect code ;-))

Otherwise, we should introduce a new field ``picking_ids`` that fixes core bad design. But I'm not sure it is possible. What do you think ?

PR #660, opened 2021-05-31
[MIG][12.0][NEW] pos_session_report

mileo:


pos_session_report/report/details_pos_session_report.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# © 2016 KMEE INFORMATICA LTDA (<http://kmee.com.br>)
+# Hendrix Costa <hendrix.costa@kmee.com.br>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp.report import report_sxw
```
rvalyi, line 6: this class doesn't exist anymore in Odoo 12+. Dead code?

rvalyi (changes requested):

pos_session_report/report/details_pos_session_report.py
mileo: I'm just doing the "git am" of version 8 modules for now.

PR #663, opened 2021-06-03
[14.0] [mig] pos_margin

dsolanki-initos:


fshah-initos (approved):

dsolanki-initos:
It would be great if we get the reviews here

dsolanki-initos:
@quentinDupont It would be nice if you can add reviews for this PR

pos_margin/models/pos_order_line.py
```diff
@@ -33,17 +29,34 @@ def _compute_multi_margin(self):
...
-        # computation that handles multi currencies context and UoM
-        SaleOrderLine = self.env["sale.order.line"]
-
-        # if used in combination with module which does add the uom_id to line
-        uom = hasattr(line, "uom_id") and line.uom_id or line.product_id.uom_id
-
-        return SaleOrderLine._get_purchase_price(
-            line.order_id.pricelist_id, line.product_id, uom, line.order_id.date_order
-        )["purchase_price"]
+        if not line.product_id:
```
legalsylvain, line 39: Hi, as the following lines are a copy past from odoo code, could you : 
- add a comment with a link to the copied code (https://github.com/odoo/odoo/blob/14.0/addons/sale_margin/models/sale_order.py#L20)
- add credits to Odoo for this file ?
-

legalsylvain (changes requested):
Hi @dsolanki-initos. 

thanks for porting this module. Some remarks inline. 

could you also cherry pick latest commit from 12.0 ? here https://github.com/OCA/pos/pull/673

thanks !

sama62 (commented):
Hi,
there is a problem here. for runbot  , we are in version 12 instead of 14 .
https://runbot.odoo-community.org/runbot/build/3489683
image

dsolanki-initos:
> Hi,
> there is a problem here. for runbot , we are in version 12 instead of 14 .
> https://runbot.odoo-community.org/runbot/build/3489683
> image

I dont why but may be because I've cherry the latest commit from v12  as per the above comment.

ivantodorovich:
No, it's ok.

What's shown there, `pos_margin 12.0.2.0.1`, is the name of your latest commit (https://github.com/OCA/pos/pull/663/commits/4c94707cd121c0a4766bc6d187c5b58ec7bc17aa) not the version runbot is running.


I believe that's the commit you cherry-picked from @legalsylvain. Btw, can you please reorder them in a way that makes sense (it should be somewhere between migration to 12.0 and migration to 13.0 commits)

ivantodorovich:
Also it seems there are two more commits to cherry-pick:

- https://github.com/OCA/pos/commit/d97f96c86e31a307c28066504fc6c99486ad46ad
- https://github.com/OCA/pos/commit/9bb3c18e5dfbfe148858a405a835ffdc0b0fc494

sama62:
> > Hi,
> > there is a problem here. for runbot , we are in version 12 instead of 14 .
> > https://runbot.odoo-community.org/runbot/build/3489683
> > image
> 
> I dont why but may be because I've cherry the latest commit from v12 as per the above comment.

Ok, but could you tell me please the best way to do a test scenario in version 14 ? it means that is there any other way to do a test? thanks in advance 
also notify @gurneyalex

ivantodorovich:
> Ok, but could you tell me please the best way to do a test scenario in version 14 ? it means that is there any other way to do a test? thanks in advance

What do you mean?

To perform a functional test, you can use runbot. You simply click on that button and log into the database with `admin` / `admin`

image

sama62:
> > Ok, but could you tell me please the best way to do a test scenario in version 14 ? it means that is there any other way to do a test? thanks in advance
> 
> What do you mean?
> 
> To perform a functional test, you can use runbot. You simply click on that button and log into the database with `admin` / `admin`
> 
> image
Thank you for your quick answer. I know how to test in Runbot. I saw yesterday that its version, not only in its name but also in the configuration and system information, was 12. Today it's 14, so I'll test it again.

flotho:
> > Ok, but could you tell me please the best way to do a test scenario in version 14 ? it means that is there any other way to do a test? thanks in advance
> 
> What do you mean?
> 
> To perform a functional test, you can use runbot. You simply click on that button and log into the database with `admin` / `admin`
> 
> image

the db produced don't seems to be an odoo 14 valid but an odoo 12 valid!

ivantodorovich:
> the db produced don't seems to be an odoo 14 valid but an odoo 12 valid!

It looks 14.0 to me :thinking: 

http://3489683-663-4c9470.runbot1.odoo-community.org/
image


Could it be that you're logging into / being redirected to a different database?

flotho:
> > the db produced don't seems to be an odoo 14 valid but an odoo 12 valid!
> 
> It looks 14.0 to me thinking
> 
> http://3489683-663-4c9470.runbot1.odoo-community.org/
> image
> 
> Could it be that you're logging into / being redirected to a different database?

good news, we actually have been redirected to a v12 db. don't really know why, we'll test this soon, thanks for the feedback

pos_margin/models/pos_order_line.py
ivantodorovich: Correct me if I'm wrong, but can't we drop the dependency to `sale_margin` now that the method has been copied here? IF so also update `ROADMAP.rst`

ivantodorovich (changes requested):
Also check pre-commit pls :)

dsolanki-initos:
I've removed the sale_margin dependency.

pos_margin/static/src/js/models.js
```diff
@@ -0,0 +1,58 @@
...
+    // /////////////////////////////
+    // Overload models.Order
+    // /////////////////////////////
+    var OrderMargin = models.Order.extend({
+        get_margin: function () {
+            return this.get_orderlines().reduce(
+                (margin, line) => (margin += line.get_margin()),
+                0
+            );
+            return margin;
```
ivantodorovich:
```suggestion
```

ivantodorovich (approved):
LGTM.

Just this that's causing pre-commit to fail

ivantodorovich:
/ocabot merge nobump

Thanks @dsolanki-initos :partying_face:

PR #664, opened 2021-06-03
[14.0][mig]pos_warning_exiting

hkapatel-initos:


legalsylvain (changes requested):
Hi. Thanks for porting this module !

This seems WIP. you added a new full file ``Chrome.js`` that is not included in the loaded files.

Note : could you rename into ``chrome.js`` BTW ?

hkapatel-initos:
> Hi. Thanks for porting this module !
> 
> This seems WIP. you added a new full file `Chrome.js` that is not included in the loaded files.
> 
> Note : could you rename into `chrome.js` BTW ?

yes, I have renamed the file.

dsolanki-initos (approved):
Looks Good To Me

hkapatel-initos:
@ivantodorovich can you please review this PR?

pos_warning_exiting/__manifest__.py
```diff
@@ -6,17 +6,13 @@
     "name": "Point Of Sale - Warning on Exiting",
     "summary": "Add warning at exiting the PoS front office UI"
     " if there are pending draft orders",
-    "version": "12.0.1.0.1",
+    "version": "14.0.1.0.1",
```
ivantodorovich: Version should be `14.0.1.0.0`

pos_warning_exiting/static/src/js/chrome.js
```diff
@@ -0,0 +1,87 @@
+odoo.define("pos_warning_exiting.Chrome", function (require) {
+    "use strict";
+
+    const {_t} = require("web.core");
```
ivantodorovich: You can just use `this.env._t` and avoid this requirement

Not blocking though

pos_warning_exiting/static/src/js/chrome.js
```diff
@@ -0,0 +1,87 @@
...
+                    });
+                }
+
+                if (draft_orders_description.length) {
+                    const reason = this.env._t(
+                        "You have some draft unpaid orders." +
+                            " You can exit temporarily the Point of Sale, but you" +
+                            " will loose that orders if you close the session: " +
+                            draft_orders_description.join(", ")
+                    );
```
ivantodorovich: Blocking.

This won't ever be translated because you're including `draft_orders_description` into the translatable string.

You need to do something like this:
```js
const message = this.env._t(
    "You have some draft unpaid orders. "
    + "You can exit temporarily the Point of Sale, but you "
    + "will loose that orders if you close the session: "
);
const reason = message + draft_orders_description.join(", ");
```

pos_warning_exiting/static/src/js/chrome.js
```diff
@@ -0,0 +1,87 @@
...
+                            // Lazy depedency to pos_restaurant
+                            // to display table information
+                            if (order_json.table) {
+                                order_description +=
+                                    _t(" - Table: ") + order_json.table.name;
+                            }
+                            draft_orders_description.push(order_description);
+                        }
+                    });
+                }
```
ivantodorovich: Not blocking here but I would do something like this here

```js
const unpaidOrders = this.env.pos.db.get_unpaid_orders();
const toLoseOrders = unpaidOrders.filter(
    order => order.pos_session_id === this.env.pos.session_id.id && order.lines.length
);
if (toLoseOrders.length) {
    const orderDescriptions = toLoseOrders.map(order => {
        let description = this.env._t("Order #") + order.sequence_number;
        // Lazy dependency to pos_restaurant to display table information
        if (order.table) {
            description += this.env._t(" - Table: ") + order.table.name;
        }
        return description;
    })
    // .. you know the rest
}
```

by using arrow functions you avoid having to define and use `self`, and you achieve a more readable result IMO

ivantodorovich (changes requested):

hkapatel-initos:
@ivantodorovich Thanks for the review  suggestions are implemented you can re-review it

ivantodorovich (approved):
LGTM :+1: 

Thanks a lot for taking the time to do this

hkapatel-initos:
@HaraldPanten can u please review this pr?

HaraldPanten (approved):
Functional review 👍

hkapatel-initos:
@brendapaniagua can u please review this pr?

hkapatel-initos:
> Hi. Thanks for porting this module !
> 
> This seems WIP. you added a new full file `Chrome.js` that is not included in the loaded files.
> 
> Note : could you rename into `chrome.js` BTW ?

Thanks for the review suggestions are implemented you can re-review it

PR #667, opened 2021-06-03
[13.0][ADD] pos_event_sale

ivantodorovich:
Superseeds: https://github.com/OCA/event/pull/215

add_event_button

event_calendar

ticket_selector


pos_event_sale/models/pos_config.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 Camptocamp SA - Iván Todorovich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    iface_event_sale = fields.Boolean("Sell Events")
```
legalsylvain: maybe a default to True ? we can consider that if people are installing the module, they want the feature ? (as other pos modules...)

pos_event_sale/models/pos_order.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2021 Camptocamp SA - Iván Todorovich
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class PosOrder(models.Model):
+    _inherit = "pos.order"
```
legalsylvain, line 8: Do we have to handle the refund somewhere in that file ? I guess that refunding a pos.order should cancel the event.registration ? Don't you think ? 

(if not in the scope of your work, maybe add something in the roadmap)

pos_event_sale/models/pos_order_line.py
```diff
@@ -0,0 +1,49 @@
...
+        res._create_event_registrations()
+        return res
+
+    def _create_event_registrations(self):
+        for line in self.filtered("event_ticket_id"):
+            ticket_id = line.event_ticket_id
+            vals = {
+                "event_ticket_id": ticket_id.id,
+                "event_id": ticket_id.event_id.id,
+                "partner_id": line.order_id.partner_id.id,
```
legalsylvain, line 39: If I want to buy 2 tickets for two people in the same order, I guess it will not possible because it seems you use the partner_id of the order_id. (or maybe I missed something, I don't use ``event`` module).

legalsylvain (commented):
very superficial review. a question regarding partner, the rest is minor.

LGTM. Thanks !

pos_event_sale/models/pos_config.py
ivantodorovich: I pushed the changes. Although I agree with you, this can have undersired effects during tests if we consider this to be the norm.. (see https://github.com/OCA/pos/pull/642 for example)

This module is only adding to core features, not reducing them.. so it shouldn't be an issue

pos_event_sale/models/pos_order.py
ivantodorovich: Added to roadmap

pos_event_sale/models/pos_order_line.py
ivantodorovich: That's correct. In fact the registration details are not handled currently by this module, as it wasn't in the scope of our project.
In my experience this is not needed most of the times, but I added a comment in the ROADMAP.

olgamarcocb (approved):
LGTM 👍

didierdonze (approved):
Functional review: ok

PR #673, opened 2021-06-10
[IMP][12.0] pos_margin: display margin on front end

legalsylvain:
* main feature : Add margin and margin rate on the PoS front end.

image

* Make the summary homogeneous when other OCA modules are installed and fix inheritance (``pos_order_mgmt`` and ``pos_meal_voucher``)

CC : @quentinDupont, @ivantodorovich 

Ref : apps/deck/#/board/144/card/1628


pos_margin/static/src/js/models.js
```diff
@@ -0,0 +1,56 @@
...
+    // /////////////////////////////
+    // Overload models.Order
+    // /////////////////////////////
+    var OrderMargin = models.Order.extend({
+
+        get_margin: function() {
+            var margin = 0;
+            this.orderlines.each(function(orderline){
+                margin += orderline.get_margin();
+            })
```
ivantodorovich:
```suggestion
            return this.get_orderlines().reduce(
                (margin, line) => margin += line.get_margin(), 0
            )
```

This achieves the same
not blocking though

pos_margin/static/src/js/models.js
```diff
@@ -0,0 +1,56 @@
...
+        },
+
+        get_margin_rate: function(){
+            var priceWithoutTax = this.get_total_without_tax();
+            if (priceWithoutTax) {
+                return ((this.get_margin()) / priceWithoutTax) * 100;
+            }
+            else {
+                return 0;
+            }
```
ivantodorovich:
```suggestion
            return priceWithoutTax ? ((this.get_margin()) / priceWithoutTax) * 100 : 0;
```
also not blocking

pos_margin/static/src/js/screens.js
```diff
@@ -0,0 +1,22 @@
...
+    "use strict";
+
+    var screens = require("point_of_sale.screens");
+
+    screens.OrderWidget.include({
+        update_summary: function () {
+            this._super.apply(this, arguments);
+            var order = this.pos.get_order();
+            if (!order.get_orderlines().length) {
+                return;
```
ivantodorovich, line 15: I haven't tested, but isn't it necessary to clear the summary in case you remove all order lines?

ivantodorovich (approved):
Minor not blocking comments

I would make this new feature configurable through the `pos.config`, and disabled by default.

I'm thinking current users of this module might not want to display the margin to their cashiers, where customers can potentially see it.

pos_margin/static/src/js/screens.js
legalsylvain: no, the summary is hidden if nolines.

length === 0

image

length !== 0

image

legalsylvain:
1) Thanks for your review !

2) thanks for the JS suggestion, I included it in the PR.
 
3)
> I would make this new feature configurable through the pos.config, and disabled by default. I'm thinking current users of this module might not want to display the margin to their cashiers, where customers can potentially see it.

OK to make it configurable, you're right depending on the configuration of the shop, the values could be visible by the customer, and in some cases, this could be embarrassing.

- But, I'd like to set it enabled by default because in this way, all the features are enabled and admin can disable some ones, if required. Otherwise admin could not know that the feature is available and could miss this feature.
- I think that not only code should be open, margin should also be Open Source. In grap, we practice price transparency. (but that is political choice) I know that in food coops, that is the same. (Park Slope Food Coop at New York, La Louve at Paris, etc..)

If it's a problem for existing instance, I can simply add a migration script to disable the feature.

could be OK ? 

Note : Context of the development : users don't allways set correctly the standard_price on each product form. So the margin is bad computed. It is more a new feature to check that your products are correctly set and fix quickly bad values.

ivantodorovich:
Personally I'm not using this module, so I'm not blocking over this, but I'm thinking on the people that have this module installed, decided to update it and unknowingly started to display cost & margin in the Point of Sale.

A migration script to disable the feature in existing installations would cover this case, if you want it enabled by default.

legalsylvain:
Hi @ivantodorovich 

- I added a configuration option in the ``pos.config``

pos_config_form


- I added a migration script to disable that feature, for existing instance
- I udpated the description and add ``CONFIGURE.rst`` section
- I added the margin rate on each line below the price (see updated print screen)


**merge should be done with nobump option**

quentinDupont (approved):

PR #678, opened 2021-06-22
 [FIX] pos_payment_terminal : order is undefined in the table screen

legalsylvain:
Use case : 
- install ``pos_payment_terminal`` and ``pos_restaurant``
- configure your pos.config with one (or more floor)
- launch the Point of sale 
- a non blocking JS error is raised, once the PoS is loaded

image

apps/deck/#/board/144/card/1687

CC : @quentinDupont


ivantodorovich (approved):
LGTM :+1:

legalsylvain:
Patch used in production since 6 monthes.
/ocabot merge patch

legalsylvain:
Two days for a merge ... retrying...

 /ocabot merge patch

PR #414, opened 2019-11-19
12.0 mig pos loyalty

wpichler:
#344 is the original PR - was closed because author is not going to work on it anymore. This is the new PR for it - at time no changes from the original PR.
Odoo 13 CE does not seem to have pos_loyalty in CE - so no backport is possible

PR #681, opened 2021-06-25
12.0 mig pos loyalty

hitrosol:
continuing of #414 since it's abandoned by the author...


dsolanki-initos:
@hitrosol  Can you make Travis happy and change as per #414 comments.

hitrosol:
I try to fixed it but no success, can anyone help on this ? IMO this module is important and needed not only in V12, but also in V13 and V14.

PR #682, opened 2021-06-25
[14.0] [mig] pos_ticket_without_price

dsolanki-initos:


dsolanki-initos:
Hi @dalonsod 
It would be nice if you can review this PR

legalsylvain (changes requested):
Hi. 

Thanks for porting this module !

Functional test : blocking point

It seems that the price is displayed if the quantity is one. 

See : 

image

Code review : suggestion.

I guess that it could great to add a simple button option for the without price, named "without price)" if checked, overload the current template to hide prices. (instead of creating a new template).

like the current "Invoice" implementation.

unchecked : 
image
checked : 
image

That way, all the other modules that adds extra data on the bill will work. like (for exemple) ``pos_environment``. (https://github.com/OCA/server-env/tree/12.0/pos_environment)

Just a suggestion, as the current module is design with two templates.

pos_ticket_without_price/static/src/js/screens.js
```diff
@@ -0,0 +1,44 @@
+/* Copyright 2021 Initos Gmbh
+   License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+*/
+
+odoo.define("pos_ticket_without_price.screens", function (require) {
```
ivantodorovich:
```suggestion
odoo.define("pos_ticket_without_price.ReceiptScreen", function (require) {
```

Please also rename the file and the assets entry

pos_ticket_without_price/static/src/js/receiptwithoutprice.js
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2021 Initos Gmbh
+   License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+*/
+odoo.define("point_of_sale.OrderReceiptWithoutPrice", function (require) {
```
ivantodorovich: Please rename this file to `OrderReceiptWithoutPrice.js` and also update it's assets entry

ivantodorovich (changes requested):

ivantodorovich (approved):
pre-approving assuming @legalsylvain's comments are resolved :)

dsolanki-initos:
> Hi.
> 
> Thanks for porting this module !
> 
> Functional test : blocking point
> 
> It seems that the price is displayed if the quantity is one.
> 
> See :
> 
> image
> 
> Code review : suggestion.
> 
> I guess that it could great to add a simple button option for the without price, named "without price)" if checked, overload the current template to hide prices. (instead of creating a new template).
> 
> like the current "Invoice" implementation.
> 
> unchecked :
> image
> checked :
> image
> 
> That way, all the other modules that adds extra data on the bill will work. like (for exemple) `pos_environment`. (https://github.com/OCA/server-env/tree/12.0/pos_environment)
> 
> Just a suggestion, as the current module is design with two templates.

Thank you for the suggestion. I think it will change the original functionality. We should keep it as it is, as it is already implemented way it was.

dalonsod (approved):
@dsolanki-initos tested on runbot 👍

dsolanki-initos:
It would be nice if the maintainers can merge this PR

ivantodorovich:
Hello!

What about  @legalsylvain 's blocking point:

> Functional test : blocking point
>
>It seems that the price is displayed if the quantity is one.
>
>See:
>
>image

Is it solved?

dsolanki-initos:
> Hello!
> 
> What about @legalsylvain 's blocking point:
> 
> > Functional test : blocking point
> > It seems that the price is displayed if the quantity is one.
> > See:
> > image
> 
> Is it solved?

Yes, it is resolved you can check it.

legalsylvain:
> Thank you for the suggestion. I think it will change the original functionality.

Indeed, but in the good way ! for the time being, this module add a new template, that is totally disconnected of the rest of Odoo. So, no possibility to have other information, added by other modules, no possibility to send by email a ticket without price, that could be a desired feature for all the people who are making a gift to some people and who want to transmit the proof of purchase, without the price.

> We should keep it as it is, as it is already implemented way it was.

I understand that it is not in the scope of the PR, but we should add a text in the ROADMAP.rst file.

> Yes, it is resolved you can check it.

Checked. Thanks !

hkapatel-initos:
Hello maintainers,
It would be nice if you are trying to merge it

hkapatel-initos:
it would be nice if the reviewer gave approval to this pull.

ivantodorovich:
Can you take care of this comment? 

> > We should keep it as it is, as it is already implemented way it was.
>
> I understand that it is not in the scope of the PR, but we should add a text in the ROADMAP.rst file.


Then it should be ready to merge AFAICS

brendapaniagua (approved):
Functional review! LGTM!

PR #684, opened 2021-06-29
[ADD] pos_payment_method_cashdro: New module

chienandalu:
This module allows to make payments with a Cashdro (https://www.cashdro.com) terminal
on the Point of Sale frontend.
cashdro

cc @Tecnativa TT29224

please review @pedrobaeza @joao-p-marques


pos_payment_method_cashdro/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,9 @@
+To configure this module, you need to:
+
+#. Go to a Point of Sale cash payment method or create a new one.
+#. In the Point of Sale tab, set *Cashdro Payment Terminal* on.
+#. Configure the Cashdro terminal hostname and credentials.
+#. Configure the desired, Cashdro terminal in the proper PoS configurations.
```
ivantodorovich, line 6:
```suggestion
#. Configure the desired Cashdro terminal in the proper PoS configuration.
```

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+
+    var screens = require("point_of_sale.screens");
+    var core = require("web.core");
+
+
+    screens.PaymentScreenWidget.include({
+
+        render_paymentlines : function(){
+            this._super.apply(this, arguments);
+            var _this = this;
```
ivantodorovich, line 12: It's kind of weird to see `_this` as a reference to `this` in js.
The "community standard" would be `self`, which is also more readable due to the lack of underscore

```suggestion
            const self = this;
```

---

Also, not related to only this line, you might want to consider using `const` for pretty much all `var` usages
- [JavaScript ES6+: var, let, or const?](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75)

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+                        _this.render_paymentlines();
+                        _this.$('.paymentline.selected .edit').text(_this.format_currency_no_symbol(amount));
+                    });
+            }).on(
+                "click", ".cashdro-transaction-cancel",
+                function (ev) {
+                _this.pos.get_order().in_cashdro_transaction = false;
+                _this.order_changes();
+                _this.journal = _this._get_journal($(this).data('cid'));
+            	_this.cashdro_finish_operation(_this.pos.get_order().cashdro_operation);
```
ivantodorovich, line 32: Seems there are mixed tabs / spaces indents here and some other places too

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+    var screens = require("point_of_sale.screens");
+    var core = require("web.core");
+
+
+    screens.PaymentScreenWidget.include({
+
+        render_paymentlines : function(){
+            this._super.apply(this, arguments);
+            var _this = this;
+            this.$(".paymentlines-container").unbind("click").on(
```
ivantodorovich, line 13: I'm unable to test, but are you sure that `unbind` is required here and/or not doing more than expected?

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+        cashdro_send_payment_request: function () {
+            // The payment is done in three concatenated steps:
+            // 1. The POS send a payment request, to which the Cashdro respondes
+            //    with an operation id.
+            // 2. Then the POS has to acknowledge that such operation id has
+            //    been received.
+            // 3. Once acknowledged the POS has to send a payment request to the
+            //    cashdro. Once the Cashdro responses with a "F" state (for
+            //    finished) we'll get the response and fill the tendered money
+            //    for the payment line.
```
ivantodorovich, line 64:
```suggestion
        /**
         * Send Cashdro payment request
         *
         * The payment is done in three concatenated steps:
         * 1. The POS send a payment request, to which the Cashdro respondes
         *    with an operation id.
         * 2. Then the POS has to acknowledge that such operation id has
         *    been received.
         * 3. Once acknowledged the POS has to send a payment request to the
         *    cashdro. Once the Cashdro responses with a "F" state (for
         *    finished) we'll get the response and fill the tendered money
         *    for the payment line.
         * 
         * @returns {Deferred}
         */
        cashdro_send_payment_request: function () {
```

nitpicking: jsdoc

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+
+        _cashdro_payment_url: function(parameters) {
+            // Compose the url for a sale report to Cashdro
+            var url = this._cashdro_url()
+            url += "&operation=startOperation&type=4";
+            url += "&posid=pos-" + this.pos.pos_session.name;
+            url += "&posuser=" + this.pos.get_cashier().id;
+            url += "&parameters=" + JSON.stringify(parameters);
+            return url;
+        },
```
ivantodorovich, line 142: Just a suggestion: [$.param](https://api.jquery.com/jquery.param/) for readability
Could do something like this, for example

```suggestion
        /**
         * Compose the url for a sale report to Cashdro
         * @param {Object} parameters
         * @returns {String}
         */
        _cashdro_payment_url: function(parameters) {
            return [
                this._cashdro_url(),
                $.param({
                    operation: "startOperation",
                    type: 4,
                    posid: "pos-" + this.pos.pos_session.name,
                    posuser: this.pos.get_cashier().id,
                    parameters: JSON.stringify(parameters),
                }),
            ].join("&")
        },
```

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+
+        _show_error: function (msg, title) {
+            if (!title) {
+                title =  _t("CashDro Error");
+            }
+            this.pos.gui.show_popup("error", {
+                "title": title,
+                "body": msg,
+            });
+        },
```
ivantodorovich, line 212:
```suggestion
```

Is this method used?
If yes, maybe remove by simply calling `show_popup`, or at least rename it because `_show_error` is too generic and that's not playing nicely with other modules

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+            var order = this.pos.get_order();
+            var lines = order.get_paymentlines();
+            for ( var i = 0; i < lines.length; i++ ) {
+                if (lines[i].cid === line_cid) {
+                    line = lines[i];
+                }
+            }
+            return line.cashregister.journal
+        },
+        _cancel_current_transaction_popup: function () {
```
ivantodorovich, line 227:
```suggestion
        _cashdro_cancel_current_transaction_popup: function () {
```
It seems this method isn't used neither.
But I'm guessing the intention is to be used in `cashdro-transaction-cancel` ?
In any case, it's better to have a more specific name

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+                },
+                cancel: function () {
+                    return;
+                }
+            });
+        },
+
+    });
+
+});
```
ivantodorovich, line 243:
```suggestion
    return screens;
});
```

This helps other modules potentially extend this one

pos_payment_method_cashdro/static/src/js/models.js
```diff
@@ -0,0 +1,20 @@
...
+    var order_super = models.Order.prototype;
+
+    models.Order = models.Order.extend({
+        initialize: function(){
+            order_super.initialize.apply(this, arguments);
+            this.in_cashdro_transaction = false;
+        },
+    });
+
+});
```
ivantodorovich, line 20:
```suggestion
    return models;
});
```

This helps other modules potentially extend this one

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+                .then(function (res_ack) {
+                    // Validate the operation
+                    console.log(res_ack);
+                    var ask_url = _this._cashdro_ask_url(operation_id);
+                    return _this._cashdro_request_payment(ask_url);
+                })
+                .then(function (operation_data) {
+                    // This might be too verbose, but it helps a lot to diagnose
+                    // issues and their reasons.
+                    console.log(operation_data);
```
ivantodorovich, line 94: If you want to keep these, at least do it only in debug mode.
It would help to add a descriptive string too

```suggestion
                    // This might be too verbose, but it helps a lot to diagnose
                    // issues and their reasons.
                    if (window.odoo.debug) {
                        console.log("Cashdro operation data", operation_data);
                    }
```

pos_payment_method_cashdro/static/src/scss/pos_cashdro_terminal.scss
```diff
@@ -0,0 +1,7 @@
+button.cashdro-transaction-start,  button.cashdro-transaction-cancel {
+    /*Same css values as ".screen .top-content .button" */
+    font-size: 20px;
+    padding: 3px 13px;
+    line-height: 32px;
+    border-radius: 3px;
+}
```
ivantodorovich, line 7:
```suggestion
.pos {
    button.cashdro-transaction-start,  button.cashdro-transaction-cancel {
        /*Same css values as ".screen .top-content .button" */
        font-size: 20px;
        padding: 3px 13px;
        line-height: 32px;
        border-radius: 3px;
    }
}
```

It'd be even better to make it more restrictive, but I don't know the widget's css classes by heart

pos_payment_method_cashdro/static/src/xml/pos_cashdro_terminal.xml
```diff
@@ -0,0 +1,27 @@
...
+    <t t-extend="PaymentScreen-Paymentlines" >
+        <t t-jquery=".col-name" t-operation="append">
+                <t t-if="line.cashregister.journal.cashdro_payment_terminal">
+                    <!-- Stop transaction button -->
+                    <button class="payment-terminal-transaction cashdro-transaction-cancel oe_hidden" t-att-data-cid="line.cid"
+                            title="Cancel transaction">
+                        <i class="fa fa-ban" />
+                    </button>
+                    <!-- Request payment button -->
+                        <button class="payment-terminal-transaction cashdro-transaction-start" t-att-data-cid="line.cid"
```
ivantodorovich, line 20: Check indents

pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -0,0 +1,243 @@
...
+            var self = this;
+            this.gui.show_popup("confirm",{
+                title: _t("Transaction ongoing"),
+                body:  _t("There is a transaction in progress, would you like to cancel it?"),
+                confirm: function () {
+                    self._cancel_current_transaction()
+                },
+                cancel: function () {
+                    return;
+                }
```
ivantodorovich, line 237:
```suggestion
```
I don't think this is necessary, is it?

ivantodorovich (changes requested):
Thanks a lot !
It's not possible to test without the actual terminal, but I've just done a code review

-----

Even though it doesn't depend on `pos_payment_terminal`, maybe consider renaming the module to `pos_payment_terminal_cashdro` ?

Seeing that not every payment terminal requires an iotbox, we might need to consider renaming `pos_payment_terminal` to something else in the future

joao-p-marques (approved):
Code review :+1:

pos_payment_method_cashdro/readme/CONFIGURE.rst
chienandalu: Well, it can be plural, as one terminal could be used by serveral pos configs

pos_payment_method_cashdro/static/src/js/screens.js
chienandalu: Well, `self` is indeed an actual const for `window` in [JS](https://developer.mozilla.org/en-US/docs/Web/API/Window/self), so `_this` is a usual convention to keep the pointer of the previous `this` scope.

> Also, not related to only this line, you might want to consider using const for pretty much all var usages

Well, I was developing this with ES5 standard before the recent commitment to drop that constraint... so I will just keep it like that for this version, and will move on to moder JS in v13 :slightly_smiling_face:

pos_payment_method_cashdro/static/src/js/screens.js
chienandalu: I'll pass prettier :)

pos_payment_method_cashdro/static/src/js/screens.js
chienandalu: Well, here I just copied from `pos_payment_terminal` :sweat_smile: so I'm not completely sure...

pos_payment_method_cashdro/static/src/js/screens.js
chienandalu: Nice!

pos_payment_method_cashdro/static/src/js/screens.js
chienandalu: Sure, they're some missing remanents when I tried to adapt it to a backport of some v13 features. It's to be dropped

FernandoRomera (approved):

FernandoRomera:
@ivantodorovich Can you remove the request and merge this PR?.

PR #691, opened 2021-07-26
Automatic Stale PR/Issues workflow

ivantodorovich:
ping @OCA/pos-maintainers 

Should we try this here?

To discuss policies, there's an ongoing discussion here: https://github.com/OCA/oca-addons-repo-template/pull/63

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,68 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+    - cron: "0 12 * * 0"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      # 120+30 day stale policy for PRs
+      # * Except PRs marked as "no stale"
+      - name: Stale PRs policy
+        uses: actions/stale@v4
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-pr-labels: "no stale"
+          days-before-stale: 120
+          days-before-close: 30
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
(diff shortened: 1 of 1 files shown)


legalsylvain:
Hi @ivantodorovich. Thanks for this discussion. I have no clear point of view. yes, it should be great to clean obsolete PR / issues.

But in the meantime, I think that 120 days is not so much ! Maybe it requires to organize some review session, something like 4 time a year. What do you think ?

Technical question : your PR is against 14.0. When we'll migrate to 15.0 do you have to move the github action settings on V15. keep the V14 code ? both code will be executed ?

ivantodorovich:
> But in the meantime, I think that 120 days is not so much ! Maybe it requires to organize some review session, something like 4 time a year. What do you think ?

Let me first do a quick summary of how it works, in case people don't know.

The policy is in fact 120+30. This means: if there's no activity in 120 days the action will label it as stale, and post a message.

If it remains without activity for 30 more days, it closes it. Please note that a simple comment from anyone counts as activity, so people interested in the PR to remain open have plenty of time to do so. Also the PSC members can use a label "no stale" to keep it open forever.

What we'd expect from this is to have an automated system that forces and remainds us to qualify PRs.. and that closes them automatically if there's no public interest in them anymore

I recommend you join the discussion in the other thread to talk and brainstorm about possible policies ☺️ your experience as maintainer is appreciated

> Technical question : your PR is against 14.0. When we'll migrate to 15.0 do you have to move the github action settings on V15. keep the V14 code ? both code will be executed ?

For scheduled actions like this one, only the workflow in the repository's default branch is executed. So yes we need to copy onto the new branches as we create them (although copier takes care of it)

PR #703, opened 2021-08-26
[14.0][FIX] pos_cash_move_reason: do not load account.account from demo data

Cedric-Pigeon:
Loading account creation from demo data could have side effect when a chart of account is loaded afterwards.
See: https://github.com/odoo/odoo/blob/14.0/addons/account/models/chart_template.py#L205

### DIFF
--- a/pos_cash_move_reason/tests/test_pos_cash_move_reason.py
+++ b/pos_cash_move_reason/tests/test_pos_cash_move_reason.py
@@ -1,14 +1,30 @@
 # © 2015 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from odoo import tools
 from odoo.exceptions import UserError
+from odoo.modules.module import get_module_resource
 from odoo.tests.common import Form, SavepointCase
+
+
+def load_file(cr, module, *args):
+    tools.convert_file(
+        cr,
+        "pos_cash_move_reason",
+        get_module_resource(module, *args),
+        {},
+        "init",
+        False,
+        "test",
+    )
 
 
 class TestPosCashMoveReason(SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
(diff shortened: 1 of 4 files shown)


pos_cash_move_reason/tests/test_pos_cash_move_reason.py
```diff
@@ -1,14 +1,30 @@
 # © 2015 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from odoo import tools
 from odoo.exceptions import UserError
+from odoo.modules.module import get_module_resource
 from odoo.tests.common import Form, SavepointCase
 
 
+def load_file(cr, module, *args):
```
legalsylvain, line 10: There is no core function that do that ?
Just a question...
Otherwise lgtm.

legalsylvain (approved):

pos_cash_move_reason/tests/test_pos_cash_move_reason.py
Cedric-Pigeon: I do not think so

ivantodorovich (approved):

ivantodorovich (changes requested):
Actually, can you check?
I think the `demo` files are still there

ivantodorovich:
@Cedric-Pigeon could you remove these files, as they should've been moved instead:

https://github.com/acsone/pos/tree/14.0-pos_reason-cpi/pos_cash_move_reason/demo

pos_cash_move_reason/tests/test_pos_cash_move_reason.py
rousseldenis: Maybe we can use this : https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L255

But not the same 'context' as we are in tests here

elvise:
> @Cedric-Pigeon could you remove these files, as they should've been moved instead:
> 
> https://github.com/acsone/pos/tree/14.0-pos_reason-cpi/pos_cash_move_reason/demo

@Cedric-Pigeon can you check this point ?

PR #708, opened 2021-09-16
[13.0][FIX] pos_payment_terminal: Graceful cancel handling

thomaspaulb:
Because `send_payment_cancel` was not implemented and the original Odoo version does not return a promise at all, I got the following error on payment cancel:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'finally') [point_of_sale.assets.js:268]
```

Not only does this error look funny in the Javascript console, but I also got problems with black screen when running POS in debug mode, because apparently once a payment is cancelled but it's not handled with a promise, it retries on page load, causing a Javascript error on page load.

While of course the best course of action is to implement the actual cancel flow, this was out of scope for the moment and so I just display a popup telling the user to cancel the payment with the red button - and then set the promise as resolved.

### DIFF
--- a/pos_payment_terminal/static/src/js/payment_terminal.js
+++ b/pos_payment_terminal/static/src/js/payment_terminal.js
@@ -25,12 +25,12 @@ odoo.define("pos_payment_terminal.payment", function(require) {
 
         send_payment_cancel: function() {
             this._super.apply(this, arguments);
-            this._show_tip(
+            this._show_error(
                 _t(
-                    "Cancel the transaction by pressing the red button on the payment terminal."
+                    "Please press the red button on the payment terminal to cancel the transaction."
                 )
             );
-            return Promise.resolve();
+            return Promise.reject();
         },
 
         _oca_payment_terminal_pay: function() {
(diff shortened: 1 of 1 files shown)


sbidoul:
@thomaspaulb thanks for this. LGTM. Could you look after the pre-commit issue ?

pos_payment_terminal/static/src/js/payment_terminal.js
```diff
@@ -169,11 +177,21 @@ odoo.define("pos_payment_terminal.payment", function(require) {
...
         _show_error: function(msg, title) {
+            if (!_.isEmpty(this.pos.gui.popup_instances))
             this.pos.gui.show_popup("error", {
                 title: title || _t("Payment Terminal Error"),
                 body: msg,
             });
         },
+
+        _show_tip: function(msg, title) {
+            if (!_.isEmpty(this.pos.gui.popup_instances))
```
sbidoul: pre-commit complains about this funny indentation

sbidoul:
I ported this to 14 in #741

ddejong-therp:
LGTM.

I've setup a PoS setup with a faulty payment terminal module that does not return a response back (does not implement get_status properly in the driver).

When I make a payment this faulty driver causes the interface to never complete the transaction, waiting indefinitely. When I then refresh the page, it stays on a black (dark grey) screen after loading, never really entering into the pos screen.

This MR fixes that, allowing you to enter the pos screen again.

ddejong-therp (approved):
As stated in the comments: LGTM

ivantodorovich (approved):

ivantodorovich:
Let's retry merging this after:
- https://github.com/OCA/pos/pull/795

PR #575, opened 2020-12-18
14.0 mig pos backend partner rebased

sebastienbeau:

PR #709, opened 2021-09-17
[14] pos_backend_communication: add maintainers in manifest

hparfr:


legalsylvain:
Hi @hparfr. Thanks for adopting this module.
Fast merging this trivial PR.

In the other hand, reading the readme, it seems that this module is a base module for other implementation like pos_backend_communication but I don't see the module for V14 and no PR.
The last PR is regarding V12 by @PierrickBrun. CI is broken and it is in a stale state : https://github.com/OCA/pos/pull/554

How do you use this V14 module ? Could you add pos_backend_partner V14 PR ?

thanks ! 

/ocabot merge patch

hparfr:
There is an existing PR for pos_backend_partner in v14 in #575  since 10 Months

PR #712, opened 2021-10-04
[14.0][ADD] pos_receipt_hide_price

paradoxxxzero:
Adds a button in the receipt view to hide price from the receipt for printing and emailing purpose


pos_receipt_hide_price/static/src/js/OrderReceipt.js
```diff
@@ -0,0 +1,19 @@
+odoo.define("pos_receipt_hide_price.OrderReceipt", function (require) {
+    "use strict";
+
+    const OrderReceipt = require("point_of_sale.OrderReceipt");
+    const Registries = require("point_of_sale.Registries");
+
+    const HidePriceOrderReceipt = (OrderReceipt) =>
+        class extends OrderReceipt {
+            constructor(_, {hpState}) {
```
legalsylvain: If I understand correctly hp is for hidePrice ? I think ``hidePriceState`` could be more explicit. don't you think ?

legalsylvain (commented):
Hi @paradoxxxzero. 

This PR (and module) is depecrating https://github.com/OCA/pos/pull/682 don't you think ?

Some remarks inline.

Nice to have : Could you add a print screen in the description ? 

Unfortunately not owl expert : no JS review.

CC : @dsolanki-initos, @ivantodorovich

pos_receipt_hide_price/static/src/js/OrderReceipt.js
paradoxxxzero: Agreed

paradoxxxzero:
Hello @legalsylvain, thanks for the review.

I think it's a slightly different take solving the same problem so yes it may deprecate #682.

I'll update the description with some screenshots.

ivantodorovich:
> Hello @legalsylvain, thanks for the review.
> 
> I think it's a slightly different take solving the same problem so yes it may deprecate #682.
> 
> I'll update the description with some screenshots.

Oh #!$& !
I just asked ocabot to merge https://github.com/OCA/pos/pull/682, before reading this :/

ivantodorovich (approved):
Technically looks good!

legalsylvain:
Thanks @paradoxxxzero  for the screenshots.

> Oh #!$& !
> I just asked ocabot to merge #682, before reading this :/

Not a big deal, maybe we can add a small text in the ROADMAP.rst of the other module, to mention to port ``pos_receipt_hide_price`` in next version and not the other one ?

/ocabot merge patch

PR #714, opened 2021-10-07
[FIX] pos_margin: avoid located CI issues

chienandalu:
Ensure test correctness when a different default localization is set on the CI as the existing pricelist could be in a different currency and that will cause all kind of test failures.

cc @Tecnativa TT32039

### DIFF
--- a/pos_margin/tests/test_module.py
+++ b/pos_margin/tests/test_module.py
@@ -11,10 +11,20 @@ class TestModule(TransactionCase):
         super(TestModule, self).setUp()
         self.PosOrder = self.env['pos.order']
         self.pos_product = self.env.ref('point_of_sale.whiteboard_pen')
-        self.pricelist = self.env.ref('product.list0')
-
+        self.pricelist = self.env["product.pricelist"].create({
+            "name": "Test pricelist",
+            "currency_id": self.env.user.company_id.currency_id.id,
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks ! 

fast tracking that patch.

/ocabot merge patch

BTW, just discovered that Odoo 15.0 handle now margin on pos.order, making this module obsolete. (good news!) In the other hand, the field is currently not stored. Any point of view on that topic ? 

see : https://github.com/odoo/odoo/pull/66689#discussion_r630341528

chienandalu:
> BTW, just discovered that Odoo 15.0 handle now margin on pos.order, making this module obsolete. (good news!) In the other hand, the field is currently not stored. Any point of view on that topic?

Well, for our current use case we need to aggregate the margin to the `sale.report` as other measures that `pos_sale` tries to merge. So we need it to be stored.

Indeed I didnt't add that `sale.report` extension as I wasn't sure if depending on `pos_sale` was going to be so much, so a glue module would be needed.

In that regard, another issue with Odoo's approach is that the margin is set on `point_of_sale` main module, while margins in sales are added in the `sale_margin` module, so maybe a future approach would need to override Odoo's core field to enable that reporting option with a `pos_sale_margin` module.

PR #726, opened 2021-11-05
 [ADD] new module pos_prevent_double_closing

legalsylvain:
This module extends Odoo Point Of Sale features, to prevent user to double close sessions.

This can happen if the user clicks twice on the "close session" button in two different windows.

In this case, Odoo will generate two sales entries, one of which will have a zero amount.
Deleting it will create a hole in the sales entries numbering.
duplicated_entries

This trivial module prevent such cases, raising an error if user try to close closed sessions.

CC : @OCA/pos-maintainers 

Note : I propose the name ``pos_prevent_account_duplicates`` but if you have a better idea, feel free to propose a better alternative. (CC : @pedrobaeza ;-))


pedrobaeza:
Seeing the code, this prevents to double close the session, avoiding the other situation about the double entry, but it can be named then more generally as `pos_prevent_double_closing`, don't you think?

legalsylvain:
@pedrobaeza : perfect suggestion ! I renamed the module and updated the description of the module and of the PR.

flotho (approved):
code review,
good expected feature!
thanks a lot @legalsylvain

pos_prevent_double_closing/models/pos_session.py
```diff
@@ -0,0 +1,26 @@
...
+        return super(
+            PosSession, self.with_context(action_pos_session_close=True)
+        ).action_pos_session_close()
+
+    @api.multi
+    def write(self, vals):
+        if self.env.context.get("action_pos_session_close", False):
+            closed_sessions = self.filtered(lambda x: x.state == "closed")
+            if ("state" in vals and vals["state"] == "closed"
+                    and closed_sessions):
```
ivantodorovich:
```suggestion
            if (vals.get("state") == "closed" and closed_sessions):
```
just to avoid the ugly line break 😅

ivantodorovich (approved):

pos_prevent_double_closing/models/pos_session.py
flotho: won't it cause lint issues ?

legalsylvain: > just to avoid the ugly line break sweat_smile

Thanks ! 

> won't it cause lint issues ?

Let's see!

PR #735, opened 2021-12-02
[FIX] Trigger orderline change on remove function

AntoniRomera:


ivantodorovich (commented):
I don't understand why this is needed, could you elaborate?

AFAICS this widget doesn't ever emit `change`

AntoniRomera:
I refer you to the line for which I think the trigger is needed, otherwise the order button is not activated (it turns green). 

[SubmitOrderButton.js#L20](https://github.com/odoo/odoo/blob/c29fe9fa8fc8ff4a5e6f042607ef9b3070d1d339/addons/pos_restaurant/static/src/js/Screens/ProductScreen/ControlButtons/SubmitOrderButton.js#L20)

ivantodorovich:
> I refer you to the line for which I think the trigger is needed, otherwise the order button is not activated (it turns green).
> 
> [SubmitOrderButton.js#L20](https://github.com/odoo/odoo/blob/c29fe9fa8fc8ff4a5e6f042607ef9b3070d1d339/addons/pos_restaurant/static/src/js/Screens/ProductScreen/ControlButtons/SubmitOrderButton.js#L20)

But that's another object.

Here you're triggering `change` on the [`Orderline` **widget**](https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/static/src/js/Screens/ProductScreen/Orderline.js).
Not to be confused with the [`orderline` **model**](https://github.com/odoo/odoo/blob/35e7655dd7673d497305d14ef4bedb3d8893bc56/addons/point_of_sale/static/src/js/models.js#L1641) that would probably be `this.props.line.trigger('change')`.
But also the link you provided is refering to the [`OrderlineCollection`](https://github.com/odoo/odoo/blob/35e7655dd7673d497305d14ef4bedb3d8893bc56/addons/point_of_sale/static/src/js/models.js#L2605)

(And by the way the orderlines collection should emit this even without this fix, as it would already be taken care of by `this.props.line.set_quantity("remove");`)


Just to be clear, I'm not saying this fix doesn't work.. only that it's a bit confusing and I didn't find any case where the Orderline widget emits a `change` event. AFAIK that event is used only on models, not widgets.

AntoniRomera:
A ok, the thing is that it is the only way that I have found that when I click on the delete button it activates correctly the button to send the order to the printer, for the moment we could put it as a FIX (until someone else finds where the error is or what line of code is missing) because if not the module was made merge without the relevant tests.

AntoniRomera:
Sorry I close this PR, my mistake

PR #736, opened 2021-12-09
[ADD] 14.0 pos_reuse_pricelist module

PierrickBrun:


pos_reuse_pricelist/static/src/js/models.js
```diff
@@ -0,0 +1,27 @@
...
+
+    models.PosModel = models.PosModel.extend({
+        on_removed_order(removed_order, index, reason) {
+            posmodel_super.on_removed_order.apply(this, arguments);
+            var order_list = this.get_order_list();
+            if (
+                (reason === "abandon" || removed_order.temporary) &&
+                order_list.length > 0
+            ) {
+                // Do nothing when this is a deletion (and there is another order available)
```
ivantodorovich: Better to inverse that `if` statement and remove this empty block

pos_reuse_pricelist/static/src/js/models.js
```diff
@@ -0,0 +1,27 @@
+/*
+ *  Copyright 2021 Akretion
+ *  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+ */
+
+odoo.define("pos_reuse_pricelist.models", function (require) {
+    "use strict";
+
+    var models = require("point_of_sale.models");
```
ivantodorovich:
```suggestion
    const models = require("point_of_sale.models");
```

Replace all `var` with `const`

https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/

pos_reuse_pricelist/static/src/js/models.js
```diff
@@ -0,0 +1,27 @@
...
+                order_list.length > 0
+            ) {
+                // Do nothing when this is a deletion (and there is another order available)
+            } else {
+                // Reuse the pricelist
+                this.get_order().pricelist = removed_order.pricelist;
+            }
+        },
+    });
+});
```
ivantodorovich, line 27:
```suggestion
    return models;
});
```

This is to play nice with other modules, making it easier to extend this one

pos_reuse_pricelist/static/src/js/models.js
PierrickBrun: The idea was to reuse the same `if` statement as the one in `posmodel_super.on_removed_order` to make it easier to understand and maintain.
I think the inversed statement may be harder to read

ivantodorovich: Something like this maybe?
Could be accompanied with a comment

```js
    const newOrderCreated = !((reason === "abandon" || removed_order.temporary) && order_list.length > 0);
    if (newOrderCreated) {
        this.get_order().pricelist = removed_order.pricelist;
    }
```

Anyway, not blocking :)

ivantodorovich (approved):

pos_reuse_pricelist/static/src/js/models.js
PierrickBrun: Good idea

legalsylvain (approved):
Lgtm

PR #737, opened 2021-12-10
[14.0][ADD] pos debranding module

ilyasProgrammer:


legalsylvain (changes requested):
Hi @ilyasProgrammer.

thanks a lot for sharing this module.

AFAIK, there is a specific repository for all the unbranding stuff here : https://github.com/OCA/server-brand

I think that it should be better put in that repo.

what do you think ?

ilyasProgrammer:
> 

Hello. Well it's related both debranding and POS. So not sure where to push it. I can create PR to server-brand if you think it's fits better there.

legalsylvain:
> Hello. Well it's related both debranding and POS. So not sure where to push it. I can create PR to server-brand if you think it's fits better there.

yes, I think it's fits better in server-brand.

ilyasProgrammer:
> > Hello. Well it's related both debranding and POS. So not sure where to push it. I can create PR to server-brand if you think it's fits better there.
> 
> yes, I think it's fits better in server-brand.

ok

PR #741, opened 2021-12-23
[14.0][IMP] pos_payment_terminal: Graceful cancel handling

sbidoul:
Co-authored-by: Tom Blauwendraat <tom@sunflowerweb.nl>

Forward port of #708


sbidoul:
I have further simplified by displaying an error when the users attempts to cancel the transaction.
This is indeed an error, as we don't support the cancellation buttons, and they have no effect. So we draw the attention of the user with a red message so they actually physically cancel the transaction on the terminal.

This is now ready.

thomaspaulb (approved):

ivantodorovich (approved):

PR #749, opened 2022-01-13
[14.0][FIX] pos_reset_search: Iterate on owl children

rousseldenis:


rousseldenis:
I don't known if it's an elegant one but I've done the same as Odoo in its own code

ivantodorovich (approved):
Thanks!!! 🙏🏻

It's just precommit complaining

legalsylvain:
there is an error when I launch pos on runboat. could you check ? 

image

rousseldenis:
> there is an error when I launch pos on runboat. could you check ?
> 
> image

Yes I'm on it

rousseldenis:
@legalsylvain @ivantodorovich How do you solve the 'no-shadow' error in js ?

legalsylvain:
> @legalsylvain @ivantodorovich How do you solve the 'no-shadow' error in js ?

no idea.

ivantodorovich:
It seems you already fixed it ^^
For future reference: https://eslint.org/docs/rules/no-shadow

rousseldenis:
> It seems you already fixed it ^^ For future reference: https://eslint.org/docs/rules/no-shadow

No. For a reason I ignore, locally, pre-commit does not accept warnings. If you look here the pre-commit run, you have warnings.

image

ivantodorovich:
Basically it comes from this:
- `ProductScreen` is declared [here](https://github.com/acsone/pos/blob/35e916272a2f850735208e7cfa9d6af2293ff01c/pos_reset_search/static/src/js/ProductScreen.js#L4)
- and `ProductScreen` is also used as an argument name for [this function](https://github.com/acsone/pos/blob/35e916272a2f850735208e7cfa9d6af2293ff01c/pos_reset_search/static/src/js/ProductScreen.js#L7)

ESLint is complaining that the function argument used in the lower scope is overlapping with the `const ProductScreen` in the upper scope.


--

But.. we should probably disable that warning as it goes against Odoo's pattern for extending POS components.

- See [here for an example](https://github.com/odoo/odoo/blob/14.0/addons/pos_restaurant/static/src/js/Screens/ReceiptScreen/ReceiptScreen.js)

rousseldenis:
> Basically it comes from this:
> 
> * `ProductScreen` is declared [here](https://github.com/acsone/pos/blob/35e916272a2f850735208e7cfa9d6af2293ff01c/pos_reset_search/static/src/js/ProductScreen.js#L4)
> * and `ProductScreen` is also used as an argument name for [this function](https://github.com/acsone/pos/blob/35e916272a2f850735208e7cfa9d6af2293ff01c/pos_reset_search/static/src/js/ProductScreen.js#L7)
> 
> ESLint is complaining that the function argument used in the lower scope is overlapping with the `const ProductScreen` in the upper scope.
> 
> --
> 
> But.. we should probably disable that warning as it goes against Odoo's pattern for extending POS components.
> 
> * See [here for an example](https://github.com/odoo/odoo/blob/14.0/addons/pos_restaurant/static/src/js/Screens/ReceiptScreen/ReceiptScreen.js)

Yes. IMHO, this should be done in repos configuration globally.

An advice where to put this exception for eslint @sbidoul  ?

ivantodorovich:
Most probably we just need to remove this line: https://github.com/OCA/oca-addons-repo-template/blob/31c17c41a539eb43346e24db8ac8ca37fcebcb9e/src/.eslintrc.yml#L129

legalsylvain (approved):
Thanks a lot @rousseldenis for fixing that module.

/ocabot merge patch

PR #757, opened 2022-02-09
[13.0] [FIX] pos_margin: avoid located CI issues

CarlosRoca13:
Ensure test correctness when a different default localization is set on
the CI as the existing pricelist could be in a different currency and
that will cause all kind of test failures.

cc @Tecnativa TT32039

Cherry pick of https://github.com/OCA/pos/commit/c672b53bf89519957e9465f4f8c2c3a91a6da0ca

ping @chienandalu @pedrobaeza


chienandalu (approved):
If tests go ok :+1:

CarlosRoca13:
Tested on CI and it's working well

PR #758, opened 2022-02-15
[14.0] mig pos picking delayed

fshah-initos:


pos_picking_delayed/views/view_pos_order.xml
```diff
@@ -1,28 +1,29 @@
...
 
     <record id="view_pos_order_form" model="ir.ui.view">
         <field name="model">pos.order</field>
-        <field name="inherit_id" ref="point_of_sale.view_pos_pos_form"/>
+        <field name="inherit_id" ref="point_of_sale.view_pos_pos_form" />
         <field name="arch" type="xml">
-            <field name="picking_id" position="after">
-                <field name="has_picking_delayed"/>
+            <field name="company_id" position="before">
+                <field name="has_picking_delayed" />
```
legalsylvain, line 26: why this change ? setting ``has_picking_delayed`` after picking_id makes sense. don't you think ? 
or maybe the field disappeared ?

legalsylvain (changes requested):
One minor comment. 

Also, could you distinct 2 commits for the migration ? 
- the pre-commit one
- the real changes

it makes the review more easy.

thanks !

PR #762, opened 2022-03-08
[12.0] pos_loyalty

eLBati:
From https://github.com/OCA/pos/pull/414

PR #763, opened 2022-03-09
[FIX] pos_customer_display and pos_payment_terminal :  disable use_proxy on demo data

legalsylvain:
remove default value that enable use_proxy and make other tests failing

CC : @eLBati 

Fixes : test in : https://github.com/OCA/pos/pull/762

**Note**
- generally, it's great that new features are enabled by default on demo data. it makes the module "ready to test" on runbot / runboat / or locally.
- however, enabling iface (proxy) features enable ``use_proxy``, so it is changing the behaviour of the point of sale, and make the JS tests of the point of sale not predictive.

### DIFF
--- a/pos_customer_display/__manifest__.py
+++ b/pos_customer_display/__manifest__.py
@@ -14,6 +14,5 @@
         "views/assets.xml",
         "views/view_pos_config.xml",
     ],
-    "demo": ["demo/pos_config.xml"],
     "installable": True,
 }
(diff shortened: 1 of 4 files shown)


flotho (approved):
Thanks @legalsylvain 
LGTM

PierrickBrun (approved):

legalsylvain:
thanks for your reviews.

/ocabot merge patch

PR #766, opened 2022-03-18
[12.0][ADD] new pos_order_zipcode addon

petrus-v:
Collect POS customer zip code to determine the catchment area.

This module add zip code field on POS order. POS sessions that require
catchment area will add an extra check before validate the payment
to ensure a zip code has been collected. If cashier choose a partner
or update its zip code this will set the pos order zip code.


pos_order_zipcode/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2022 Moka Tourisme (https://www.mokatourisme.fr).
+# @author Pierre Verkest <pierreverkest84@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    pos_order_zipcode_zip_required = fields.Boolean(
```
legalsylvain, line 10: In a functionnal point of view, I see this part very bad. 
No personal data should be required to make a sale.
A customer can refuse to provide such information.
The result of such constraint is having database with incorrect data, because customer say anything (or cashier will enter anything).

pos_order_zipcode/models/pos_order.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2022 Moka Tourisme (https://www.mokatourisme.fr).
+# @author Pierre Verkest <pierreverkest84@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class PosOrder(models.Model):
+    _inherit = "pos.order"
+
+    zipcode = fields.Char("Zip")
```
legalsylvain, line 10: The rest of odoo name the zipcode ``zip``. (partner / bank / company).

legalsylvain (commented):
2 remarks inline. 
LGTM otherwise. (code review / no test).

thanks for sharing this work.

kind regards.

pos_order_zipcode/models/pos_order.py
petrus-v: I thought about it, `zipcode` is used in `base_address_city` module and I'm not english I've feeling zipcode is less ambiguis than zip ?

pos_order_zipcode/models/pos_config.py
petrus-v: I'll add something in the road map something that cashier could click on a button to by pass that step anyway.

The purpose of this module is also to remind the cashier to ask about it, In the meantime I suppose they set a special code to exclude those orders.

As Today the analysis is done by an other software.

pos_order_zipcode/models/pos_order.py
petrus-v: .... and also because [zip](https://docs.python.org/3/library/functions.html#zip) is a python built-in function !

legalsylvain: Well. 
regarding naming, ``zip_code`` (``zipcode``) is indeed better. 

if you execute ``rgrep "zipcode.*field" ./ --include=*.py``, you'll see that most of the fields are ``zip``. But indeed there a lost ``zipcode`` in base_adress_city``. (and also a ``zip_code`` in ``lunch`` V15 module).

My preference is to follow the (pseudo-)convention of odoo and wait they change.

Not a blocking point for me.

legalsylvain (approved):

chienandalu (commented):
I find kinda weird this flow. Is it on purpose?

1. Payment screen -> Validate
2. Enter zip code -> Validate
3. Payment screen -> Validate
4. Receipt screen
zipcode-flow

petrus-v:
> I find kinda weird this flow. Is it on purpose?
> 
>     1. Payment screen -> Validate
> 
>     2. Enter zip code -> Validate
> 
>     3. Payment screen -> Validate
> 
>     4. Receipt screen
>        zipcode-flow

Thanks for testing.

Well, you can click on zip code button before validate the payment (notice that zip code button is highlight while validate button is still grey until you set Zip code). I've though it's quicker to open zip code edition instead raise an error to invite user to click on the Zip code button while trying to validate , that's why you had to validate twice.

I'm thinking... it can probably improved by recalled the method validate method in case user clicked on validate button before collect zipcode as in your gif preview with a possible way to stop the loop if user don't want to collect that zipcode. Or I could raise an exception :thinking:

petrus-v:
> I'm thinking... it can probably improved by recalled the method validate method in case user clicked on validate button before collect zipcode as in your gif preview with a possible way to stop the loop if user don't want to collect that zipcode. Or I could raise an exception thinking

Using `show_screen` function I can't see a way to provide a callback on validate, maybe I should move this screen as a popup using the `show_popup` function using the confirm option.

I'm not sure we can activate virtual keyboard in popup screen. I'm thinking to move that in the roadmap.

Any thoughts ?

PR #769, opened 2022-03-28
[ADD] pos_ticket_static_qrcode

petrus-v:
This module add QRCode on the POS ticket.
This QRCode value and message can be set per point of sale.


petrus-v:
I don't understand why travis failed with that change :thinking:

chienandalu:
Looks like a patch from Odoo was added to `12.0` but isn't compatible with the travis python version (`f""` for format strings): https://app.travis-ci.com/github/OCA/pos/jobs/564912567#L1047

petrus-v:
@chienandalu good catch indeed python fstring came with python 3.6 we are using python 3.5.

A new module [fetchmail_outlook](https://github.com/OCA/OCB/commit/7d5207b4fc41d876faf917a233b3ca8cb2542af5) with l `"auto_install": True,` has been added.

Not sure what's the best way for us, contribute to odoo to transform f-string to %-string ?!

Related odoo PR: https://github.com/odoo/odoo/pull/87040

petrus-v:
> @chienandalu good catch indeed python fstring came with python 3.6 we are using python 3.5.

and the oldest python version maintain with security fixes is python 3.7.

As far I remember odoo is supposed to support the python version installed on debian the day of the odoo release, I suppose travis version has been decide with this pattern, anyway I'll create an other PR with python 3.7 to give a try a get result travis

legalsylvain (commented):
1. Naive question : you overloaded PosTicket and XmlTicket, but you say it doesn't if it's not via proxy. i don't understand why ? 

2. seeing a module named ``pos_ticket_qrcode`` I was expecting a module that add a qrcode that identify the ticket. = That could allow to scan it again, to reload / refund / etc.... If I uderstand correctly, the qrcode is static. (not a critic, but maybe the name could be changed.) something like ``static_qrcode`` or ``global_qrcode``.) Don't you think ?

thanks for sharing that work.

petrus-v:
@legalsylvain Thanks for your prompt reply

> 1. Naive question : you overloaded PosTicket and XmlTicket, but you say it doesn't if it's not via proxy. i don't understand why ?

I were talking about odoo feature if you provide xml in  footer, it won't be displayed on browser but [XmlTicket](https://github.com/OCA/OCB/blob/12.0/addons/point_of_sale/static/src/xml/pos.xml#L734) use it, I'll rewords or remove that part.

> 
>     2. seeing a module named `pos_ticket_qrcode` I was expecting a module that add a qrcode that identify the ticket. = That could allow to scan it again, to reload / refund / etc.... If I uderstand correctly, the qrcode is static. (not a critic, but maybe the name could be changed.) something like `static_qrcode` or `global_qrcode`.) Don't you think ?

hum. good point ! :thinking:   I'll rename to `pos_ticket_static_qrcode`

> 
> thanks for sharing that work.

hitrosol:
Functional review okay.

pos_ticket_static_qrcode/models/pos_config.py
```diff
@@ -0,0 +1,60 @@
...
+                    qr.add_data(self.pos_ticket_static_qrcode_value)
+                    qr.make()
+                    img = qr.make_image(
+                        fill_color="white", back_color="black"
+                    )
+                    png = io.BytesIO()
+                    img.save(png)
+                    png.seek(0)
+                except Exception:
+                    raise ValueError("Cannot convert into barcode.")
```
legalsylvain: Into QR code, i guess ?

pos_ticket_static_qrcode/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2022 Moka Tourisme (https://www.mokatourisme.fr).
+# @author Pierre Verkest <pierreverkest84@gmail.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
legalsylvain: Nitpicking (not blocking) copyright are useless un init files.

legalsylvain: Nitpicking (not blocking) copyright are useless un init files.

legalsylvain (approved):
Code review. No test.

pos_ticket_static_qrcode/models/pos_config.py
petrus-v: Thanks for your review I've improve this message and the exception is now covered by unittest !

pos_ticket_static_qrcode/tests/__init__.py
petrus-v: done!

matteoopenf (approved):

PR #770, opened 2022-03-28
[FIX] use python 3.7 instead 3.5

petrus-v:
A new module fetchmail_outlook has been added with auto_install=True using
f-string which start from python 3.6. As long python3.5 and 3.6 are not
suppoted by python, move to the oldest python version supported

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache:
     - $HOME/.cache/pip
 
 python:
-  - "3.5"
+  - "3.7"
 
 addons:
   postgresql: "9.6"


petrus-v:
I'm trying to fix CI discovered via https://github.com/OCA/pos/pull/769

legalsylvain (commented):
why not 3.6. If i understand correctly, gitlab ci 12.0 use that revision. 

https://github.com/grap/oca-addons-repo-template/blob/grap/src/.github/workflows/%7B%25%20if%20ci%20%3D%3D%20'GitHub'%20%25%7Dtest.yml%7B%25%20endif%20%25%7D.jinja#L15

rousseldenis:
> why not 3.6. If i understand correctly, gitlab ci 12.0 use that revision.
> 
> https://github.com/grap/oca-addons-repo-template/blob/grap/src/.github/workflows/%7B%25%20if%20ci%20%3D%3D%20'GitHub'%20%25%7Dtest.yml%7B%25%20endif%20%25%7D.jinja#L15

IMHO, this has to be fixed in Odoo (at least in OCB).

I think OCA directive is to be on the minimal Odoo python version requirement:

https://github.com/odoo/odoo/blob/12.0/setup.py#L59

legalsylvain:
> IMHO, this has to be fixed in Odoo (at least in OCB).

indeed.

petrus-v:
@rousseldenis ok thanks for your input, I'm closing this issue then !

rousseldenis:
FYI https://github.com/odoo/odoo/pull/87389

PR #671, opened 2021-06-08
[14.0] [mig] pos_order_return

dsolanki-initos:

PR #771, opened 2022-03-29
14.0 meta update copier 2022 03 29

legalsylvain:
- applying https://github.com/OCA/oca-addons-repo-template/pull/134
- disable test of ``pos_order_return`` and set devolpment status to Alpha

**Note** : move from travis to github action.


pos_order_return/tests/test_pos_order_return.py
```diff
@@ -116,7 +116,9 @@ def setUp(self):
         res = self.pos_order.action_pos_order_invoice()
         self.invoice = self.env["account.move"].browse(res["res_id"])
 
-    def test_pos_order_full_refund(self):
+    # Disabled test.
+    # TODO: FIXME.
+    def _test_pos_order_full_refund(self):
```
petrus-v: I'm wondering if odoo as test launcher support [@unittest.skip("FIXME: skip while moving to github action")](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) decorator ?

At least pytest users will see skipped test in test reports

Also it could be nice to explain why we skip it in the mean time.

petrus-v (approved):
Thanks a lot taking care OCA chages. As far CI passed on github action it LGTM.

Few comment on the way to disable test but wont block for that !

legalsylvain:
hi @OCA/pos-maintainers. 

I applied https://github.com/OCA/oca-addons-repo-template/pull/134 and fixed some minors points.

A test is failing. locally, all is ok. I suspect a conflict with another module in OCA/pos, but it's hard to investigate. 

I so disabled the tests of ``pos_order_return`` et set the module to alpha level, to have the possibility to move on.

I see it acceptable. Any review welcome on that topic.

Note : I just discovered the decorator @unittest.skip that explicitely log the skip. see : 
https://github.com/OCA/pos/runs/5785246557?check_suite_focus=true#step:8:204

rousseldenis (approved):
Great!

chienandalu (approved):
:+1: Thanks!

legalsylvain:
@OCA/community-maintainers : any idea regarding that message ? 

```failed to push some refs to 'https://***@github.com/OCA/pos'```

**Comment by sbidoul** [comment]
> a Personal Access Token to create or update workflow `.github/workflows/test.yml` without `workflow` scope

This is https://github.com/OCA/oca-github-bot/issues/173

Can you try again now ?

File: pos_order_return/tests/test_pos_order_return.py [hunk 2]
```diff
@@ -116,6 +118,7 @@ def setUp(self):
         res = self.pos_order.action_pos_order_invoice()
         self.invoice = self.env["account.move"].browse(res["res_id"])
 
+    @unittest.skip("Errors when other OCA / Pos modules are installed")
```
sbidoul: Out of curiosity, why were these test passing with travis ?

legalsylvain: I got it ! 

Travis became red very recently.

https://github.com/OCA/pos/commit/3b955a12895cdd6a1c72ce34aaaf684c6c859a2b was green.
https://app.travis-ci.com/github/OCA/pos/jobs/557872233#L1609

https://github.com/OCA/pos/commit/f7131a15bdfb2ebdc95d1efe7019a33dde91b9ec was red : 
https://app.travis-ci.com/github/OCA/pos/jobs/564379816#L1606

image

the only changes are translation, so that's not because of recent OCA changes.

I tested again ``locally`` with an updated odoo core, and the module doesn't work now. So something changed in Odoo core. (point of sale or stock), that makes the tests failing.
I updated the text of the "skip" part.

@dsolanki-initos : for your information, the module you migrated (https://github.com/OCA/pos/pull/671) is now failing.

PR #774, opened 2022-04-01
[FIX] pos_cash_move_reason

AntoniRomera:
- Changed the field where the reason for the withdrawal or receipt of money is indicated, if left in the field name, it changes the name of the accounting entry and gives an error.

### DIFF
--- a/pos_cash_move_reason/wizard/wizard_pos_move_reason.py
+++ b/pos_cash_move_reason/wizard/wizard_pos_move_reason.py
@@ -104,7 +104,6 @@ class WizardPosMoveReason(models.TransientModel):
             "statement_id": self.statement_id.id,
             "journal_id": self.journal_id.id,
             "amount": amount,
-            "name": self.name,
-            "payment_ref": self.session_id.name,
+            "payment_ref": f"{self.session_id.name} - {self.name}",
             "counterpart_account_id": account_id,
         }


AntoniRomera:
@OCA/pos-maintainers Travis is in red due to another module test, can anyone help? The error in this module blocks all billing in the system.

chienandalu:
There's ongoing work here: https://github.com/OCA/pos/pull/771

FernandoRomera (approved):

AntoniRomera:
@OCA/pos-maintainers Anyone can review?

legalsylvain (changes requested):
thanks for sharing your patch.

- I don't understand how an error is raised. could explain  steps to reproduce ?
- I don't think loosing session name is acceptable for an accountant. What do you think ?

AntoniRomera:
- The steps are simple.

> 1. Open a session
> 2. You make a transaction to withdraw money (I understand that you have already configured this module).
> 3. And as you can see now when you create an "account.bank.statement.line", if you pass the field "name" it is put in the field "name" of the accounting entry that odoo creates automatically (To see it better you can even see how Odoo creates the account.bank.statement.line when you use the original option to withdraw or deposit money). When this happens it crashes the system because in Odoo 14 it checks the name of the accounting entries with a regular expression from the sequence

- If you do not want to lose the session name you can put in front of the reason "POS/0000 - {reason}".

legalsylvain:
thanks for the explanation. I'll test this week.
regards.

AntoniRomera:
@OCA/pos-maintainers @legalsylvain Can anyone else check? I'm in a bit of a hurry

elvise:
> thanks for the explanation. I'll test this week.
> regards.

@legalsylvain any news about this ?

legalsylvain (approved):
code review / no test.

/ocabot merge patch

PR #776, opened 2022-04-06
[ADD] pos_hide_cost_price_and_margin: Add to 15.0

hailangvn:
This module is to hide cost, margin, and replenishment info from product detail in pos. It utilize a setting from `pos_margin` module but doesn't need to depend on the later module thanks to the mechanism that if there is no field `iface_display_margin`, the value of `t-if` is False.
image


flotho (approved):
tested on ruboat

leemannd (approved):
Code LGTM.

But functionnaly, I'm wondering if we don't want to keep the fields visibility for admin users or a group of pos admin

pos_hide_cost_price_and_margin/static/src/xml/pos_margin.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+Copyright (C) 2022 - Today: camptocamp (https://www.camptocamp.com)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+<templates id="template" xml:space="preserve">
+
+    <t t-inherit="point_of_sale.ProductInfoPopup" t-inherit-mode="extension">
+        <xpath expr="//table/tr/td/t[@t-esc='costCurrency']/../.." position="replace" />
```
leemannd: This can be an issue. If another module is trying to make an xpath expression based on the field you're removing. it lead to an error that is time consuming to debug.
It should be hidden, not removed.

Same is for others below

pos_hide_cost_price_and_margin/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2022 - Today: camptocamp (https://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
ajaniszewska-dev, line 2:
```suggestion
# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
```

there is LGPL in the line 12

cvinh: AFAIK, all OCA modules are AGPL-3.0

pos_hide_cost_price_and_margin/static/src/xml/pos_margin.xml
gurneyalex: can't we set a priority on qweb templates, similar to what we do on views, to ensure that this replacement is done in the end?

hailangvn:
Thanks all for your comments. The PR has been updated accordingly.

leemannd (approved):

legalsylvain:
/ocabot merge patch

(no review from my part).

PR #748, opened 2022-01-13
[14.0][MIG] pos_customer_required: Migration to 14.0

KNVx:

PR #779, opened 2022-04-07
[13.0][MIG][pos_customer_required] Migrated to v13.0

robyf70:
Migrated to v13.0


robyf70:
@francesco-ooops Can you please review so we can proceed quicker?

robyf70:
@ivantodorovich Can you please also review this?

pos_customer_required/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "name": "Point of Sale Require Customer",
+    "version": "13.0.1.0.0",
+    "category": "Point Of Sale",
+    "summary": "Point of Sale Require Customer",
+    "author": "Apertoso NV, La Louve, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/pos",
+    "license": "AGPL-3",
+    "depends": ["point_of_sale"],
+    "data": [
+        "static/src/xml/templates.xml",
```
ivantodorovich: The location of this file is misleading.
`static/src` is meant for content served statically to the frontend, not `ir.ui.view` definitions.

Could you please move this file to:
```suggestion
        "views/assets.xml",
```

pos_customer_required/models/pos_make_payment.py
```diff
@@ -0,0 +1,28 @@
...
+
+class PosMakePayment(models.TransientModel):
+    _inherit = "pos.make.payment"
+
+    def check(self):
+        # Load current order
+        order_obj = self.env["pos.order"]
+        order = order_obj.browse(self.env.context.get("active_id", False))
+
+        # Check if control is required
```
ivantodorovich:
```suggestion
        order = self.env["pos.order"].browse(self.env.context.get("active_id", False))
```

Nitpicking here but would be nice to remove the obvious comments and do a one-liner

pos_customer_required/static/src/js/pos_customer_required.js
```diff
@@ -0,0 +1,60 @@
...
+    The licence is in the file __manifest__.py
+*/
+
+odoo.define("pos_customer_required.pos_customer_required", function(require) {
+    "use strict";
+
+    var screens = require("point_of_sale.screens");
+    var gui = require("point_of_sale.gui");
+    var core = require("web.core");
+    var _t = core._t;
```
ivantodorovich:
```suggestion
    const screens = require("point_of_sale.screens");
    const gui = require("point_of_sale.gui");
    const core = require("web.core");
    const _t = core._t;
```

pos_customer_required/static/src/js/pos_customer_required.js
```diff
@@ -0,0 +1,60 @@
...
+            !this.pos.get_order().get_client() &&
+            screen_name != "clientlist"
+        ) {
+            // We call first the original screen, to avoid to break the
+            // 'previous screen' mecanism
+            _show_screen_.call(this, screen_name, params, refresh);
+            screen_name = "clientlist";
+        }
+        _show_screen_.call(this, screen_name, params, refresh);
+    };
```
ivantodorovich: Could you split this file in two separate files: `screens.js` and `gui.js`, each with their corresponding overrides.
At the end of each file, return the `screens` or `gui` object, so that potential modules depending on this one can continue overriding if needed.

pos_customer_required/tests/test_pos.py
```diff
@@ -0,0 +1,49 @@
+import odoo.tests.common as common
+from odoo import exceptions
+
+
+class TestPosRequireCustomer(common.TransactionCase):
+    def setUp(self):
+        super(TestPosRequireCustomer, self).setUp()
+        self.pos_config = self.env.ref("point_of_sale.pos_config_main").copy()
```
ivantodorovich:
```suggestion
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        self.pos_config = self.env.ref("point_of_sale.pos_config_main").copy()
```

`setUp` runs for every case
`setUpClass` runs only once for all

ivantodorovich (approved):
thanks !!! 🥳

francesco-ooops (approved):
functional review ok

robyf70:
> thanks !!! partying_face

I guess we should get the #748 to be aligned in a similar way, right?

pos_customer_required/tests/test_pos.py
robyf70: mmmhhhh!!! This is breaking the test case. Should I revert?

robyf70: I did!

ivantodorovich: I had a typo in my suggestion
It should've been like this

```suggestion
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.pos_config = self.env.ref("point_of_sale.pos_config_main").copy()
```

Anyway feel free to keep it as it is

robyf70: Let's keep as before because using ``cls`` instead of ``self`` doesn't seem to work, in my patch I've replace ``self -> cls`` and indeed ``self`` isn't available.

robyf70:
I forgot on thing on the test case, now fixed

PR #780, opened 2022-04-08
[FIX] #777 pos_customer_required_fields

### LINKED ISSUE #777: [12.0][pos_customer_required_fields] if no customer required fields defined on pos config it's not possible to validate a pos order with customer
## Module
``pos_customer_required_fields``

## Describe the bug
if no customer required fields defined on pos config it's not possible to validate a pos order with customer.

## To Reproduce
**Affected versions**: 12.0

1. ensure no customer required fields defined on pos config
2. start pos session
3. pick a customer and start payment screen
4. try to validate payment

A popup with error message is displayed:

image

**Expected behavior**

payment should be validate without error

**Additional context**

technical explanation about this issue:

`this.pos.config.res_partner_required_fields_names.split(",")` return `[""]` one element which is not filtered because `this.pos.config.res_partner_required_fields_names == ""` in case there are no required fieds

petrus-v:
fix #777

In case there is no required field we don't want to
lock user on payment screen.

As required fields list is stored in string variable
spliting an empty string return an array of one empty
string which would return one missing field which
is locking users

### DIFF
--- a/pos_customer_required_fields/static/src/js/screens.js
+++ b/pos_customer_required_fields/static/src/js/screens.js
@@ -3,7 +3,7 @@
     @author Pierre Verkest <pierreverkest84@gmail.com>
     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
  */
-odoo.define('pos_order_zipcode.screens', function (require) {
+odoo.define('pos_customers_required_fields.screens', function (require) {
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

robyf70 (approved):
LGTM

PR #783, opened 2022-04-15
[ADD] new pos_margin_account_invoice_margin glue module

legalsylvain:
**Context**
- install ``pos_margin`` and ``account_invoice_margin`` (OCA/margin-analysis repo).
- go to point of sale and create a pos order, activating the generation of invoice.
- the pos.order has correct margin : OK
- the account.invoice has a bad margin (100%) : KO.

This simple glue module fixes this problem.

GRAP-411


chienandalu (approved):
Thanks!

pos_margin_account_invoice_margin/post_install.py
```diff
@@ -0,0 +1,43 @@
...
+
+from odoo import api, SUPERUSER_ID
+import logging
+logger = logging.getLogger(__name__)
+
+
+def set_margin_on_pos_invoices(cr, registry):
+    logger.info(
+        "Define purchase prices on invoice lines"
+        " created from Point of Sale. This could take a while ..."
```
chienandalu, line 13: Have you explored doing it with SQL?

legalsylvain: No : 
- it doesn't take a long time. On my production server, it took 30 second to handle all the invoices
- it's just about the pos order invoiced.
- it will take no time if you install ``pos_margin`` and ``pos_margin_account_invoice_margin`` at the same. (that will occur for new installation).

but if people want to convert into SQL, no problem.

Design of this module is temporary as ``pos_margin`` will disappear in new Odoo version (included in core feature). so, I guess we will rename the proposed module into ``account_invoice_margin_pos`` and move it ``margin-analysis`` in next version. (or maybe it will be unnecessary, if computation is done differently)

chienandalu: :+1:

ivantodorovich (approved):

PR #786, opened 2022-04-22
[12.0][ADD] pos_refresh_customer

carmenbianca:
Pinging @legalsylvain 

This is a simple module that will be soft-required by https://github.com/coopiteasy/addons/pull/222.

### DIFF
--- a/pos_refresh_customer/static/src/js/models.js
+++ b/pos_refresh_customer/static/src/js/models.js
@@ -1,0 +1,54 @@
+// Copyright 2022 Coop IT Easy SCRLfs
+// License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+odoo.define("pos_customer_wallet.screens", function (require) {
+    "use strict";
+    var models = require("point_of_sale.models");
+    var rpc = require("web.rpc");
+
+    models.PosModel = models.PosModel.extend({
+        load_partners_by_ids: function (ids) {
+            var self = this;
+            var def = new $.Deferred();
+            var fields = _.find(this.models, function (model) {
+                return model.model === "res.partner";
+            }).fields;
+
+            rpc.query(
+                {
+                    model: "res.partner",
+                    method: "read",
+                    args: [ids, fields],
+                },
+                {
+                    timeout: 3000,
+                    shadow: true,
+                }
+            ).then(
+                function (partners) {
+                    // Context: `add_partners` fetches partner models from the
+                    // Odoo database and inserts those models into its own db.
+                    // As an efficiency measure, it does not insert partner
+                    // models if their `write_date` is LESS recent than the last
+                    // time it recorded that a partner model was successfully
+                    // added to the pos db. Computed fields do not normally
+                    // increase `write_date`. Therefore, to make sure that
+                    // computed fields also get updated, just reset the cached
+                    // date to the Unix epoch.
+                    self.db.partner_write_date = "1970-01-01 00:00:00";
+
+                    if (self.db.add_partners(partners)) {
+                        // check if the partners we got were real updates
+                        def.resolve();
+                    } else {
+                        def.reject();
+                    }
+                },
+                function (type, err) {
+                    def.reject();
+                }
+            );
+            return def;
+        },
+    });
+});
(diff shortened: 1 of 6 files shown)


pos_refresh_customer/static/src/js/models.js
```diff
@@ -0,0 +1,54 @@
...
+                    // Context: `add_partners` fetches partner models from the
+                    // Odoo database and inserts those models into its own db.
+                    // As an efficiency measure, it does not insert partner
+                    // models if their `write_date` is LESS recent than the last
+                    // time it recorded that a partner model was successfully
+                    // added to the pos db. Computed fields do not normally
+                    // increase `write_date`. Therefore, to make sure that
+                    // computed fields also get updated, just reset the cached
+                    // date to the Unix epoch.
+                    self.db.partner_write_date = "1970-01-01 00:00:00";
```
carmenbianca, line 38: This works, but I'm not happy with this, and needs some tweaking.

It introduces one specific problem:

1. `add_partners()` is run on all partners. All is well and `partner_write_date` is consistent. Let's say it's equal to `2022-01-01 11:00:00`.
2. Partner 1 is edited at 13:00.
3. Partner 2 is edited at 14:00.
4. `load_partners_by_id()` (or `add_partners()`) is run on [Partner 2]. `partner_write_date` is now `2022-01-01 14:00`
5. `add_partners()` is run on [Partner 1]. This partner is skipped, because 13:00 is less recent than 14:00.

Part of this problem is a weird design decision by upstream `point_of_sale`. It kind of assumes that you're running `add_partners` on all partners (this is approximately what `load_new_partners()` does). We're not doing that.

Maybe we should remember `partner_write_date`'s value before editing it to the Unix epoch, and then reset `partner_write_date` it to its prior value after doing our thing?

carmenbianca:
I bump and ping @legalsylvain

legalsylvain (commented):
i have no idea if the conception is good. @ivantodorovich Any PoV on that PR ? 

Otherwise, I propose to merge this PR as is, setting an "Alpha" status in the manifest. 
This PR is for V12, and i'm sure that in V16, with all the refactoring, the situation will be different. (and I hope, better ;-))

em230418:
I would suggest to implement or use something like https://github.com/itpp-labs/pos-addons/tree/12.0/pos_partner_sync
It's depenency https://github.com/itpp-labs/sync-addons/tree/12.0/pos_longpolling

Briefly pos_partner_sync sychronizes certain fields realtime.

pos_refresh_customer/static/src/js/screens.js
```diff
@@ -0,0 +1,41 @@
...
+
+            if (client) {
+                this.pos.load_partners_by_ids([client.id]).then(function () {
+                    // Update the client list as well, because it may now be out
+                    // of date.
+                    var clientlist = self.gui.screen_instances["clientlist"];
+                    clientlist.partner_cache = new screens.DomCache();
+                    clientlist.render_list(
+                        self.pos.db.get_partners_sorted(1000)
+                    );
```
ivantodorovich, line 21: I'd move this block to `load_partners_by_ids`.
Seems you may want to re-render that list whenever the partners are loaded.

not blocking

ivantodorovich (approved):
> i have no idea if the conception is good. @ivantodorovich Any PoV on that PR ?


Seems a bit "hacky", but I couldn't find an alternative way of doing it. `12.0` is like that..
IMO it's ok like this

legalsylvain:
Hi @em230418.

> I would suggest to implement or use something like https://github.com/itpp-labs/pos-addons/tree/12.0/pos_partner_sync

Unfortunately, OCA policy doesn't allow to depend on non OCA modules.

em230418:
> Unfortunately, OCA policy doesn't allow to depend on non OCA modules.

@yelizariev, how do you look to idea to make pos_longpolling as OCA module?

yelizariev:
> > Unfortunately, OCA policy doesn't allow to depend on non OCA modules.
> 
> @yelizariev, how do you look to idea to make pos_longpolling as OCA module?

I'd be happy if someone do it

PR #788, opened 2022-04-29
[14.0][ADD] pos_disable_pricelist_selection

ivs-cetmix:
Disable pricelist selection button in the POS screen


legalsylvain:
Hi @isserver1. Thanks for sharing your work.
I don't understand the reason of this module. why not just set a single pricelist on the ``pos.config`` ?

ivs-cetmix:
> Hi @isserver1. Thanks for sharing your work. I don't understand the reason of this module. why not just set a single pricelist on the `pos.config` ?

Hi @legalsylvain  the reason is that sometimes you can have eg 10+ pricelist in POS and 20+ in the system.
Adding them back and forth takes time and not error prone.
Also keep in mind that if you select a pricelist which depends on another pricelist and the parent pricelist will not be in the list POS will trigger an error while loading

francesco-ooops:
> Also keep in mind that if you select a pricelist which depends on another pricelist and the parent pricelist will not be in the list POS will trigger an error while loading

yes, specifically the use case is: 

- We have a pricelist "POS Pricelist" with a line = compute price: formula, based on "Main Pricelist"
- in POS, we want user to use only "POS Pricelist", so we set Default Pricelist = "POS Pricelist" in pos config
- When POS interface loads, error: "An error occurred when loading product prices. Make sure all pricelists are available in the POS."

So this flag allows to have all necessary pricelist for formula calculation in POS (add them to "advanced pricelists" in POS setting), without allowing POS user to select pricelists not intended to be used in POS

francesco-ooops (approved):
Functional review ok!

legalsylvain (approved):
Thanks for the explanation, I get it.

In my opinion, some core feature is mission on pricelist like ``for_computation_only``, to avoid to select technical pricelist. (but that's another story).

francesco-ooops:
> Thanks for the explanation, I get it.
> 
> In my opinion, some core feature is mission on pricelist like ``for_computation_only``, to avoid to select technical pricelist. (but that's another story).
> 
> 

Can we merge then? :)

legalsylvain:
just tested on runbaot. works well.

however, is it normal to have the possilbity to use pricelist on partner view, in the PoS if the button is hidden ? functional question.

image

francesco-ooops:
> just tested on runbaot. works well.
> 
> however, is it normal to have the possilbity to use pricelist on partner view, in the PoS if the button is hidden ? functional question.
> 
> image
> 

Very good point. @isserver1 can you hide feature in partner too?

GabbasovDinar:
> > just tested on runbaot. works well.
> > however, is it normal to have the possilbity to use pricelist on partner view, in the PoS if the button is hidden ? functional question.
> > image
> 
> Very good point. @isserver1 can you hide feature in partner too?

done :)

francesco-ooops:
@GabbasovDinar looks good!
@legalsylvain can you take a look?

francesco-ooops:
@legalsylvain merge pls?

legalsylvain:
Well. Thinking again on that topic, i think OCA needs a module named sale-thechnical-pricelist to hide technical pricelist on back office, when editing partners or on sale orders ans invoices. 
Then developping a glue module pos-technical-pricelist. 

I'll make a PoC in a while. In the meantime :

/ocabot merge patch

legalsylvain:
@isserver1 : For your information : 
-  sale_pricelist_technical and account_pricelist_techical : https://github.com/OCA/sale-workflow/pull/2014
- pos_pricelist_technical : https://github.com/OCA/pos/pull/790

if you have time for a feedback, I would be interested.

ivs-cetmix:
> @isserver1 : For your information :
> 
> * sale_pricelist_technical and account_pricelist_techical : [12.0 add sale_technical_pricelist and account_invoice_technical_pricelist sale-workflow#2014](https://github.com/OCA/sale-workflow/pull/2014)
> * pos_pricelist_technical : [[ADD] new module_pos_pricelist_technical: #790](https://github.com/OCA/pos/pull/790)
> 
> if you have time for a feedback, I would be interested.

Looks like I'm a little bit late because both are merged already)

PR #790, opened 2022-05-12
[ADD] new module_pos_pricelist_technical:

legalsylvain:
Avoid to display in the point of sale pricelists that are used only to compute other pricelist.

based on https://github.com/OCA/sale-workflow/pull/2014

GRAP Task#477 : CC : @quentinDupont, @mounasb


pos_pricelist_technical/__manifest__.py
```diff
@@ -0,0 +1,34 @@
+# Copyright (C) 2014 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Point of Sale - Technical Pricelists",
+    "summary": "Prevent to display technical pricelist in the"
+    " point of sale front end UI.",
```
carmenbianca:
```suggestion
    "summary": "Prevent technical pricelists from being displayed in the Point of Sale front-end UI",
```

pos_pricelist_technical/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module extends the functionality of point of sale to prevent
+to select in the point of sale, technical pricelists that used for the
+computation of other pricelists, but that should not be selected.
```
carmenbianca:
```suggestion
This module extends the functionality of Point of Sale to prevent
technical pricelists that are used for the
computation of other pricelists from being selected.
```

carmenbianca (commented):
LGTM, although I'm a little unsure why the technical pricelist can be put in 'available pricelists' on the POS config. If that was disallowed, wouldn't that remove the need for most changes in the `src/[js,xml]`?

legalsylvain:
Hi @carmenbianca. Thanks for your review !

The discussion regarding what is a technical pricelist is deeply done here  https://github.com/OCA/sale-workflow/pull/2014

but in a few word, a technical pricelist is a pricelist used to compute other pricelist (that are available). So if the technical pricelist is not loaded in the PoS the computation of the other pricelist will fail. However selecting the technical pricelist doesn't make sense, so that is the reason to load but to hide technical pricelists in the PoS.
Let me know if I'm clear.

regards.

carmenbianca (approved):
Perfectly clear, thank you @legalsylvain :thumbsup:

pos_pricelist_technical/static/src/js/models.js
```diff
@@ -0,0 +1,24 @@
...
+    models.PosModel = models.PosModel.extend({
+        initialize: function(session, attributes) {
+            for (var i = 0 ; i < this.models.length; i++){
+                if (this.models[i].model == "product.pricelist") {
+                    this.models[i].fields.push("is_technical");
+                }
+            }
+            return _PosModelSuper.initialize.apply(this, arguments);
+        }
+    });
```
petrus-v: I would suggest something like this:

```suggestion
    var models = require("point_of_sale.models");

    models.load_fields("product.pricelist", ["is_technical"]);
```

petrus-v (approved):
I suggest an other syntax otherwise looks good to me

petrus-v:
... and of course you should take care CI complains ;)

legalsylvain:
Thanks a lot for your review, @petrus-v. I'll take a look on your comment.
If you have time, could you take a look on the other PR https://github.com/OCA/sale-workflow/pull/2014 ?
(The CI is red because dependencies are not available)

PR #791, opened 2022-05-14
[15.0] [MIG] `pos_event_sale` 🎫

ivantodorovich:
**New features:**

* Print event.registration receipts.
* Refunding a pos.order will cancel the event registrations.
* Full support for pricelists.
* Integrated with pos reports.


petrus-v (approved):
Looks so good with nice comments and tours !

Code review only

PR #793, opened 2022-05-17
[MIG] pos_order_mgmt: Migration to 14.0

pedropote:


ivantodorovich:
Hello, I think most of this module's features have already been implemented in core through the new Orders menu:

image

Are you sure we need this module in `14.0`+ ?

Moreover, if we do, I can't find the widget button so I'm guessing migration didn't went through correctly

pedropote:
@ivantodorovich  You're right. We were migrating from Odoo 13 to 14 and we thought this module wasn't migrated yet. It's my fault because I didn't realize that the desired functionality is covered with the core's new features. I'm sorry for the waste of time. I cancel de PR.
Thanks a lot for your apreciations.

PR #795, opened 2022-05-18
[13.0] [UPD] Use copier template and Github Actions

ivantodorovich:
This should fix random issues we're getting on 13.0 prs, like: https://github.com/OCA/pos/pull/708


sbidoul:
Looks like the frontend tests did run correctly in the container :partying_face:

legalsylvain:
Thanks a lot !

PR #768, opened 2022-03-22
[ADD] 14.0 pos_product_multi_barcode

PierrickBrun:
glue module for pos/product_multi_barcode

https://github.com/OCA/stock-logistics-barcode/tree/14.0/product_multi_barcode

PR #801, opened 2022-06-23
[14.0][FIX] pos_product_multi_barcode: enable feature to search for p…

ilyasProgrammer:
…roducts by multiple barcodes in search bar of POS.

Fix of the issue mentioned here https://github.com/OCA/pos/pull/768

### DIFF
--- a/pos_product_multi_barcode/static/src/js/db.js
+++ b/pos_product_multi_barcode/static/src/js/db.js
@@ -7,6 +7,14 @@ odoo.define("pos_product_multi_barcode.db", function (require) {
     models.load_fields("product.product", ["barcodes_json"]);
 
     PosDB.include({
+        _product_search_string: function (product) {
+            var str = this._super(product);
+            if (product.barcodes_json) {
+                const barcodes = JSON.parse(product.barcodes_json).join(",");
+                str = str.replace("\n", "|" + barcodes) + "\n";
+            }
+            return str;
+        },
         add_products: function (products) {
             var res = this._super(products);
             var self = this;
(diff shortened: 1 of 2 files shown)


pos_product_multi_barcode/static/src/js/db.js
```diff
@@ -7,6 +7,27 @@ odoo.define("pos_product_multi_barcode.db", function (require) {
     models.load_fields("product.product", ["barcodes_json"]);
 
     PosDB.include({
+        _product_search_string: function (product) {
+            var str = product.display_name;
```
legalsylvain: I think it's possible to overload this function to call super instead of copy the all function and avoid other modules to overload correctly the function. Dont you think ?

legalsylvain (changes requested):

pos_product_multi_barcode/static/src/js/db.js
ilyasProgrammer: yes right.

francesco-ooops (approved):
Functional review ok!

@ilyasProgrammer I don't think there's the need to update version, and if you can squash commits I think it could be ready for merge, thanks!

francesco-ooops:
@legalsylvain think it's good for merge?

legalsylvain:
> @ilyasProgrammer I don't think there's the need to update version, and if you can squash commits I think it could be ready for merge, thanks!

Indeed bumping version is done by people who merge with ocabot command.

legalsylvain (approved):
Thanks for the changes ! 

/ocabot merge patch

PR #804, opened 2022-07-01
[15.0] [IMP] `pos_event_sale`: Create only the missing `event.registrations`

ivantodorovich:
This allows to play nice with other modules, that may either be:

* Trying to modify the created registrations
* Creating registrations themselves
* Or even calling `_create_event_registrations` programatically.

It has no impact on functionality, though.


flotho (approved):
LGTM

PR #805, opened 2022-07-01
[15.0] [ADD] `pos_sale_pos_event_sale`: Glue module between `pos_sale` and `pos_event_sale`

ivantodorovich:
Glue module between `pos_event_sale` and `pos_sale`, to handle events when importing Sales Orders from the Point of Sale.


legalsylvain:
- ``pos_event_sale`` [depends](https://github.com/OCA/pos/blob/15.0/pos_event_sale/__manifest__.py#L14) on   ``event_sale`` and ``point_of_sale``
- ``event_sale`` [depends](https://github.com/odoo/odoo/blob/15.0/addons/event_sale/__manifest__.py) on ``event`` and ``sale_management`` (auto_installable)
- ``pos_sale`` [depends](https://github.com/odoo/odoo/blob/15.0/addons/pos_sale/__manifest__.py#L15) on ``point_of_sale`` and ``sale_management`` (auto_installable)

So if ``pos_event_sale`` is installed : ``event``, ``sale_management``, ``point_of_sale`` are installed.
So ``pos_sale`` is installed _for sure_.

That way no need to create an extra glue module between both module. You should move the code of the current glue module into ``pos_event_sale`` directly.

Or did I missed something ?

Kind regards.

That way, no need to overload ``_create_event_registrations`` function and it will reduce the code you propose.

ivantodorovich:
I thought about this myself, but glue modules can still be manually uninstalled. So I rather keep the dependency chain, no?

legalsylvain:
> glue modules can still be manually uninstalled

Well auto_installable module can be uninstalled, yes. or with some [funny module](https://github.com/OCA/server-tools/tree/14.0/module_change_auto_install) switched to not_auto_installable. ;-)

in fact there are two kind of auto_installable modules : 
A) glue module that make working two modules together. If the glue module is not installed, some feature will not work properly. (for example I think that uninstalling sale_stock is risky!)
B) promotional module that Odoo SA set to auto_install to make sure the wow effect is present for each customer. (like ``partner_autocomplete``, ``mail_bot``...)

-> pos_sale is a glue module. (A)

Two possibilities : 
1) keep the PR as it.
2) add an extra explicit dependency to ``pos_sale`` in the ``pos_event_sale`` manifest. That will reduce the complexity of the code, and you are not forcing people to use extra module. (because ``pos_sale`` should be installed by design).

As I just don't see a reason to not want to have ``pos_sale`` installed if ``pos_event_sale`` is installed, I'm in favor of the second option. (not a blocking point though.)

kind regards.

ivantodorovich:
Thanks! I rather keep it like this though.

* `pos_event_sale` is already too big, no need to add more complexity into it
* IMO the separation is clear. There's no real `sale.order` flow or model dependency in `pos_event_sale`, we just use the hability to set products and ticket prices on events from `event_sale`, but that's it.
* Odoo is already moving parts of `event_sale` to `event` (it moved event.tickets, partially, recently).. so I even see a future where `event_sale` dependency is dropped
* This new module on the other hand is meant to hold real `sale.order` <--> `pos.order` features, same as the core one

petrus-v (approved):

PR #808, opened 2022-07-07
[15.0] [ADD] `pos_event_sale_session` : Sell events with sessions on the PoS

ivantodorovich:
**Depends on:**
- https://github.com/OCA/event/pull/276
- https://github.com/OCA/event/pull/275

### DIFF
--- a/pos_event_sale_session/__init__.py
+++ b/pos_event_sale_session/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import reports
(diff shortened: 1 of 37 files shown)


pos_event_sale_session/static/src/js/models/EventTicket.js
```diff
@@ -0,0 +1,37 @@
...
+
+    models.EventTicket = models.EventTicket.extend({
+        /**
+         * @override
+         * @param {Order} options.session Limit count to the given EventSession
+         * @returns {Number} ordered quantity
+         */
+        getOrderedQuantity: function () {
+            /* eslint-disable no-param-reassign */
+            // TODO: Disabled for now. Re-enable when we implement session ticket limits
```
petrus-v: Maybe we can remove this file or implement it ? or if tickets limits require extra dependency do it in an other module ?!

ivantodorovich: Indeed. I removed it because this should rather be implemented in `event_sale_session` first, and it's not trivial

gregory-moka (approved):

PR #809, opened 2022-07-08
[14.0][ADD] pos_global_discount_in_line

ilyasProgrammer:
Include POS order discount in lines instead of adding discount product.


francesco-ooops (approved):
functional review ok 

from description: Include POS order discount in line instead of adding discount product.

---
this module works exactly like pos_discount, but instead of adding a "discount product", the value of discount is added in PoS order lines

pos_global_discount_in_line/models/pos_config.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    global_discount_in_line = fields.Boolean(
+        string="Add Discount in Line", default=False
```
ivantodorovich:
```suggestion
        string="Add Discount in Line"
```

it's already `False` by default

ivantodorovich (approved):

pos_global_discount_in_line/models/pos_config.py
ilyasProgrammer: It is null by default. At least from the point of database there is difference between null and false.

ivantodorovich: For boolean fields, the ORM will convert all falsy values to `False`: https://github.com/odoo/odoo/blob/939ffdfadd82789d8ad17245f48814270755b042/odoo/fields.py#L1210

No default value is the same as `default=False`, so it's just redundant.
In any case, I'm not blocking on this. Just a suggestion

francesco-ooops:
@OCA/pos-maintainers can we merge this?

ivantodorovich:
I believe we do.
Could you squash your commits before, please? 🙏🏻

francesco-ooops:
@ilyasProgrammer actually now that I check again order total is not updated correctly after applying discount ([video](https://recordit.co/9W120YcaJD))

francesco-ooops:
@ilyasProgrammer ok squash again 😎

ilyasProgrammer:
@ivantodorovich  Squashed. + upd to remove rounding done by pos_discount module in onClick().

francesco-ooops (approved):
Fix works well (functional review)

Decimals for discount value were not passed correctly to PoS order lines

@ivantodorovich can you take a look? thanks!

ivantodorovich:
LG 👍🏻 
Could you squash again, however? I still see two commits 😅

francesco-ooops:
@ivantodorovich should be good?

ivantodorovich:
indeed

/ocabot merge nobump

PR #812, opened 2022-07-11
[IMP] 14.0 pos_product_template: hint if variant exists

PierrickBrun:
Before: only show selected attributes

After: show selected attributes but also show "unavailable" attributes, the ones that do not have any active variant corresponding.


legalsylvain:
superseded by #814

PR #813, opened 2022-07-12
[12.0] [UPD] Use copier and update dotfiles

ivantodorovich:
Besides the fact that it's a good idea to follow the oca repo template, ideally I'd like to see if this fixes the issues in https://github.com/OCA/pos/pull/428


legalsylvain:
Technical PR. 
Let's go ahead.

/ocabot merge patch

PR #796, opened 2022-05-28
[ADD] add new module: pos_edit_order

GabbasovDinar:

PR #819, opened 2022-07-28
[FIX][DOC] pos_edit_order_line : correct reference to description image

legalsylvain:
fix : https://github.com/OCA/pos/pull/796#issuecomment-1197916762

FYI : @GabbasovDinar; @francesco-ooops

here is the correct way to reference images in the readme files.


legalsylvain:
trivial.

/ocabot merge patch

PR #820, opened 2022-07-29
[14.0][FIX][pos_margin] Hide margin in Front End if config is false

cvinh:
Before this fix, Margin is displayed on the frontend on the sale order screen even if the option is not selected in config


flotho (approved):
code review LGTM

legalsylvain:
/ocabot merge patch 

Thanks for this fix !

PR #823, opened 2022-08-04
[ADD][12.0] hw_posbox_ssl

thomaspaulb:


legalsylvain (approved):
Thanks ! Code review. No test.

thomaspaulb:
@legalsylvain Wow, you are quicker than Travis here :-)

cvinh:
Thanks for your PR
Does your module fix the issue of the self-generated certificate of the IoTBox (in case of https://x.x.x.x setup for the printers) ?
Actually, we have a workaround : we have to go everyday on the IoTBox webpage to accept the certificate but it's not optimal

thomaspaulb:
@cvinh If you talk about IoTBox, which image do you mean? This module is meant for the 12.0 POSbox image of which I am not aware it uses even a self-signed certificate, by default it just listens on HTTP only. This module changes it to HTTPS and makes it use a certificate of choice, and if you combine it with `acme.sh`, you can have the certificate auto-refreshed when it expires. So, you never have to do a manual action again.

For IOT box of Odoo Enterprise, Odoo SA offers some kind of phone-home to Odoo, which then offers the HTTPS link. (Similar to `ngrok`) So for that this module is not needed.

cvinh:
> Wow, you are quicker than Travis here :-)

Quicker and as good as Travis

cvinh:
> For IOT box of Odoo Enterprise, Odoo SA offers some kind of phone-home to Odoo, which then offers the HTTPS link. (Similar to `ngrok`) So for that this module is not needed.

We use the IoTBox with Odoo CE, we indeed have to accept the self certificate every day on the pos. We did not find other solution

thomaspaulb:
Ugh, we were using a custom image, and I didn't realize that all the [official POSbox/IOTbox images](https://nightly.odoo.com/master/iotbox/) images have NGINX installed, and so this won't work. It's better then to install the certificate directly in NGINX.

----

@cvinh The idea still works, but now it doesn't need a module, it just needs a variant of the following:

```
# Installation of acme.sh
sudo su -
mount -o remount,rw /
mount -o remount,rw /root_bypass_ramdisks
cd /root_bypass_ramdisks/root
git clone https://github.com/acmesh-official/acme.sh .acme.sh
# here: also install the below script in crontab
mount -o remount,ro /
mount -o remount,ro /root_bypass_ramdisks
```

```
# Recurring script to be called as root
export URL_FULL_DOMAIN="some.domain.tld"
export URL_DNS_PROVIDER="dns_gd"  # GoDaddy - many other providers are supported, see acme.sh docs
export GD_Key="....."
export GD_Secret="......."
export CERTIFICATE_PRIVKEY_FILE="/etc/ssl/certs/nginx-selfsigned.crt"
export CERTIFICATE_FULLCHAIN_FILE="/etc/ssl/private/nginx-selfsigned.key"
export ACME_WORK_DIRECTORY="/root_bypass_ramdisks/root/.acme.sh"

#### Don't change variables after this line ######

mount -o remount,rw /
mount -o remount,rw /root_bypass_ramdisks

$ACME_WORK_DIRECTORY/acme.sh \
--home $ACME_WORK_DIRECTORY \
--issue \
--dns $URL_DNS_PROVIDER \
-d $URL_FULL_DOMAIN \
--key-file $CERTIFICATE_PRIVKEY_FILE \
--fullchain-file $CERTIFICATE_FULLCHAIN_FILE

mount -o remount,ro /
mount -o remount,ro /root_bypass_ramdisks
```

PR #826, opened 2022-08-09
[15.0] Add module pos_lot_barcode

grindtildeath:
This module extends the point of sale functionalities to allow scanning a lot/serial number
using a barcode reader instead of having to enter it manually.

### DIFF
--- a/pos_lot_barcode/static/src/js/Screens/ProductScreen.js
+++ b/pos_lot_barcode/static/src/js/Screens/ProductScreen.js
@@ -1,0 +1,207 @@
+/*
+    Copyright 2022 Camptocamp SA
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+*/
+odoo.define("pos_lot_barcode.ProductScreen", function (require) {
+    "use strict";
+
+    const ProductScreen = require("point_of_sale.ProductScreen");
+    const Registries = require("point_of_sale.Registries");
+    const {useBarcodeReader} = require("point_of_sale.custom_hooks");
+    const {isConnectionError} = require("point_of_sale.utils");
+    const {_lt} = require("@web/core/l10n/translation");
+
+    const PosLotBarcodeProductScreen = (ProductScreen) =>
+        class extends ProductScreen {
+            constructor() {
+                super(...arguments);
+                this.scan_lots_active = true;
+                useBarcodeReader({
+                    lot: this._barcodeLotAction,
+                });
+            }
+            async _barcodeLotAction(code) {
+                // Do not do anything if lot scanning is not active
+                if (!this.scan_lots_active) return;
+                // Get the product according to lot barcode
+                const product = await this._getProductByLotBarcode(code);
+                // If we didn't get a product it must display a popup
+                if (!product) return;
+                // Get possible options not linked to lot selection
+                const options = await this._getAddLotProductOptions(product, code);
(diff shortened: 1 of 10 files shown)


pos_lot_barcode/static/src/js/edit_list_popup.js
```diff
@@ -0,0 +1,30 @@
+/*
+    Copyright 2022 Camptocamp SA
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+*/
+odoo.define("pos_lot_barcode.EditListPopup", function (require) {
+    "use strict";
+    
+    const EditListPopup = require("point_of_sale.EditListPopup");
+    const Registries = require("point_of_sale.Registries");
+    const { onChangeOrder, useBarcodeReader } = require('point_of_sale.custom_hooks');
```
ivantodorovich:
```suggestion
    const { useBarcodeReader } = require('point_of_sale.custom_hooks');
```
`onChangeOrder` isn't used

ivantodorovich (approved):
Minor comment, LG 👍🏻

ivantodorovich (approved):
LG 👍🏻 just maybe squash your commits so that it's ready for a merge

pos_lot_barcode/static/src/js/Screens/ProductScreen.js
```diff
@@ -0,0 +1,207 @@
...
+            }
+            async _barcodeLotAction(code) {
+                // Do not do anything if lot scanning is not active
+                if (!this.scan_lots_active) return;
+                // Get the product according to lot barcode
+                const product = await this._getProductByLotBarcode(code);
+                // If we didn't get a product it must display a popup
+                if (!product) return;
+                // Get possible options not linked to lot selection
+                const options = await this._getAddLotProductOptions(product, code);
```
i-vyshnevska, line 31: as here we expect serial code which differs from barcode should we try to replace it on product barcode? and maybe we will not need to override _getAddProductOptions then

grindtildeath: `_getAddProductOptions` is overridden to activate/deactivate the lot scanning from ProductScreen if we have the Lot Selection Popup open.
Then, I guess your question is if we could avoid reimplementing `_getAddProductOptions` into `_getAddLotProductOptions` ? That's what I would have preferred not having to do, but since the original function doesn't provide any hook to "inject" the proper lot number into the results from the Popup, I don't see how we could have done it.

i-vyshnevska: in this case, we can push SN to pack_lot_lines, and popup should be initialized correctly by the default implementation
there will be another issue, as pack_lot_lines is created if popup confirmed with Ok button so we will need additionally handle Cancel button as we already add product and/or SN to order lines

grindtildeath: But we don't want the popup to be displayed in case we did scan a serial number from the product screen. That's also why I did reimplement without this part.

i-vyshnevska: that is the reason why I suggest it, if you push SN to pack_lot_line the rest of the work like weight, and product configuration should be properly handled by the default implementation

well, opt to you, non-blocking

gurneyalex (approved):

PR #827, opened 2022-08-17
[15.0] Create module pos_lot_base

grindtildeath:
This module extends the point of sale functionalities to allow loading lot/serial numbers
into the POS session.

Please note that this module is a base to build other modules and doesn't provide any
features in itself.


pos_lot_base/models/pos_config.py
```diff
@@ -0,0 +1,56 @@
...
+            )
+               SELECT l.id
+                 FROM stock_production_lot l
+            LEFT JOIN product_product p ON l.product_id=p.id
+            LEFT JOIN product_template t ON p.product_tmpl_id=t.id
+            LEFT JOIN lm ON l.id=lm.lot_id
+                WHERE l.available_in_pos
+                    AND (l.company_id=%(company_id)s OR l.company_id IS NULL)
+                    AND p.id IN %(product_ids)s
+                    AND t.available_in_pos
```
i-vyshnevska, line 33: pos configuration has a warehouse that is used to take products, meaning not all lots will be available for sealing if they are in different warehouses

pos_lot_base/models/stock_production_lot.py
```diff
@@ -0,0 +1,12 @@
...
+from odoo import fields, models
+
+
+class StockProductionLot(models.Model):
+
+    _inherit = "stock.production.lot"
+
+    product_sale_ok = fields.Boolean(related="product_id.sale_ok")
+    product_available_in_pos = fields.Boolean(related="product_id.available_in_pos")
+    available_in_pos = fields.Boolean()
```
i-vyshnevska, line 12: this also can be covered with the convention of warehouse linked to pos

pos_lot_base/models/pos_config.py
```diff
@@ -0,0 +1,56 @@
...
+                        OR t.pos_categ_id=ANY(%(available_categ_ids)s)
+                    )
+             ORDER BY t.priority DESC,
+                      COALESCE(lm.date,l.write_date,p.write_date) DESC
+                LIMIT %(limit)s
+        """
+        params = {
+            "company_id": self.company_id.id,
+            "available_categ_ids": self.iface_available_categ_ids.mapped("id")
+            if self.iface_available_categ_ids
```
i-vyshnevska, line 46: this check is redundant if the product is added means you need to load lots here, too much knowledge of details of implementation

pos_lot_base/models/pos_config.py
```diff
@@ -0,0 +1,56 @@
...
+               SELECT l.id
+                 FROM stock_production_lot l
+            LEFT JOIN product_product p ON l.product_id=p.id
+            LEFT JOIN product_template t ON p.product_tmpl_id=t.id
+            LEFT JOIN lm ON l.id=lm.lot_id
+                WHERE l.available_in_pos
+                    AND (l.company_id=%(company_id)s OR l.company_id IS NULL)
+                    AND p.id IN %(product_ids)s
+                    AND t.available_in_pos
+                    AND t.sale_ok
```
i-vyshnevska, line 34:
```suggestion
```
no need to have if product requested here we just need to attach lots
I am not sure we even need to have product_template here

grindtildeath:
As discussed internally, loading all the lots into the POS is counter productive and we better rely on backend calls to make sure lot information is up to date.
We could still use this as an extension for barcode scanning or lot selection if required, but shouldn't use it as a dependency.

PR #828, opened 2022-08-22
[14.0][ADD] pos_product_template_configurator: Add new module

GabbasovDinar:


francesco-ooops (approved):
Functional review ok!
--- 

after [pos_product_template](https://github.com/OCA/pos/tree/14.0/pos_product_template) was merged for v14, we noticed issues on dbs with large numbers of products, so decided to develop another variant selection module beginning from odoo standard pos configurator. 

@legalsylvain @ivantodorovich could you take a look? Thanks!

SicurSam (approved):
LGTM

francesco-ooops:
@OCA/pos-maintainers could you review? thanks!

legalsylvain (changes requested):
1) Could you add 2 printscreen in the description, and a reference in the description.rst file ? 

2) if I understand correctly, 

without that module : 
image

with that module : 
image

That's right ? 

I don't understand exactly what is improved ? Could you explicit ? AFAIK, 
- you add a price difference field (+6,90€)
- you remove the image

thanks !

francesco-ooops:
@legalsylvain the main reason this module was developed was not to change specific features of pos_product_template (the flow was ok overall), but to make it work in a db with ~ 30k variants, as pos_product_template could not handle them correctly

First approach was to try to fix the original module, then we moved to an implmentation of the standard odoo pos configurator with storable/consumable products

The dependency from pos_product_template was added to avoid having conflicting modules in pos runboat

Please instruct us on how we should change readme in order to get merged. Thanks!

legalsylvain:
> Please instruct us on how we should change readme in order to get merged. Thanks! 

adding explicit text in readme.rst file, with what you explain here, and add images and reference in the description.rst is enough for me.

francesco-ooops:
> > Please instruct us on how we should change readme in order to get merged. Thanks!
> 
> adding explicit text in readme.rst file, with what you explain here, and add images and reference in the description.rst is enough for me.

Hi @legalsylvain , this was done, think it's good?

legalsylvain:
Indeed. OK for me.  @hpar any point of view ?

francesco-ooops:
@legalsylvain can you approve?

legalsylvain (approved):
@hparfr @ivantodorovich review welcome

ivantodorovich (approved):
LG 👍🏻

hparfr:
> we noticed issues on dbs with large numbers of products, so decided to develop another variant selection module beginning from odoo standard pos configurator.

Not normal, it has been tested with a db with lots of variants an sometime lot of of axes of variations. I'm bit far now from the project which required this module so I do not know.

pos_product_template_configurator/static/src/js/ppt.js
```diff
@@ -0,0 +1,118 @@
...
+                        };
+                        attributes_by_id[id].ptav[x.id] = x;
+                    });
+                    var product_attribute_by_id = this.env.pos.db
+                        .product_attribute_by_id;
+                    var product_attribute_value_by_id = this.env.pos.db
+                        .product_attribute_value_by_id;
+                    var attributes = Object.values(attributes_by_id).map((x) => {
+                        return _.extend({}, product_attribute_by_id[x.attribute[0]], {
+                            values: _.sortBy(Object.values(x.ptav), function (item) {
```
hparfr: you may use an arrow function here to be constistent with the rest of the file

pos_product_template_configurator/static/src/js/ppt.js
```diff
@@ -0,0 +1,118 @@
...
+                            ptav: {},
+                        };
+                        attributes_by_id[id].ptav[x.id] = x;
+                    });
+                    var product_attribute_by_id = this.env.pos.db
+                        .product_attribute_by_id;
+                    var product_attribute_value_by_id = this.env.pos.db
+                        .product_attribute_value_by_id;
+                    var attributes = Object.values(attributes_by_id).map((x) => {
+                        return _.extend({}, product_attribute_by_id[x.attribute[0]], {
```
hparfr, line 39: it there a good reason to use `_.extend` instead of `Object.assign` ? 

(some question for `_.sortBy`)

pos_product_template_configurator/static/src/js/ppt.js
```diff
@@ -0,0 +1,118 @@
...
+                                    template_id: product.product_tmpl_id,
+                                    products: products,
+                                });
+
+                                if (ret.confirmed)
+                                    return this._clickProduct({detail: ret.payload});
+                            }
+                        } else {
+                            this.showPopup("ErrorPopup", {
+                                title: this.env._t("Not found product."),
```
hparfr:
```suggestion
                                title: this.env._t("Product not found."),
```

hparfr (approved):
Code read.
Some minor remarks about the code write style.

pos_product_template_configurator/static/src/js/ppt.js
GabbasovDinar: Many places in Odoo source code use both `extend` and `Object.assign`  If you know the criteria in which cases what should be used would be appreciated. I use examples solely from Odoo source code

hparfr: _.extends is a dependency from UnderscoreJS (in odoo15 (released in 2021), it's still UnderscoreJS 1.8.2 from 2015 !).

I guess there is lot of old code in odoo code base and like in python code quality vary from one developer to another.

Today Object.assign is available in supported browsers, so there is no need to rely on _ for that. Underscore will be dropped at some point, so for new code it's usually better to write it in javascript directly. It's more future-proof and probably better in terms of perfs.

Rely on  _ is still valid in v14, and I don't request any re-write. I just ask if you have any good reasons to do so, and may be spark a new thinking for your next code to be written.

GabbasovDinar: Thank you for the explanation, I will keep it in mind!

pos_product_template_configurator/static/src/js/ppt.js
GabbasovDinar: done

pos_product_template_configurator/static/src/js/ppt.js
GabbasovDinar: done

francesco-ooops:
@OCA/pos-maintainers merge? :)

francesco-ooops:
@legalsylvain go? :)

francesco-ooops:
@hparfr have all your remarks been addressed correctly?

hparfr:
> @hparfr have all your remarks been addressed correctly?

Yes, thanks. I don't have merging rights on pos.

ivantodorovich:
Thanks 🙏🏻 !!

/ocabot merge nobump

PR #829, opened 2022-08-24
[15.0][MIG] pos_lot selection

i-vyshnevska:
related https://github.com/OCA/pos/pull/827


pos_lot_selection/static/src/js/popup/LotSelectorPopup.js
```diff
@@ -0,0 +1,35 @@
...
+    const Registries = require("point_of_sale.Registries");
+    const {_lt} = require("@web/core/l10n/translation");
+
+    class LotSelectorPopup extends AbstractAwaitablePopup {
+        constructor() {
+            super(...arguments);
+            // If there's a product, get lots available related to this product
+            this.slected_items = [];
+            if (this.props.product) {
+                this.lots = this.env.pos.db.lot_by_product_id[this.props.product.id];
```
grindtildeath: Can we please avoid loading lots in the POS session and do and RPC call to get the list of lots from the popup instead?

pos_lot_selection/static/src/js/popup/ProductScreen.js
```diff
@@ -0,0 +1,46 @@
...
+odoo.define("pos_lot_selection.ProductScreen", function (require) {
+    "use strict";
+
+    const ProductScreen = require("point_of_sale.ProductScreen");
+    const Registries = require("point_of_sale.Registries");
+    const NumberBuffer = require("point_of_sale.NumberBuffer");
+
+    const PosLotSaleProductScreen = (ProductScreen) =>
+        class extends ProductScreen {
+            async _clickProduct(event) {
```
grindtildeath: Overriding `_clickProduct` won't provide this module's feature if the product barcode was scanned.. I feel like it would be better to override `_getAddProductOptions` or even `showPopup` instead.

pos_lot_selection/static/src/js/popup/ProductScreen.js
```diff
@@ -0,0 +1,46 @@
...
+                const product = event.detail;
+                if (product.tracking !== "none") {
```
grindtildeath: With this condition, we're not calling `_getAddProductOptions` which is an issue if for example you have a product tracked by lot that needs to be weighed with an electronic scale.

pos_lot_selection/static/src/js/popup/LotSelectorPopup.js
cvinh: AFAIK, pos has been designed to work offline, if you call RPC in real time, wouldn't you break that feature ?

pos_lot_selection/static/src/js/popup/LotSelectorPopup.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const AbstractAwaitablePopup = require("point_of_sale.AbstractAwaitablePopup");
+    const Registries = require("point_of_sale.Registries");
+    const {_lt} = require("@web/core/l10n/translation");
+
+    class LotSelectorPopup extends AbstractAwaitablePopup {
+        constructor() {
+            super(...arguments);
+            // If there's a product, get lots available related to this product
+            this.slected_items = [];
```
cvinh: Why do you define slect_items ? I don't see that you use that later

pos_lot_selection/static/src/js/db.js
```diff
@@ -0,0 +1,32 @@
...
+        init: function () {
+            this._super.apply(this, arguments);
+            this.lot_by_product_id = {};
+        },
+        add_lots: function (lots) {
+            if (!lots instanceof Array) {
+                lots = [lots];
+            }
+            // We don't constrain lot selection to already sold products
+            // as pos terminal doing inventory records only on session closing
```
cvinh: As you noticed here https://github.com/OCA/pos/pull/827#discussion_r948780559 would it be better to filter lots to have only available in the linked warehouse ?

pos_lot_selection/i18n/es.po
```diff
@@ -1,27 +0,0 @@
...
-"MIME-Version: 1.0\n"
-"Content-Type: text/plain; charset=UTF-8\n"
-"Content-Transfer-Encoding: \n"
-"Plural-Forms: \n"
-
-#. module: pos_lot_selection
-#. openerp-web
-#: code:addons/pos_lot_selection/static/src/xml/pos.xml:12
-#, python-format
-msgid "Select a Serial/Lot Number"
```
simahawk: this should stay no?

pos_lot_selection/static/src/js/db.js
simahawk: is done in models.js

pos_lot_selection/static/src/js/popup/LotSelectorPopup.js
simahawk: Indeed. Moreover, now only the good ones are loaded, LG.

pos_lot_selection/static/src/js/popup/LotSelectorPopup.js
```diff
@@ -0,0 +1,35 @@
...
+            if (this.props.product) {
+                this.lots = this.env.pos.db.lot_by_product_id[this.props.product.id];
+            }
+        }
+        getPayload() {
+            return {
+                newArray: [Object.assign({}, {text: $("#new_select").val()})],
```
simahawk: I don't get why you should use Object.assign here...

pos_lot_selection/static/src/js/popup/ProductScreen.js
i-vyshnevska: I don't know if that is possible when the product and tracked and weighed, usually when it is tracked it is not divisible 
if so this need to be tasted with the electronic weights

pos_lot_selection/static/src/js/popup/LotSelectorPopup.js
grindtildeath: Even with POS being designed to work offline, it can still work offline without having to load the lots. It's just that you'd have to enter the number manually instead of having the possibility to select, not having a lot loaded is not blocking in any way, as if you wouldn't have a product loaded and would be unable to sell it.

The possibility to select is therefore nice to have but again it's not blocking if you are offline, whereas loading the lots in the POS means that we have to sync the lots if there are many movements, to make sure they are up to date, what could easily be avoided by not loading the lots in the POS.

Finally, please have a look at pos_coupon, and you'll notice that the POS is doing a RPC request when a coupon code is entered on the point of sale, and it's impossible to apply a coupon in offline mode. I guess the reason is more or less the same than here, where you prefer to make sure that you're online to validate a coupon that having to load and sync all the time to make sure it's valid...

pos_lot_selection/static/src/js/popup/ProductScreen.js
grindtildeath: It is not divisible in case of serial numbers, but there could be lots of cases where a good tracked by lot has to be weighed. Think of food or pharma industries...

grindtildeath:
@i-vyshnevska please check my last comoments as it's definitely blocking

i-vyshnevska:
@grindtildeath updated

pos_lot_selection/models/stock_production_lot.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+from odoo.tools import float_compare
+
+
+class ProductionLot(models.Model):
+    _inherit = "stock.production.lot"
+
+    available_on_pos = fields.Boolean(compute="_compute_available_on_pos", store=True)
```
grindtildeath: I guess we could get rid of this class by modifying the search from the POS to use the `product_qty` field? cf suggestion below.

pos_lot_selection/static/src/js/popup/ProductScreen.js
```diff
@@ -0,0 +1,57 @@
...
+                var lots = await this.rpc(
+                    {
+                        model: "stock.production.lot",
+                        method: "search_read",
+                        kwargs: {
+                            domain: [
+                                "&",
+                                "&",
+                                ["product_id", "=", product.id],
+                                ["available_on_pos", "=", true],
```
grindtildeath:
```suggestion
                                ["product_qty", ">", 0],
```

pos_lot_selection/static/src/js/popup/OrderWidget.js
```diff
@@ -0,0 +1,47 @@
...
+
+    const Registries = require('point_of_sale.Registries');
+    const OrderWidget = require('point_of_sale.OrderWidget');
+
+    
+    var CustomOrderWidget = (OrderWidget) => class extends OrderWidget{
+        async _editPackLotLines(event) {
+            const orderline = event.detail.orderline;
+            
+            this.env.session.lots = await this.get_lots(orderline.product, this.env.session);
```
grindtildeath: I'm not sure if the session object is the right place where to store the possible values. Maybe the POS DB? @ivantodorovich any better idea?

pos_lot_selection/static/src/js/popup/OrderWidget.js
```diff
@@ -0,0 +1,47 @@
...
+            return super._editPackLotLines(event);
+        }
+
+        async get_lots(product){
```
grindtildeath: I guess the best place where to define this function is in the POS DB, so that you could call it using `this.env.pos.db.get_lots` in both cases.

pos_lot_selection/static/src/js/popup/ProductScreen.js
```diff
@@ -0,0 +1,57 @@
...
+    const PosLotSaleProductScreen = (ProductScreen) =>
+        class extends ProductScreen {
+            async _getAddProductOptions(product, base_code) {
+                if (["serial", "lot"].includes(product.tracking)) {
+                    // Set lots to context as it one possible way to extract arguments
+                    // later in popup
+                    this.env.session.lots = await this.get_lots(product);
+                } else {
+                    // Cleanup
+                    this.env.session.product = false;
```
grindtildeath: Why is this needed? :thinking:

pos_lot_selection/static/src/js/popup/ProductScreen.js
```diff
@@ -0,0 +1,57 @@
...
+                }
+
+                return super._getAddProductOptions(product, (base_code = base_code));
```
grindtildeath: Shouldn't we reset the value of lots in the session (or in the DB cf comment above) ?
```suggestion
                res = super._getAddProductOptions(product, (base_code = base_code));
                this.env.session.lots = []
                return res
```

pos_lot_selection/static/src/js/popup/EditListPopup.js
```diff
@@ -0,0 +1,20 @@
...
+    "use strict";
+
+    const EditListPopup = require("point_of_sale.EditListPopup");
+    const Registries = require("point_of_sale.Registries");
+
+    const LotSelectEditListPopup = (EditListPopup) =>
+        class extends EditListPopup {
+            constructor() {
+                super(...arguments);
+                this.props.lots = this.env.session.lots;
```
grindtildeath: I guess we could do this to make sure the value is set only if we are on the `EditListPopup` for lot selection. (in case this Popup is used in another module)
```suggestion
                if (this.props.title === this.env._t("Lot/Serial Number(s) Required")) {
                    this.props.lots = this.env.session.lots;
                }
```

pos_lot_selection/static/src/xml/LotSelectorPopup.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2022 Camptocamp SA
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<templates id="template" xml:space="preserve">
+
+    <t t-inherit="point_of_sale.EditListPopup" t-inherit-mode="extension">
+
+        <xpath expr="//t[@t-foreach='state.array']" position="replace">
+            <t t-foreach="state.array" t-as="item" t-key="item._id">
+                <EditListInput item="item" lots="props.lots" />
```
grindtildeath: Shouldn't we do this only if `props.lots` is defined? To avoid messing with possible other usages of `EditListPopup`

pos_lot_selection/static/src/xml/LotSelectorPopup.xml
```diff
@@ -0,0 +1,38 @@
...
+            </t>
+        </xpath>
+    </t>
+
+
+    <t t-inherit="point_of_sale.EditListInput" t-inherit-mode="extension">
+
+        <xpath expr="//input" position="attributes">
+            <attribute name="list">prepared_lots</attribute>
```
grindtildeath, line 24: Same as above, maybe to all of this only if we have `props.lots` defined?

grindtildeath (commented):
Looks much better overall, thank you :+1: 

IMO there are still a few points to be improved, @ivantodorovich can you have a look please?

pos_lot_selection/static/src/js/popup/OrderWidget.js
```diff
@@ -0,0 +1,47 @@
...
+    const Registries = require('point_of_sale.Registries');
+    const OrderWidget = require('point_of_sale.OrderWidget');
+
+    
+    var CustomOrderWidget = (OrderWidget) => class extends OrderWidget{
+        async _editPackLotLines(event) {
+            const orderline = event.detail.orderline;
+            
+            this.env.session.lots = await this.get_lots(orderline.product, this.env.session);
+            return super._editPackLotLines(event);
```
grindtildeath: Forgot about this, but maybe we should "reset" the values of lots that are stored after the call to super (as I wrote for the call in `ProductScreen`)

pos_lot_selection/static/src/js/popup/ProductScreen.js
i-vyshnevska: product_qty is not a stored field so we need to use a filter here, having a boolean field looks like more explicit way

pos_lot_selection/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "version": "15.0.1.0.0",
+    "category": "Point of Sale",
+    "author": "Tecnativa, Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/pos",
+    "license": "AGPL-3",
+    "depends": [
+        "point_of_sale",
+    ],
+    "assets": {
+        "web.assets_qweb": ["pos_lot_selection/static/src/**/*"],
```
ivantodorovich:
```suggestion
        "web.assets_qweb": ["pos_lot_selection/static/src/xml/**/*.xml"],
```

We don't want to load everything here, just the xml files

pos_lot_selection/models/stock_production_lot.py
ivantodorovich: I agree this is not cool.

As it's a stored computed field, it will recompute everytime the quants quantities change, which is on every stock.move for this product.. it's totally unnecessary.

It's even worse when we consider the quants in the customer location, where all delivered lots will be linked to in the end. It won't scale well as the db grows.

I think you could improve this by making it NOT stored, and implement a search method instead.

Alternatively, and maybe even better.. consider exposing the list of lots in a public method instead, specifically built for the pos. This means, in js, you won't call `search_read` directly.. you'd call this method instead.

For example, it could be a method in the `product.product` model: `get_available_lots_for_pos`. This allows you to do some python processing/filtering before sending back the records.

pos_lot_selection/models/stock_production_lot.py
```diff
@@ -0,0 +1,18 @@
...
+
+    @api.depends("quant_ids.quantity")
+    def _compute_available_on_pos(self):
+        for lot in self:
+            lot.available_on_pos = (
+                float_compare(lot.product_qty, 0, precision_digits=2) > 0
```
ivantodorovich: FTR, it you are keeping this, `precision_digits` shouldn't be hardcoded.
It should be read from the product's uom

pos_lot_selection/static/src/js/popup/OrderWidget.js
ivantodorovich: I don't think we need to store them at all.
IMO the lots should be fetched from the server in the EditListPopup popup, before it's shown, and that's it

pos_lot_selection/static/src/js/popup/OrderWidget.js
i-vyshnevska: appeared db implementation doesn't have rpc component so I used PosModel instead

pos_lot_selection/static/src/js/popup/ProductScreen.js
i-vyshnevska: this will not work as it is async method

pos_lot_selection/static/src/js/popup/OrderWidget.js
i-vyshnevska: EditListPopup is synchronous so to be able to fetch products there we need to load lots on session start, that implementation was declined 
second, is that there is no product in EditListPopup, so I need somehow to send the product there, or override part of the method, but that brings another issue as default implementation can call a chain of popups to configure the product and get a weight and we have an issue because lot popup is in the middle of that chain

pos_lot_selection/static/src/xml/LotSelectorPopup.xml
i-vyshnevska: that is handled fine datalist will be empty

grindtildeath (approved):
To me looks much better right now, thank you :+1: 

Might still not be perfect but enough to be usable, so you get my approval.

ivantodorovich (approved):

PR #830, opened 2022-08-30
[14.0][NEW] pos_debranding

mileo:
Signed-off-by: Luis Felipe Mileo <mileo@kmee.com.br>


pos_debranding/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "maintainers": ["mileo"],
+    "depends": [
+        "point_of_sale",
+    ],
+    # Version 16 =D
+    # 'assets': {
+    #     'web.assets_qweb': [
+    #         'pos_debranding/static/src/xml/**/*',
+    #     ],
+    # },
```
chienandalu: This can go away
```suggestion
```

pos_debranding/static/src/xml/Chrome.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template" xml:space="preserve">
+
+    <t
+        t-name="Chrome"
+        t-inherit="point_of_sale.Chrome"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//img[hasclass('pos-logo')]" position="replace" />
```
chienandalu: I tend to prefer adding a `d-none` class to avoid breaking inheritance from any other module that could hook into that element.

chienandalu (changes requested):
First of all, thanks for your contribution @mileo !

Some remarks:

- Remove this folder and file:
image

pos_debranding/static/src/xml/Chrome.xml
mileo: d-none don´t work well then i use: 

```
<attribute name="style">display: none;</attribute>
```

pos_debranding/__manifest__.py
mileo: I migrate this module to version 16 first, sorry.

pos_debranding/static/src/xml/Chrome.xml
chienandalu: Looks lile `oe_hidden` is the one to go in PoS (https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/static/src/css/pos.css#L57) So it looks like BS4 is not available in the POS

mileo:
Hi Guys, 

This module hide the Odoo SA logo.

pos_debranding/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+To configure this module, you need to do nothing.
```
ivantodorovich, line 1: > To configure this module, you need to do nothing.

Then remove this file ;)

pos_debranding/static/src/xml/Chrome.xml
```diff
@@ -0,0 +1,15 @@
...
+<templates id="template" xml:space="preserve">
+
+    <t
+        t-name="Chrome"
+        t-inherit="point_of_sale.Chrome"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//img[hasclass('pos-logo')]" position="attributes">
+            <attribute name="style">display: none;</attribute>
```
ivantodorovich, line 11: Seems to me this could've been done with css only, no?

```scss
img.pos-logo {
    display: none;
}
```

Anyway, not blocking. Nitpicking here ^^
Although with the css approach you'd probably save some cpu ticks, as it's one less qweb template to process.

ivantodorovich (approved):

ivantodorovich:
To be consistent with other OCA modules that due the same (see https://github.com/OCA/server-brand), maybe rename it to `pos_odoo_debranding`.

I'm wondering if this is the right repo to hold this. It is a pos module, yes, but we have a specific repo for debrandings

PR #831, opened 2022-08-30
[14.0][NEW] Pos Show Clock

mileo:


pos_show_clock/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+To configure this module, you need to do nothing.
```
legalsylvain: Useless file if no configuration is needed.

pos_show_clock/static/src/js/ChromeWidgets/Clock.js
```diff
@@ -0,0 +1,40 @@
...
+        }
+        mounted() {
+            // V16
+            // onMounted() {
+            this.updateCurrentTime();
+        }
+        updateCurrentTime() {
+            var self = this;
+            const now = Date.now();
+            const next_update = 60050 - parseInt(moment(now).format("ss")) * 1000;
```
legalsylvain: Dont understand what is 60050 ? Could you add a comment ?

legalsylvain (commented):
Lignt code review. No test. Same comment as chienandalou regarding V16 comments.

Lgtm otherwise. Thanks ! Coule you add a printscreen in the description ?

pos_show_clock/readme/CONFIGURE.rst
mileo: > Lignt code review. No test. Same comment as chienandalou regarding V16 comments.
> 
> Lgtm otherwise. Thanks ! Coule you add a printscreen in the description ?

Can we keep the js comment of v16? Because i made this module to version 16 first to better understand OWL v2

pos_show_clock/static/src/js/ChromeWidgets/Clock.js
mileo: - 1 minute has 60000ms
- 60000 + 50(less then 100ms rule), is the time we can update the minute without the user noticing the difference when comparing it to another watch.

legalsylvain (approved):
Thanks for the changes. Lgtm. 
Some comment left in the manifest. 
Otherwise. +1

mileo:
> Thanks for the changes. Lgtm. Some comment left in the manifest. Otherwise. +1

Solved!

etobella (approved):
LGTM

pos_show_clock/static/src/js/ChromeWidgets/Clock.js
```diff
@@ -0,0 +1,40 @@
...
+            // 1 min = 60.000ms
+            // 60000 + 50 (less then 100ms rule)
+            // it's the time we can update the minute without the user noticing
+            // the difference when comparing it to another watch.
+            this.state.current_time = moment(now).format("HH:mm");
+            setTimeout(() => {
+                self.updateCurrentTime();
```
ivantodorovich:
```suggestion
                this.updateCurrentTime();
```

💡 tip: With arrow functions you can simply use `this`, and then you avoid the old `var self = this`

pos_show_clock/static/src/js/ChromeWidgets/Clock.js
```diff
@@ -0,0 +1,40 @@
...
+        updateCurrentTime() {
+            var self = this;
+            const now = Date.now();
+            const next_update = 60050 - parseInt(moment(now).format("ss")) * 1000;
+            // 1 min = 60.000ms
+            // 60000 + 50 (less then 100ms rule)
+            // it's the time we can update the minute without the user noticing
+            // the difference when comparing it to another watch.
+            this.state.current_time = moment(now).format("HH:mm");
+            setTimeout(() => {
```
ivantodorovich: I would consider setInverval rather than setTimeout.
It feels better for this use case.

I would also store the timer reference (`this.invervalReference = setInterval(...)`). You can do it when mounting the component (`mounted`), and then clear it when the component is unmounted to keep things clean (`clearInterval(this.intervalReference)`)

ivantodorovich (approved):
Some suggestions

Tested functionally 👍🏻

legalsylvain:
@mileo : could you take into acount minor @ivantodorovich remarks ? 
so that we can merge the module after, (and port it in more recent version). kind regards.

mileo:
I will finish it this week. Thank you for the remind.

mileo:
> @mileo : could you take into acount minor @ivantodorovich remarks ? so that we can merge the module after, (and port it in more recent version). kind regards.

Done!! =D

legalsylvain (approved):
Lgtm. Code review. No test. 
Thanks !

pos_show_clock/static/src/js/ChromeWidgets/Clock.js
```diff
@@ -0,0 +1,42 @@
...
+            this.invervalReference = setInterval(() => {
+                this.updateCurrentTime();
+            }, next_update);
+        }
+        unmount() {
+            clearInterval(this.intervalReference);
+        }
+        get time() {
+            return this.state.current_time;
+        }
```
ivantodorovich: Maybe I wasn't clear about setInterval. It's not like setTimeout, you only have to set it up once, and it will automatically run the callback on each interval, forever.. until you clearit. That's why I suggest to set it up when the widget is mounted, and then clearit when it's unmounted.

Some other things:
- you can't call `mounted` on `setup`. `mounted` is an official lifecycle hook for components, you need to let the owl framework call it when it's actually being mounted.
- the unmount hook is `willUnmount` (I always forget that name)

I took the liberty of explaining with code, here below.
I also changed a few things, like I remove the need to store the formated time in a state. I haven't tested it, but see if it works and if it makes sense for you ^^

I do not understand the 60050 rule, so I made it update every second (`1000` ms, see `refreshInterval`). Anyway, feel free to take what you want or simply ignore.

```suggestion
        mounted() {
            this.intervalRef = setInterval(() => this.render(), this.refreshInterval);
        }
        willUnmount() {
            if (this.intervalRef) {
                clearInterval(this.intervalRef);
            }
        }
        get refreshInterval() {
            return 1000;
        }
        get timeFormat() {
            return "HH:mm";
        }
        get time() {
            return moment(now).format(this.timeFormat);
        }
```

pos_show_clock/static/src/xml/ChromeWidgets/Clock.xml
```diff
@@ -0,0 +1,23 @@
...
+    <t
+        t-name="Chrome"
+        t-inherit="point_of_sale.Chrome"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//SyncNotification" position="before">
+            <Clock />
+        </xpath>
+    </t>
```
ivantodorovich:
```suggestion
    <t t-name="Clock" owl="1">
        <div class="oe_status time">
            <span>
                <span t-esc="time" class="time" />
            </span>
        </div>
    </t>

    <t
        t-name="Chrome"
        t-inherit="point_of_sale.Chrome"
        t-inherit-mode="extension"
        owl="1"
    >
        <xpath expr="//SyncNotification" position="before">
            <Clock t-if="!env.isMobile" />
        </xpath>
    </t>
```

Better to not render the component itself. That prevents the js code execution as well, as the component will never be mounted if it's on mobile.

ivantodorovich (changes requested):
Still pending previous comment x.x

felipezago:
> Still pending previous comment x.x

Done!

rousseldenis (approved):
Functional review

pos_show_clock/static/src/js/ChromeWidgets/Clock.js
```diff
@@ -0,0 +1,50 @@
...
+
+        mounted() {
+            this.intervalRef = setInterval(() => this.updateCurrentTime(), 1000);
+        }
+
+        willUnmount() {
+            if (this.intervalReference) {
+                clearInterval(this.intervalReference);
+            }
+        }
```
ivantodorovich: Looks much better now! Thanks!

I think there's a typo here, thuogh:

On `mounted` the interval reference is stored in `internalRef`.
On `willUnmount`, it's checking and clearing `intervalReference`.

felipezago: You're right, thanks! I just fixed it

ivantodorovich (approved):

ivantodorovich:
Thanks!!

/ocabot merge nobump

PR #833, opened 2022-08-31
[14.0][NEW] pos_customer_tree_view_vat

mileo:
Signed-off-by: Luis Felipe Mileo <mileo@kmee.com.br>


legalsylvain (commented):
I think the name is not explicit. (We dont know if its about customer or product) Something like poste_customer_tree_view_vat could be better. Dont you think ?

legalsylvain:
Naïve question : why do you need the vat of the customer on this screen ?

mileo:
> screen

To help the user search by VAT, for example:

image

mileo:
> I think the name is not explicit. (We dont know if its about customer or product) Something like poste_customer_tree_view_vat could be better. Dont you think ?

Renamed =D

ivantodorovich (approved):

PR #836, opened 2022-09-01
[14.0] pos_ask_vat (formerly pos_crm) 

mileo:
- [x] Search by TAX ID;
 - [x] Implement a better popup with usability for touch screens;
 - [x] Remove repeated code;
 - [x] Add a pos.config field for the user to say the best time to ask for customer data: at the beginning or at the end of the sale;
 - [x] Create a new partner if does not exist and the number is not a barcode;


rvalyi (commented):
@mileo something I dislike here is the name "pos_crm" kind of makes believe it depends on the CRM module while it is actually no the case. May be rename the module? pos_contact_info? any other idea?

mileo:
@ygcarvalh @felipezago could you please rename this module as proposed on https://github.com/OCA/l10n-brazil/pull/2565

And continue the work needed to merge the pull request.

lfdivino:
@mileo, @rvalyi, @ygcarvalh, @felipezago Changes done in the module.

mileo (commented):
LGTM, thanks guys for finishing the job.

AndreMarcos (approved):

rvalyi:
Hello, there are these lint warnings, would be cool to fix them:


Also there is no runboat link, could you push again so we can test it on Ruboat?

pos_ask_vat/models/pos_order.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    @api.model
+    def _order_fields(self, ui_order):
+        res = super(PosOrder, self)._order_fields(ui_order)
+        res["customer_tax_id"] = ui_order.get("customer_tax_id")
+        return res
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
```
rvalyi, line 24: wouldn't a compute method be more robust?

pos_ask_vat/static/src/js/Screens/OrderManagementControlPanel.js
```diff
@@ -0,0 +1,27 @@
...
+        "pos_reference",
+        "partner_id.display_name",
+        "date_order",
+        "customer_tax_id",
+    ];
+
+    const OrderManagementControlPanelCRM = (OrderManagementControlPanel) =>
+        class extends OrderManagementControlPanel {
+            get searchFields() {
+                return SEARCH_FIELDS;
```
rvalyi, line 17: I think it would be better to get the list from calling super and then append "customer_tax_id" to it. Otherwise it would conflict if another module also tries to extends the search fields.

pos_ask_vat/static/src/js/Misc/NumberBuffer.js
```diff
@@ -0,0 +1,95 @@
...
+ License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+*/
+
+odoo.define("pos_ask_vat.NumberBuffer", function (require) {
+    "use strict";
+
+    var NumberBuffer = require("point_of_sale.NumberBuffer");
+    const {Gui} = require("point_of_sale.Gui");
+
+    const _updateBuffer = (input) => {
```
rvalyi, line 12: as you are completely redefining this heavily used 70+ lines method, I think it would be better to explain why you did it and what you changed (possibly with explicit line comments). Otherwise it will be very hard to guess the side effects of this module and follow the Odoo updates...

pos_ask_vat/static/src/js/Misc/NumberBuffer.js
```diff
@@ -0,0 +1,95 @@
...
+            const currentBufferValue = self.state.buffer
+                ? parseFloat(self.state.buffer)
+                : 0;
+            self.state.buffer = self.component.env.pos.formatFixed(
+                inputValue + currentBufferValue
+            );
+        } else if (!isNaN(parseInt(input, 10))) {
+            if (isFirstInput) {
+                self.state.buffer = String(input);
+            } else if (self.state.buffer.length > 13) {
```
rvalyi, line 78: max length is 12 in the original code. Did you change it specifically for Brazil? Is this generic?

pos_ask_vat/static/src/js/Misc/NumberBuffer.js
rvalyi: if the goal is to change only the VAT field buffer, isn't there a cleaner way or may be check field name? I'm no Javascript expert, but I still think this redefinition is a bit dirty. cc @legalsylvain any idea?

PR #837, opened 2022-09-01
[ADD] pos_zticket: ZTicket generator

robinkeunen:
Hi there,

I'm submitting this work with the authorisation of @PierrickBrun. This was originally developed for vracoop.fr.

This module allows to generate a POS Z Ticket under .xls format. A Z Ticket is a summary of the daily point of sale operations.

I'm not sure wether this module fits here (under another name ?) or rather directly in oca/l10n-france ?

All the best


legalsylvain (commented):
Hi. Afaik, z ticket is a french word. Not sûre it is understandable in english. https://fr.mobiletransaction.org/x-de-caisse-et-z-de-caisse-definition/

Son i'm in favor of renaming the module. 

After that point. I dont see this module as a french one. So keep it in oca/pos is OK for me.

robinkeunen:
Hhhm there is more to refactoring to do than I thought (lots of French strings), I'm afraid I'll have to keep it as is in our incubator repos for the time being. We may adapt the code in a future migration. For those interested, the code is here : https://github.com/coopiteasy/cie-pos/pull/22

PR #841, opened 2022-09-22
[14.0][NEW] pos_sale_order_load

ygcarvalh:
Currently, this functionality already exists in Odoo 15. What it does is that it is possible to import a sales order into an open POS session and fill an order with the same items.

As it is something that in Odoo 15, is part of the `pos_sale` module itself, I created a module that inherits from this one and, in turn, does this kind of import to the frontend.


legalsylvain:
Hi. Thanks a lot for sharing your work.

A similar feature exists in a module named ``pos_picking_load`` (https://github.com/OCA/pos/tree/12.0/pos_picking_load) I think that the name of your module should be aligned = ``pos_sale_load`` or ``pos_sale_order_load`` ("import" is reserved in odoo to the action of importing data from Excel / csv files). What do you think ?

Thanks.

legalsylvain:
Hi. Thanks a lot for sharing your work.

A similar feature exists in a module named ``pos_picking_load`` (https://github.com/OCA/pos/tree/12.0/pos_picking_load) I think that the name of your module should be aligned = ``pos_sale_load`` or ``pos_sale_order_load`` ("import" is reserved in odoo to the action of importing data from Excel / csv files). What do you think ?

Thanks.

ygcarvalh:
Hey, thanks for the review, I really appreciate it.

I took a look at this module that you told me and it is very similar, even if it is picking from the `stock.picking` model and it will not always have a `sale.order` associated with it.

I didn't know about this more specific concept around importing to Odoo. I think both cases for module name are good, I'll make the changes.

Thank you very much.

legalsylvain:
Hi. 
just tested on runboat. partial functional review / no code review.

Thanks a lot for your contribution ! Some remarks below.

1) I created a new sale order in quotation state with other existing quotations.

image

When I go in the PoS, only my quotation is displayed, not the others. I read the code, but I don't understand why. Could you explain me ? thanks ! 

image


2) the order I created contains product not available in PoS. When I click on the button "settle the order", a warning is displayed to ask me to load the product : OK. But when I confirm, an error not handled is raised : 

image

3) when I select the sale.order, the PoS doesn't select the customer associated to the sale order. I think it could be done to improve usability. Also (maybe I missed something), but I didn't see any reference to the SO in the PoS, once the order is loaded.

4) when i pay my pos.order and I close the session, my initial sale.order is not marked as invoiced. As a result, it is possible to invoice again the a paid order. I think something is wrong in the design. Don't you think ? 
(I don't know exactly what should be done...)

image

ygcarvalh:
Hi @legalsylvain, sorry for the delayed response. Regarding the points mentioned:

1. I had the same problem while testing. In this case it is the `domain` which is related to the company currency. As the demo orders are in Euro and the company is Dollar, it ends up not showing the other sales orders.

2. I didn't catch this error, I'll take a look to see what problem may have happened. Some error handling is already native in Odoo 15 - many related to the issue of connection to the back - and I had to adapt them.

3. In this case, the correct thing would be to update the customer button with this loading of sales order info, I'll take a look at what may happened.

4. In the POS, it doesn't really have any reference to the original sales order in a visual way, it is referenced in the POS sales order line object - something that is also done in the backend. If I'm not mistaken, there is a line view where you can see which sales orders are linked to that `pos.order.line`.

5. This invoice creation is a flag that is set in `pos.config`. As I remember the flow, without this flag selected, it will only confirm the sales order.

I'll run some tests locally with this branch and see if i can catch the erros that were found. Also see these points I've added - like creating the invoice and sales order reference on the line. I will get back to you asap.

Thanks.

mileo:
@ygcarvalh could you please change the author of the first commit as said in the guideline:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#backporting-odoo-modules

elvise:
@ygcarvalh great PR!

elvise:
@ygcarvalh Do you have any plans to fix this PR?

elvise:
@ygcarvalh can you rebase ?

sbidoul:
Should this be closed, now that https://github.com/OCA/pos/pull/1142 is merged?

chienandalu:
Closing as it was completed in 

- #1142

PR #843, opened 2022-09-23
[ADD] 12.0 - pos_cashback_warning

legalsylvain:
This modules extends the Point of sale functionnalies to add a warning if confirming an order generate a cashback.

By default, if the payment lines of an Odoo PoS order generate change, the journal for the change will be the Cash journal.
That is ok if the original payment line is a Cash payment, but if it's a payment by credit card, check, meal voucher, etc. confirming the transaction will generate a cashback.

This module simply add a message in such case.

pos_cashback_warning

**Note** : The cashback is an operation that consists, for a merchant to collect more money in check, credit card, etc. than the amount of the sale and give change in cash.

CC : @ivantodorovich, @carmenbianca, @robinkeunen, @quentinDupont


pos_cashback_warning/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+        "point_of_sale",
+    ],
+    "data": [
+        "views/templates.xml",
+    ],
+    "qweb": [
+        'static/src/xml/pos.xml',
+    ],
+    "images": [
+        "static/description/pos_cashback_warning.png",
```
ivantodorovich: Question: does this `images` key do something? 👀

pos_cashback_warning/static/src/js/screens.js
```diff
@@ -0,0 +1,47 @@
...
+                this.extradue_cashback =  - (total_cash_received + extradue);
+            }
+            else {
+                this.extradue_cashback = 0.0;
+            }
+            this._super();
+        },
+
+    });
+});
```
ivantodorovich, line 47:
```suggestion

    return screens;
});
```

Always a good idea to return the object your extending, so that other modules can easily override if needed

pos_cashback_warning/static/src/js/screens.js
```diff
@@ -0,0 +1,47 @@
...
+                // Note Copy from original function render_paymentlines
+                // to recompute extradue value
+                var lines = order.get_paymentlines();
+                var due   = order.get_due();
+                if (due && lines.length  && due !== order.get_due(lines[lines.length-1])) {
+                    extradue = due;
+                }
+                _.each(lines, function (line) {
+                    console.log("=> line.cashregister.journal.type", line.cashregister.journal.type);
+                    console.log("=> line.amount", line.amount);
```
ivantodorovich:
```suggestion
```

ivantodorovich (changes requested):

pos_cashback_warning/__manifest__.py
legalsylvain: AFAIK, for the apps.odoo.com, the images is used to display the image in the appstore.

See V13 doc : https://www.odoo.com/documentation/13.0/developer/reference/addons/module.html?highlight=manifest

but in master, the key seems to be removed : https://www.odoo.com/documentation/15.0/developer/reference/backend/module.html?highlight=manifest 

I'll remove it.

thanks for the review.

ivantodorovich (approved):

legalsylvain:
Hi @github-actions. Thanks for your friendly reminder ! 
@ivantodorovich : I applied your suggestions.

regards.

legalsylvain:
Used in production since 6 monthes without problem.

/ocabot merge patch

PR #844, opened 2022-09-27
[ADD] Module pos_stock_mrp

gabrielcardoso21:


pos_stock_mrp/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module extends the functionality of Point of Sale (point_of_sale) to display the current time on the main screen.
```
rvalyi: seems some bad copy paste from another module here.

PR #846, opened 2022-09-28
[15.0] [FIX+IMP] `pos_lot_selection`

ivantodorovich:
This PR achieves the following things, each on a separate commit:

1. Use the proper module name in the js definition. This made the module incompatible with `pos_lot_barcode` because of the overlapping js definitions.
2. Fix an issue where this module doesn't work for Point of Sale users with minimal access rights.
3. Modernize js code
4. Simplify view definitions, mainly to avoid the `replace` operation that reduces modules compatibility

---

C2C ref: BSATE-58


i-vyshnevska (approved):

grindtildeath (approved):
LGTM :+1:

Issue #839, opened 2022-09-16
[13.0][pos_order_mgmt] Wrong format date and time in pos receipt when reprinting


javierjcf:
## pos_order_mgmt
When I do a ticket I get the date and hour in my timezone. When I reprint that order is in american format and in UTC.

When ticket is printed firs time, odoo computes a field called formatted_validation_date with new Date(), whitch includes timezone.
When reprint, we get order_data.date_order un UTC, so when in the module you do:

```js
 // Get Date
            if (["print"].indexOf(action) !== -1) {
                order.formatted_validation_date = moment(order_data.date_order).format(
                    "YYYY-MM-DD HH:mm:ss"
                );
```

The date does not respect original format, in my case dd/MM/YYYY, and is not converted from UTC to user timezone.

Somethgn like this solves tge issue I think:

```js
    var local_date_hour = new Date(order_data.date_order + ' UTC').toISOString()
    order.formatted_validation_date = moment(local_date_hour).format(
        "DD/MM/YYYY HH:mm:ss"
    );
```

When the module calls the server, doing self.date_order I get UTC date, so in javascript is in UTC too. It should be converted to get the same date and same format than original receipt

chienandalu:
Hi, @javierjcf Can you make the PR?

javierjcf:
Hi @chienandalu.

New PR: #850

PR #850, opened 2022-10-04
[13.0]pos_prder_mgmt: fix format time when reprint ticket

javierjcf:
Fix datetime format when reprint ticket in order to fix issue:
https://github.com/OCA/pos/issues/839

### DIFF
--- a/pos_order_mgmt/static/src/js/widgets.js
+++ b/pos_order_mgmt/static/src/js/widgets.js
@@ -262,8 +262,11 @@ odoo.define("pos_order_mgmt.widgets", function(require) {
 
             // Get Date
             if (["print"].indexOf(action) !== -1) {
-                order.formatted_validation_date = moment(order_data.date_order).format(
-                    "YYYY-MM-DD HH:mm:ss"
+                var local_date_hour = new Date(
+                    order_data.date_order + " UTC"
+                ).toISOString();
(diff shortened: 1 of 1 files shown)


pos_order_mgmt/static/src/js/widgets.js
```diff
@@ -262,8 +262,11 @@ odoo.define("pos_order_mgmt.widgets", function(require) {
 
             // Get Date
             if (["print"].indexOf(action) !== -1) {
-                order.formatted_validation_date = moment(order_data.date_order).format(
-                    "YYYY-MM-DD HH:mm:ss"
+                var local_date_hour = new Date(
+                    order_data.date_order + " UTC"
+                ).toISOString();
+                order.formatted_validation_date = moment(local_date_hour).format(
+                    "DD/MM/YYYY HH:mm:ss"
```
ivantodorovich, line 269: I think the format shouldn't be hardcoded. It really depends on the language.
You could try something like this:

```js
const time = require("web.time");

// ...

formatted_date = moment(date).format(time.getLangDatetimeFormat());
```

As for the tz conversion, I recommend leveraging `moment` capabilities using the `moment.utc` and `moment.local` functions.


As a result you'd have something like this:

```js
    const localOrderDate = moment.utc(order_data.date_order).local();
    order.formatted_validation_date = localOrderDate.format(time.getLangDatetimeFormat());
```

PR #683, opened 2021-06-28
[13.0][MIG]pos_order_to_sale_order, standard migration

manuelregidor:
Standard migration

PR #854, opened 2022-10-07
16.0 mig pos order to sale order

legalsylvain:
### Description 

This module extends the functionality of point of sale to allow sale orders
creation from the Point of Sale.

In the POS UI, button has been added to create a sale order and discard
the current POS order.

This module is usefull in many cases, for exemple :

* take orders with a very simple interface

* if you have some customers that come every day in your shop, but want to
  have a unique invoice at the end of the month. With that module, you can
  create a sale order and deliver products every time to keep your stock value
  correct, and to create a unique invoice, when you want.

### Usage

* Open your Point of sale
* Create a new order and select products
* Select a customer

* Click on the "Create Order" button

pos_frontend_button

Four options are available (depending on the PoS settings):

* **Create a draft Order**
  A new sale order in a draft mode will be created that can be changed later.

* **Create a Confirmed Order**
  A new sale order will be created and confirmed.

* **Create Delivered Sale Order**
  A new sale order will be created and confirmed.
  the associated picking will be marked as delivered.

* **Create Invoiced Sale Order**
  A new sale order will be created and confirmed.
  the associated picking will be marked as delivered.
  An invoice will be created and confirmed.

pos_frontend_popup


legalsylvain:
Added tests as proposed by @ivantodorovich [here](https://github.com/OCA/pos/pull/683).

PierrickBrun (approved):
code LGTM + tested in dev env

ivantodorovich (approved):

luke-stdev001:
@legalsylvain ,

This module looks fantastic! One quick issue i've encountered when attempting to test this module (with these boolean fields all ticked in pos config: https://github.com/OCA/pos/blob/16.0/pos_order_to_sale_order/views/view_res_config_settings.xml#L14-L46), the button (https://github.com/OCA/pos/blob/16.0/pos_order_to_sale_order/static/src/xml/CreateOrderButton.xml#L5-L9) does not appear in the POS UI, and in addition to this it removes many of the default buttons from the core POS UI. Please see screenshot as follows:

Screen Shot 2022-10-18 at 2 58 09 pm

PR #856, opened 2022-10-13
[14.0][ADD] pos_report_engine

luismalta:
# This PR
This module adds a reporting engine to the POS front-end, aiming to facilitate and standardize the insertion of reports accessible by the POS operator.

# Adding new reports
To add a new report, just extend the PosReportEngine component and use the "addReport()" interface, passing the report object with the following format:

```
report: {
      id: string -> report id
      name: string -> Report button name
      reportGeneratorHandle: Function -> Function responsible for generating and returning the report
      reportInputs: list -> List of data input component objects for report generation

  }

  report_input_object: {
      id: string -> ID used to capture input value
      component: class -> Input component class
      label: string -> Input field label
  }
```
A usage example file can be found at examples/exampleReports.js

The module already introduces two types of data inputs for the report (datePickerInput and textInput ), but it is possible to create the component itself for a new data input.

cc: @mileo


mileo:
@luismalta could you move the session sumary to the report screen?

image

PR #859, opened 2022-10-17
[16.0][ADD] new module  pos_discount_all

legalsylvain:
### Description 

This module extends the functionality of point of sale module, to display on the PoS ticket, the amount of the discount applied for this sale. Contrary to the logic of Odoo and the OCA
(https://github.com/OCA/sale-workflow/tree/14.0/sale_discount_display_amount)
the amount of the discount is the difference between the theoretical sale with the default price list and the actual sale amount.

So it take into accounts:

- explicit discount set on pos.order.line. (as for odoo ``sale`` module)
- fixed price set on pos.order.line
- discount generated by specific pricelist

Also the module ignores in the computation of undiscounted amount lines, the
lines with a 'Discount Product'.

### Usage

Open your point of Sale

Make an order. for exemple,
- select the product 'Conference chair (39.40$)'
- select the pricelist -10%
- add a discount of 1$

The total discount is 1$ + 10% * 39.40 = 4.94$

order_summary

Note, the discount displayed on the ticket is updated, to take into account all the discount.

pos_receipt

### Extra configuration

* Go to "Point of Sale > Products"
* Create or edit your discount products
* Check the box "Is a Discount"

product_template_form

CC : @PierrickBrun


PierrickBrun (approved):

pos_discount_all/static/src/js/models.js
```diff
@@ -0,0 +1,101 @@
...
+    const PosDiscountAllOrder = (Order) =>
+        // eslint-disable-next-line no-shadow
+        class PosDiscountAllOrder extends Order {
+            // eslint-disable-line no-shadow
+            // @override
+            _get_ignored_product_ids_total_discount() {
+                const productIds = super._get_ignored_product_ids_total_discount(
+                    ...arguments
+                );
+                _.map(this.pos.db.product_by_id, function (product) {
```
hparfr, line 24:
```suggestion
                this.pos.db.product_by_id.forEach(function (product) {
```

pos_discount_all/static/src/js/models.js
```diff
@@ -0,0 +1,101 @@
...
+                        total_excluded: 0.0,
+                        total_included: 0.0,
+                    };
+                }
+                var price_unit_without_any_discount = product.get_price(
+                    this.pos.default_pricelist,
+                    this.get_quantity()
+                );
+                var taxes_ids = this.tax_ids || product.taxes_id;
+                taxes_ids = _.filter(taxes_ids, (t) => t in this.pos.taxes_by_id);
```
hparfr, line 82:
```suggestion
                taxes_ids = taxes_ids.filter( (t) => t in this.pos.taxes_by_id);
```

hparfr (approved):
code review
minor remarks about removing calls to a legacy js lib.

ivantodorovich (approved):
Agree w/ @hparfr although not blocking

Maybe you want to change it, or not @legalsylvain . I let you merge

ivantodorovich:
I suggest trying to come up with a more specific name, ideally something that says `show` in it?
`pos_show_total_discount` maybe? or `pos_display_discount_amount` ?

legalsylvain:
@ivantodorovich i'm not sure regarding the name. Odoo pos still compute discount but not all discounts. This module mainly change the way the computation is done. So for me "discount all" is quite OK. Dont you think ?

ivantodorovich:
I see. Well.. "discount all" can mean many many things.
Like, you could have a module with a button that applies a discount on all lines, for example [?]

I'm not sure what the right name should be, though. I just think it's a bit vague at the moment.

How about `pos_display_effective_discount` ?

legalsylvain:
hi @ivantodorovich. Happy new years ! 

> I see. Well.. "discount all" can mean many many things.
> Like, you could have a module with a button that applies a discount on all lines, for example [?]

If exists, that module will be named pos_global_discount (as other OCA module for sale.order for exemple).
but the feature exists natively in odoo for pos.order...

> I'm not sure what the right name should be, though. I just think it's a bit vague at the moment.

> How about pos_display_effective_discount ?

Well, in my opinion, the objective of this module is to take into account all kind of discounts. So the name is quite ok for me ! But I don't speak very well in english, so I agree to apply changes proposed by reviewers.
Propose to wait a little bit other point of views, and if ``pos_display_effective_discount`` or other is promoted, I'll do the  change.
@hparfr : do you think the same ? 

regards !

ivantodorovich:
Hello @legalsylvain ! Happy new year to you too !! 🍾 🎉 

As far as I'm concerned, it was a totally non-blocking suggestion so feel free to ignore it. I already approved it 😉

PR #860, opened 2022-10-17
[16.0][ADD] new module ``pos_membership``

legalsylvain:
This module extends the functionality of point of sale module, to add informations related to Odoo ``membership`` module.

For instance :

- It displays the field 'Current Membership Status' of the partners in the point of sale screen. (partner list and partner form view)

**Partner List View**

pos_ui_partner_tree

**Partner Form View**

pos_ui_partner_form

- it raises an error, if cashier try to sell a membership product, without having selected the 'Invoice' option.

pos_warning_sell_membership_product


PaulGoubert:
Ok, RAS

pos_membership/static/src/xml/PartnerLine.xml
```diff
@@ -0,0 +1,25 @@
...
+-->
+<templates id="template" xml:space="preserve">
+
+    <t
+        t-name="PartnerLine"
+        t-inherit="point_of_sale.PartnerLine"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//button[@class='edit-partner-button']" position="after">
```
ivantodorovich, line 15: In terms of UX, isn't it better to display the membership status in a new column, rather than below the "Details" button?


pos_ui_partner_tree



// not blocking

ivantodorovich (approved):

pos_membership/static/src/xml/PartnerLine.xml
legalsylvain: I remember that I tried, but was not a better choice. (don't remember exactly why).
propose to merge it as it. If people has better implementation, totally open to accept it.

PierrickBrun (approved):

PR #862, opened 2022-10-18
[12.0][FIX] pos_payment_method_cashdro: floating point quirks

chienandalu:
Overcome floating point issues: https://stackoverflow.com/a/588014

For expample 2.3 * 100 ->  229.99999999999997

cc @Tecnativa TT39975

please review @CarlosRoca13

### DIFF
--- a/pos_payment_method_cashdro/static/src/js/screens.js
+++ b/pos_payment_method_cashdro/static/src/js/screens.js
@@ -68,7 +68,7 @@ odoo.define("pos_payment_method_cashdro.screens", function (require) {
             // Cashdro treats decimals as positions in an integer we also have
             // to deal with floating point computing to avoid decimals at the
             // end or the drawer will reject our request.
-            var amount = parseInt(order.get_due(payment_line).toFixed(2) * 100);
+            var amount = Math.round(order.get_due(payment_line) * 100);
             var url = this._cashdro_payment_url({"amount": amount});
             var operation_id = "";
             var request = this._cashdro_request(url)


pos_payment_method_cashdro/static/src/js/screens.js
```diff
@@ -68,7 +68,7 @@ odoo.define("pos_payment_method_cashdro.screens", function (require) {
             // Cashdro treats decimals as positions in an integer we also have
             // to deal with floating point computing to avoid decimals at the
             // end or the drawer will reject our request.
-            var amount = parseInt(order.get_due(payment_line).toFixed(2) * 100);
+            var amount = Math.round(order.get_due(payment_line) * 100);
```
pedrobaeza, line 71: No 2 decimals indication? Or is it the default?

chienandalu: It doesn't have the same meaning as python `round()`.  From mozdev (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round):

> The `Math.round()` function returns the value of a number **rounded to the nearest integer**.

So for a price of **2.30**, the Cashdro expects an input of **230**, but in JS: **2.3 * 100 = 229.99999999999997** (you can test it in your browser :slightly_smiling_face: ) due to floating point operations accuracy.

pedrobaeza: OK then.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #866, opened 2022-10-23
14.0 imp pos access right

hitrosol:
adding group_delete_item to prevent user to delete the order item

### DIFF
--- a/pos_access_right/static/src/xml/NumpadWidget.xml
+++ b/pos_access_right/static/src/xml/NumpadWidget.xml
@@ -13,8 +13,12 @@
         >
             <attribute
                 name="t-att-class"
+                t-translation="off"
             >{'disabled-mode': !hasMinusControlRights}</attribute>
-            <attribute name="t-att-disabled">!hasMinusControlRights</attribute>
+            <attribute
+                name="t-att-disabled"
+                t-translation="off"
+            >!hasMinusControlRights</attribute>
         </xpath>
 
         <xpath
@@ -23,8 +27,12 @@
         >
             <attribute
                 name="t-att-class"
+                t-translation="off"
             >{'disabled-mode': !hasDeleteItemControlRights}</attribute>
-            <attribute name="t-att-disabled">!hasDeleteItemControlRights</attribute>
+            <attribute
+                name="t-att-disabled"
+                t-translation="off"
+            >!hasDeleteItemControlRights</attribute>
         </xpath>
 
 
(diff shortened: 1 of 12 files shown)


pos_product_template_configurator/i18n/it.po
```diff
@@ -6,31 +6,33 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 14.0\n"
 "Report-Msgid-Bugs-To: \n"
-"Last-Translator: Automatically generated\n"
+"PO-Revision-Date: 2022-10-21 08:21+0000\n"
+"Last-Translator: Francesco Foresti <francesco.foresti@ooops404.com>\n"
```
ivantodorovich, line 10: This file is from `pos_product_template_configurator` module. It should have been changed on this PR

pos_access_right/static/src/xml/NumpadWidget.xml
```diff
@@ -16,6 +16,19 @@
...
         </xpath>
+
+        <xpath
+            expr="div[hasclass('numpad')]//button[hasclass('numpad-backspace')]"
+            position="attributes"
+        >
+            <attribute
+                name="t-att-class"
+            >{'disabled-mode': !hasDeleteItemControlRights}</attribute>
+            <attribute name="t-att-disabled">!hasDeleteItemControlRights</attribute>
```
ivantodorovich:
```suggestion
            <attribute
                name="t-att-class"
                t-translation="off"
            >{'disabled-mode': !hasDeleteItemControlRights}</attribute>
            <attribute name="t-att-disabled" t-translation="off">!hasDeleteItemControlRights</attribute>
```

Could you add `t-translation="off"` to the attributes, to prevent odoo from trying to translate this string/value?
Otherwise.. bugs can arise from translations, which isn't good.

(please correct the other attributes as well.. 🙏🏻 )

ivantodorovich (changes requested):

pos_access_right/static/src/xml/NumpadWidget.xml
hitrosol: sorry for delay response. Could you please review again. Thanks.

pos_product_template_configurator/i18n/it.po
hitrosol: I didn't touch this file. What should I do ? Thanks

PR #867, opened 2022-10-26
[WIP][15.0][MIG]pos_ticket_without_price

manuelregidor:
Standard migration.


manuelregidor:
@HaraldPanten @ValentinVinagre

manuelregidor:
@ivantodorovich  I have migrated the module from v14 to v15 and there is an incompatibiilty with the module pos_event_sale. The thing is that not only the selected receipt type (with or without price) is printed along with the event ticket, but both recepit types are included in the printed document. How could we sort it out? Thank you.

ivantodorovich:
I see. I can't really tell why that's happening. I see nothing weird on `pos_event_sale`.

But on this side, I wonder.. why rendering 2 receipts at all?
Why not simply hide the prices on the original receipt component, instead of rendering both ?

manuelregidor:
@ivantodorovich When the module pos_event_sale is not installed, only the selected receipt is rendered, not both. It is line 9 (display: block !important;) in file event_registration_receipt.scss that makes both receipts to be rendered.

HaraldPanten:
@ivantodorovich Did you have avilability to have a look?

THX!

ivantodorovich:
IMO the implementation on this module is wrong. It shouldn't rely on css to hide/show one receipt or the other.
It should use `props` and `t-if` directives instead.

But if you wanna stick to this, a quick fix without changing `pos_event_sale` would be to set the style to `display: none !important` instead. That would give it priority over the `scss` stylesheet.

If I recall correctly, `display: block !important` on the `pos_event_sale` module is to be able to properly display more than 1 receipt at the same time (the order receipt + the registration vouchers). Maybe it could be removed but it has to be analized

Issue #869, opened 2022-10-26
pos_empty_home bottom text collapses


xabispacebiker:
## Module
pos_empty_home

## Describe the bug
The home text collapses and the scroll bars are displayed in the bottom space when the categories ocuppy upper space
<img width="1280" alt="Captura de Pantalla 2022-10-26 a las 14 54 11" src="https://user-images.githubusercontent.com/1733884/198032690-3765ce9c-83b4-4e42-8383-9f5036e3bc6a.png">

## To Reproduce
Just install the module and open a POS session

1. Install the module pos_empty_home
2. Add as many categories so that the top space occupies half of the screen
3. Open the POS screen

**Expected behavior**
The home text should be centered vertically and horizontally
<img width="1280" alt="Captura de Pantalla 2022-10-26 a las 14 55 42" src="https://user-images.githubusercontent.com/1733884/198032728-2d9f7562-5bc8-4fd9-b6de-3da4e0eb05c0.png">

**Additional context**
I have the solution, I will send a PR. 

The solution is to add the text within a span and add the folowing css:

.```
product-list-empty-home {
    position: absolute;
    height: 100%;
    width: 100%;
    display: table;
}
.product-list-empty-home span {
    display: table-cell;
    vertical-align: middle;
    text-align: center;
    font-size: x-large;
}
```

legalsylvain:
thanks for reporting! 
please ping me when the PR is available.

regards.

xabispacebiker:
@legalsylvain , pr is ready https://github.com/OCA/pos/pull/870

legalsylvain:
Closed by #870

PR #870, opened 2022-10-26
Fixes https://github.com/OCA/pos/issues/869

xabispacebiker:
Fixes:
- https://github.com/OCA/pos/issues/869


ivantodorovich (approved):
Seems legit

PR #875, opened 2022-10-31
[14.0][IMP] pos_cash_move_reason: Restrict move reasons by pos.config. Sele…

ilyasProgrammer:
…ct first journal by default when move reason is selected in the wizard.

### DIFF
--- a/pos_cash_move_reason/models/pos_move_reason.py
+++ b/pos_cash_move_reason/models/pos_move_reason.py
@@ -48,6 +48,8 @@ class PosMoveReason(models.Model):
         required=True,
     )
 
+    allowed_pos_ids = fields.Many2many("pos.config", string="Allowed POS")
+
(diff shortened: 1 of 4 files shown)


legalsylvain (commented):
Hi. thanks for your improvment.

I understand your need, but the current implementation requires extra configuration. I mean, user has to configure all the ``allowed_config_ids`` for all the pos.move.reason ? Otherwise, nothing willl works.

what about to change the domain with something like : 

```
'|', 
                        ("allowed_pos_ids", "in", self.session_id.config_id.ids),
                        ("allowed_pos_ids", "=", False),
```

francesco-ooops (approved):
functional review ok!

francesco-ooops:
@legalsylvain looks good to you? :)

legalsylvain (approved):
LGTM, code review / no test.

/ocabot merge minor

PR #878, opened 2022-11-08
[14.0][IMP] `pos_product_template`: Sort product attribute

ygcarvalh:
This improvement allows the product attributes - in the PopUp that is opened in the POS - to be organized by sequence and not by the registration ID in the system.


ivantodorovich (approved):
Seems legit, but please check pre-commit as it's red

ivantodorovich:
@ygcarvalh rebasing should solve the issue

GabbasovDinar (approved):
Code LGTM!

PR #879, opened 2022-11-09
[ADD] pos_partner_no_create

victor-champonnois:
Forbid customer creation from the POS;

PR #880, opened 2022-11-09
[ADD] pos_hide_partner_info

victor-champonnois:
Hides phone and address field in pos customer list.


pos_hide_partner_info/static/src/xml/templates.xml
```diff
@@ -0,0 +1,38 @@
...
+                <td>
+                    <t t-esc='partner.name' />
+                </td>
+            </tr>
+        </t>
+    </t>
+
+    <t t-extend="ClientDetails">
+        <t t-jquery=".client-details-left" t-operation="replace">
+            <div class='client-details-left' />
```
legalsylvain: sale remark as https://github.com/OCA/pos/pull/879

For exemple, this module can conflict with
-  ``pos_partner_firstname`` (``<t t-jquery=".client-details-left" t-operation="prepend">``)
- ``pos_customer_required_fields`` (``<t t-jquery="div.client-details-left" t-operation="append">``)

legalsylvain (changes requested):
thanks for your contribution. one remark inline.
Do you think, it could be great to have the field list configurable ? (not the topic of the current PR, but can be added in the roadmap).

pos_hide_partner_info/static/src/xml/templates.xml
victor-champonnois: I don't see a way to select [these `client-detail` elements](https://github.com/odoo/odoo/blob/c53081f10befd4f1c98e46a450ed3bc71a6246ed/addons/point_of_sale/static/src/xml/pos.xml#L346)  that have a child with class `client-address-street`. Apparently it's not possible to select the parent of an element with CSS selectors. 
Is it OK if I make the whole [client-details-left](https://github.com/odoo/odoo/blob/c53081f10befd4f1c98e46a450ed3bc71a6246ed/addons/point_of_sale/static/src/xml/pos.xml#L345) invisible, with a warning in Usage.rst ?

robinkeunen: @victor-champonnois I don't think that would work. We will actually use both `pos_partner_firstname` and `pos_hide_partner_info` on our production server :/

victor-champonnois: OK I found ways to do it without brute force replacing.

pos_hide_partner_info/static/src/xml/templates.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template">
+    <t t-extend="ClientListScreenWidget">
+
+        <t t-jquery="th:contains('Address')" t-operation="attributes">
```
robinkeunen, line 5: I don't think it will work with translations unfortunatly :/

robinkeunen (changes requested):

pos_hide_partner_info/static/src/xml/templates.xml
victor-champonnois: It does I think, I tested it on a DB with french translation and it works)

victor-champonnois: I ask for it to be tested in the functional tests

robinkeunen:
@legalsylvain any time for another review ?

pos_hide_partner_info/static/src/xml/templates.xml
```diff
@@ -0,0 +1,53 @@
...
+        <t t-jquery="th:contains('Phone')" t-operation="attributes">
+            <attribute name="class">oe_hidden</attribute>
+        </t>
+    </t>
+
+    <t t-extend="ClientLine">
+        <t t-jquery="td:has([t-esc='partner.address'])" t-operation="attributes">
+            <attribute name="class">oe_hidden</attribute>
+        </t>
+        <t t-jquery="td:has([t-esc*='phone'])" t-operation="attributes">
```
legalsylvain, line 17: ``*`` typo ?

victor-champonnois: No, it means 'contains "phone"'

victor-champonnois: https://api.jquery.com/category/selectors/

legalsylvain: oh ! thanks !

legalsylvain (approved):

robinkeunen (approved):

PR #881, opened 2022-11-09
[IMP] pos_invoicing : confirm invoices that are created in the back-office by the UI, to avoid user to change / delete invoices created and paid from PoS

legalsylvain:
Rational : 

In point of sale there are two way to create invoices. 
1. by enabling invoicing in the UI (front office)
2. by clicking on "invoice" on a pos.order if the session is not closed. (back office)

In the first case, the invoice is open, so it is not possible to change it. (as the module disables the cancel option).
But in the second case, the invoice is in a draft state, so it is possible to change / update / delete the invoice that should not be possible.

### DIFF
--- a/pos_invoicing/models/pos_order.py
+++ b/pos_invoicing/models/pos_order.py
@@ -3,7 +3,8 @@
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import _, api, models
+from odoo.exceptions import UserError
 
 
 class PosOrder(models.Model):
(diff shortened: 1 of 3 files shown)


pos_invoicing/models/pos_order.py
```diff
@@ -15,3 +16,52 @@ def _prepare_invoice(self):
...
+        res = super().action_pos_order_invoice()
+        if not self.env.context.get("pos_order_invoiced", False):
+            for order in self:
+                order.invoice_id.sudo().with_context(
+                    force_company=self.env.user.company_id.id,
+                    pos_picking_id=order.picking_id,
+                ).action_invoice_open()
+                # Check if total amount are the same
+                if (
+                    order.invoice_id.amount_total_signed != order.amount_total
```
ivantodorovich: Use float_compare with a the currency digits/precision.
Otherwise you may get some unexpected results due to `2.0000001 != 2.00`

legalsylvain: totally right ! thanks for your comment.

legalsylvain: done. Could you take a look ?

pos_invoicing/models/pos_order.py
```diff
@@ -15,3 +17,57 @@ def _prepare_invoice(self):
...
+                    force_company=self.env.user.company_id.id,
+                    pos_picking_id=order.picking_id,
+                ).action_invoice_open()
+                # Check if total amount are the same
+                prec = order.invoice_id.currency_id.rounding
+                if float_compare(
+                    order.invoice_id.amount_total_signed - order.amount_total,
+                    0.0,
+                    precision_digits=prec
+                ):
```
ivantodorovich: That works, but this is how float_compare is supposed to be used:

```suggestion
                if float_compare(
                    order.invoice_id.amount_total_signed,
                    order.amount_total,
                    precision_digits=prec
                ):
```

It compares arg1 vs arg2 with the provided precision
possible return values are: `-1` (it's lower), `0` (equal) or `1` (it's bigger)

ivantodorovich (approved):

pos_invoicing/models/pos_order.py
legalsylvain: you're right again ! 
sorry for the noice. Fixed.

ivs-cetmix (approved):
Looks good

PR #886, opened 2022-11-18
[ADD] 14.0 pos_pricelist_show_discount

PierrickBrun:
Maybe I should propose a patch to odoo instead, but I'm not sure if this is the best way to fix this.

### DIFF
--- a/pos_pricelist_show_discount/__manifest__.py
+++ b/pos_pricelist_show_discount/__manifest__.py
@@ -1,0 +1,15 @@
+{
+    "name": "PoS show discount from pricelist",
+    "summary": "PoS: show appropriate discount when discounting from pricelist",
+    "version": "14.0.1.0.0",
+    "category": "Point of Sale",
+    "website": "https://github.com/OCA/pos",
+    "author": "Odoo Community Association (OCA), Akretion",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "point_of_sale",
+    ],
+    "data": ["templates/assets.xml"],
+}
--- a/pos_pricelist_show_discount/readme/DESCRIPTION.rst
+++ b/pos_pricelist_show_discount/readme/DESCRIPTION.rst
@@ -1,0 +1,2 @@
+When using pricelist based on another pricelist the discount is not correctly showed by odoo.
+This module fixes it.
--- a/pos_pricelist_show_discount/static/src/js/models.js
+++ b/pos_pricelist_show_discount/static/src/js/models.js
@@ -1,0 +1,46 @@
+odoo.define("pos_pricelist_show_discount.models", function (require) {
+    "use strict";
+    var models = require("point_of_sale.models");
+    var orderline_super = models.Orderline.prototype;
+
+    models.Orderline = models.Orderline.extend({
+        get_lst_price: function () {
+            var lst_price = orderline_super.get_lst_price.apply(this, arguments);
+            if (this.display_discount_policy() === "without_discount") {
+                var self = this;
+                var category_ids = [];
+                var category = self.product.categ;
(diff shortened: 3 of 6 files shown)


legalsylvain:
Is not the same as https://github.com/OCA/pos/pull/859 ?

PierrickBrun:
> Is not the same as #859 ?

Not sure, I need to check

francesco-ooops:
@PierrickBrun can you please rebase for testing?

elvise:
> Is not the same as https://github.com/OCA/pos/pull/859 ?

@legalsylvain no this is another story.

PierrickBrun:
> @PierrickBrun can you please rebase for testing?

done

elvise:
> > @PierrickBrun can you please rebase for testing?
> 
> done

Thanks! Let me test it

houssine78 (approved):
LGTM

houssine78:
@PierrickBrun @legalsylvain does there is a way to test it on runboat without merging beforehand or it needs to be tested locally ?

francesco-ooops:
@houssine78 yes you need to click here: 

image

houssine78:
> @houssine78 yes you need to click here:
> 
> image

Ok I didn't saw it ! It needed to scroll down to have it shown. thanks a lot !

houssine78:
@PierrickBrun I tested it and it works well. The only thing that I saw is if there is a chain of dependency on pricelist. it doesn't handle the discount until the top but just with the mother pricelist. It's not blocking for my use case anyway. Up to you to handle it or not but could you just add it as a limitation in the readme.

francesco-ooops:
@PierrickBrun I second @houssine78 request, I think the best possible implementation would be allowing user to define in POS config the pricelist to use as "base" for the discount, so basically discount = ratio between "base" pricelist and the one used in POS

PierrickBrun:
> @PierrickBrun I tested it and it works well. The only thing that I saw is if there is a chain of dependency on pricelist. it doesn't handle the discount until the top but just with the mother pricelist. It's not blocking for my use case anyway. Up to you to handle it or not but could you just add it as a limitation in the readme.

I will probably need to handle this in the future so it should be taken into account.

Thank you for thinking about it.

I think the easiest way is just to find the first pricelist which does not depend on another one

francesco-ooops:
> I think the easiest way is just to find the first pricelist which does not depend on another one

TBH this would not solve our customer's scenario, where 4 different pricelists are used to calculate final price, but the one to keep as reference for discount is the 2nd, and the one used in POS is the 4th

I think there is much more flexibility in just comparing price between two pricelists, regardless of how they're related. 

On a side note, this makes me think there should be an option to only display negative discount, as it would not be very convenient to inform customer they've paid 20% more on a product lol

PierrickBrun:
> @PierrickBrun I second @houssine78 request, I think the best possible implementation would be allowing user to define in POS config the pricelist to use as "base" for the discount, so basically discount = ratio between "base" pricelist and the one used in POS

This is a good idea I think, this would probably be the pos.config's default pricelist then

francesco-ooops:
@PierrickBrun why force the default pricelist and not allow to choose the one to use as reference? The domain for reference pricelist selection could be the allowed pricelists on that POS.

francesco-ooops:
@PierrickBrun I tested standard flow on odoo runbot with 

- pricelist 1 setting a fixed price 100 on all products
- pricelist 2 setting a 50% discount on pricelist 1, with discount displayed to the customer

here is the result with pricelist 2 in POS:

image

Can you help me understand what this PR does differently?

francesco-ooops:
I think an odoo PR to rethink the logic would be the best way to go, as "Show public price & discount to the customer" pricelist policy on odoo runbot provides the most random results:

image

PR #888, opened 2022-11-23
[FIX][14.0] parsing fixed amount for discount

Rad0van:
The original line is obviously 1:1 copy from percentual discount which aimed to get value between 0 and 100 (percent) but in case of fixed discount doesn't make any sense and actually caps any input at 100.

### DIFF
--- a/pos_fixed_discount/static/src/js/FixedDiscountButton.js
+++ b/pos_fixed_discount/static/src/js/FixedDiscountButton.js
@@ -17,7 +17,7 @@ odoo.define("pos_fixed_discount.FixedDiscountButton", function (require) {
                 startingValue: 0,
             });
             if (confirmed) {
-                var val = Math.round(Math.max(0, Math.min(100, parseFloat(payload))));
+                var val = Math.round(parseFloat(payload));
                 this.apply_discount(val);
             }
         }


Rad0van:
For some reason pre-commit fails across this repo (not only this branch) with something like this:
```
[INFO] Initializing environment for https://gitlab.com/PyCQA/flake8.
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: could not read Username for 'https://gitlab.com/': No such device or address
```

Anyone has any idea? @legalsylvain maybe?

ivantodorovich (approved):
Try rebasing, I think it's already fixed

ivs-cetmix:
Module is fine. However would be great to add a configuration part.
Eg "Enable global discounts in POS settings"
image

Rad0van:
> Module is fine. However would be great to add a configuration part. Eg "Enable global discounts in POS settings" image

Well I was only trying to fix it really quickly. I guess such change would call for separate PR.

ivs-cetmix:
> > Module is fine. However would be great to add a configuration part. Eg "Enable global discounts in POS settings" image
> 
> Well I was only trying to fix it really quickly. I guess such change would call for separate PR.

Yup I agree with you. Just not to forget about this. Because It took me some extra time to figure this out during testing.

ivs-cetmix (approved):
LGTM!

PR #891, opened 2022-11-25
[15.0][MIG] pos_fixed_discount

Rad0van:
Standard migration.


ivantodorovich (approved):
Could you check pre-commit please, seems red?

GabbasovDinar (approved):
Code LGTM!

pos_fixed_discount/static/src/js/FixedDiscountButton.js
```diff
@@ -17,7 +17,7 @@ odoo.define("pos_fixed_discount.FixedDiscountButton", function (require) {
                 startingValue: 0,
             });
             if (confirmed) {
-                var val = Math.round(Math.max(0, Math.min(100, parseFloat(payload))));
+                var val = parseFloat(payload);
```
chienandalu, line 20: What's the rationale for this change?

chienandalu (changes requested):

pos_fixed_discount/static/src/js/FixedDiscountButton.js
ivs-cetmix: Looks like it comes from https://github.com/OCA/pos/pull/888
Either it is needed to be merged first or we need to migrate "as is" and than apply this patch with a new PR

Rad0van: Yeah - it is from #888 I had it applied in local repo and forgot about it when doing the migration so the proper way should be as @isserver1 suggests...

PR #894, opened 2022-11-26
[16.0][NEW] pos_order_reorder: new module to create new order from order history

GabbasovDinar:


pos_order_reorder/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "https://github.com/OCA/pos",
+    "author": "Cetmix,Odoo Community Association (OCA)",
+    "data": [
+        "views/res_config_settings_view.xml",
+    ],
+    "installable": True,
+    "assets": {
+        "point_of_sale.assets": [
+            "pos_discount/static/src/js/**/*",
+            "pos_discount/static/src/xml/**/*",
```
sbidoul: If I read correctly, the test error come from here. Not sure why we reference pos_discount here?

pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const {useListener} = require("@web/core/utils/hooks");
+    const PosComponent = require("point_of_sale.PosComponent");
+    const Registries = require("point_of_sale.Registries");
+    const {Orderline} = require("point_of_sale.models");
+
+    class ReorderButton extends PosComponent {
+        setup() {
+            super.setup();
+            useListener("click", this._onClick);
```
PhilDL: Is the `useListener` intentional here? to be able to have contextual / overriden button templates that don't define the `t-on-click` directive (abstract behavior, without template)?

If that's not the case, since `useListener` is a **deprecated** hook, please prefer to bind the click event on the template directly:

```xml
<t t-name="ReorderButton" owl="1">
        <span class="control-button" t-on-click="_onClick">
            <i class="fa fa-refresh" />
            <span></span>
            <span>Re-order</span>
        </span>
    </t>
```
With that you can completely remove that setup() method.

pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.js
```diff
@@ -0,0 +1,74 @@
+odoo.define("pos_order_reorder.ReorderButton", function (require) {
```
PhilDL: A lot of POS code is written in the old `odoo.define` way but since this is a pure OWL Component I would suggest using the new Odoo module syntax:

```js
/** @odoo-module **/
import {useListener} from "@web/core/utils/hooks"; // probably remove, see other review
import PosComponent from "point_of_sale.PosComponent";
import Registries from "point_of_sale.Registries";
import {Orderline} from "point_of_sale.models";

class ReorderButton extends PosComponent {
    // ....
}

ReorderButton.template = "ReorderButton";
Registries.Component.add(ReorderButton);

export default ReorderButton;
```
If you go with that rewriting, you will also have to probably rename the file "ReorderButton.esm.js" so the file will be picked up by the OCA `.eslint` rule for ESM modules.

pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.js
```diff
@@ -0,0 +1,74 @@
...
+                    });
+                }
+                new_line.set_unit_price(line.get_unit_price());
+                new_line.set_quantity(line.get_quantity());
+                new_line.set_discount(line.get_discount());
+                newOrder.add_orderline(new_line);
+            }
+            this.trigger("click-order", newOrder);
+        }
+        async _onClick() {
```
PhilDL: The pile of execution starting from here is synchronous, you can remove all the `async` and `await` here, `_reOrder` should also get its `async` keyword removed.

pos_order_reorder/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "website": "https://github.com/OCA/pos",
+    "author": "Cetmix,Odoo Community Association (OCA)",
+    "maintainers": ["GabbasovDinar", "CetmixGitDrone"],
+    "data": [
+        "views/res_config_settings_view.xml",
+    ],
+    "installable": True,
+    "assets": {
+        "point_of_sale.assets": [
+            "pos_order_reorder/static/src/js/**/*",
```
PhilDL: I suggest here to be explicit about the extensions `*.js` and `*.xml`.

PhilDL (changes requested):

pos_order_reorder/__manifest__.py
GabbasovDinar: @PhilDL Done

PhilDL: LGTM

PhilDL (approved):

ivs-cetmix:
@sbidoul would appreciate your review

sbidoul:
Sorry, besides maintaining pos_payment_terminal I'm not really involved in this repo.

ivs-cetmix:
@OCA/pos-maintainers would appreciate your review

legalsylvain (commented):
Thanks for sharing your work.

Partial review. don't you think this icon could be more accurate ? https://fontawesome.com/icons/copy?s=regular&f=classic
The refresh icon means different things to me.
(not a blocking point)

regards.

ivs-cetmix:
> Thanks for sharing your work.
> 
> Partial review. don't you think this icon could be more accurate ? https://fontawesome.com/icons/copy?s=regular&f=classic The refresh icon means different things to me. (not a blocking point)
> 
> regards.

Thank you @legalsylvain looks reasonable. Icon is changed. Would appreciate your review

ivs-cetmix (approved):

pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.esm.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+        if (order.fiscal_position) {
+            newOrder.fiscal_position = order.fiscal_position;
+        }
+        if (order.pricelist) {
+            newOrder.set_pricelist(order.pricelist);
+        }
+        const lines = order.get_orderlines();
+        for (var i = 0; i < lines.length; i++) {
+            const line = lines[i];
```
ivantodorovich, line 31:
```suggestion
        for (const line of order.get_orderlines()) {
```

pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.esm.js
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    pos: pos,
+                    order: newOrder,
+                    product: pos.db.get_product_by_id(line.get_product().id),
+                    description: line.name,
+                    price: line.price_unit,
+                    tax_ids: newOrder.fiscal_position ? undefined : line.tax_id,
+                    price_manually_set: true,
+                    customer_note: line.customer_note,
+                }
```
ivantodorovich: Move this vals preparation to another method, for example: `_prepareReorderLineVals` or similar
This way we have a hook point for other modules to add their own fields/properties

pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.esm.js
GabbasovDinar: Could you explain in more detail why this change is necessary?

ivantodorovich: It's not necessary, just a suggestion to write modern js ;)

ivantodorovich (approved):

PR #899, opened 2022-12-05
[15.0][MIG]pos_payment_change

manuelregidor:
Standard migration.


manuelregidor:
@HaraldPanten @ValentinVinagre

legalsylvain:
please remove 9d8a4808745660e6c6a652451a7344956aee6785
I'll do correct specific PR.

legalsylvain:
Once https://github.com/OCA/pos/pull/903 is done, you could rebase your PR.

regards.

manuelregidor:
@legalsylvain Could you check if the commits history is correct, please? Thank you.

pos_payment_change/models/pos_order.py
```diff
@@ -35,14 +35,12 @@ def change_payment(self, payment_lines):
...
-                self.env.user.name,
-                datetime.today(),
-            )
+            "The payments of the Order {} (Ref: {}) has been changed" " by {} at {}."
+        ).format(
+            self.name,
+            self.pos_reference,
+            self.env.user.name,
+            datetime.today(),
         )
```
ivantodorovich: This change is unnecessary and it will deem translations useless.
But if you really want to change it then I recommend doing it like this to use named parameters:

```py
comment = _(
    "The payments of the Order %(order)s (Ref: %(ref)s have been changed by %(user_name)s on %(today)s",
    order=self.name,
    ref=self.pos_reference,
    user_name=self.env.user.name,
    today=datetime.today(),
)
```

(same for other cases btw)

And please make sure to update any .pot or .po files accordingly (if there are any)

ivantodorovich (changes requested):

ivantodorovich (approved):
Thanks! Looks good!

GabbasovDinar (approved):
Code LGTM!

PR #907, opened 2022-12-07
[FIX][12.0] pos_order_to_sale_order : allow to create sale order from PoS when the products are only services

legalsylvain:
**Rational :** 
- Open your PoS
- begin a pos order, selecting only a service product
- set a customer
- try to create a delivered order
- -> there is an blocking error

**Analysis**
- confirm picking should not be called, if there are not picking associated to the created order

Trivial patch, thanks for your review.

GRAP Ticket 747 CC @quentinDupont


ivantodorovich (approved):
Code LG

GabbasovDinar:
Code LG

GabbasovDinar (approved):

PR #911, opened 2022-12-20
[FIX] pos_order_to_sale_order

legalsylvain:
set correct image path in the description of the module.


legalsylvain:
Trivial.
Fast tracking.
/ocabot merge patch

PR #915, opened 2023-01-12
[ADD] pos_product_quick_info: Add new module to display product without add in to card

GabbasovDinar:


legalsylvain:
Hi. tested on runboat. Looks ok. 
Functional question : how do you use this popup ? I see the interest, but It looks hugly and not responsive at all. (close button is hidden if the screen is not big).

ivs-cetmix:
> Hi. tested on runboat. Looks ok.
> Functional question : how do you use this popup ? I see the interest, but It looks hugly and not responsive at all. (close button is hidden if the screen is not big).

Thank you for the feedback! The idea behind this module is to be able to check product info without adding it to cart.
Regarding "Close" button what do you think if popup will be closed when clicked on popup itself?

legalsylvain (approved):
Code review and test. Lgtm. Thanks !

legalsylvain:
> Thank you for the feedback! The idea behind this module is to be able to check product info without adding it to cart.
Regarding "Close" button what do you think if popup will be closed when clicked on popup itself?

Well. Just tested on smartphone and the display is correct. In fact, it seems bad displayed only on médium screen. Should be fixed in odoo or in another dedicated module.

Thanks for your contribution.

ivs-cetmix:
@OCA/pos-maintainers would appreciate someones review

pos_product_quick_info/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "PoS Product Quick Info",
+    "version": "16.0.1.0.0",
+    "summary": "Display product info by one click in Point of Sale",
+    "author": "Cetmix,Odoo Community Association (OCA)",
```
ivantodorovich:
```suggestion
    "author": "Cetmix, Odoo Community Association (OCA)",
```

ivantodorovich (approved):

pos_product_quick_info/__manifest__.py
ivs-cetmix: Thank you very much!

ivs-cetmix (approved):
Functional ok!

ivs-cetmix (approved):

PhilDL (approved):

PR #919, opened 2023-01-20
[14.0][FIX]add filter on pos_reference field when get pos records orders

DorianMAG:
Error get pos orders records.
When pos_reference field is empty regex iteration breaks.
To fix, add filter to exclude empty pos_reference.


flotho:
Hi @alexis-via we found a strange issue here.
When using your awesome module pos_session_pay_invoice pos order are created with no pos_reference 
then ins the POS when you try to get all the orders with the right button on the right, the method `export_for_ui` crash because it doesn't succeed to apply the regexp here https://github.com/OCA/OCB/blob/14.0/addons/point_of_sale/models/pos_order.py#L710
Totally normal because the pos order you create are not originated from the POS.
So we proposed to not load the pos order coming from your module when you try to get the order history.

flotho (changes requested):
here the pylint will fail
you need a space

pos_session_pay_invoice/models/pos_order.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class PosOrder(models.Model):
+    _inherit = "pos.order"
+    
+    @api.model
+    def search_paid_order_ids(self, config_id, domain, limit, offset):
+        domain = [("pos_reference", "!=", False), ("lines", "!=", False)]
+        res = super(PosOrder, self).search_paid_order_ids(config_id, domain, limit, offset)
+        return res
```
flotho: same thing here

flotho (changes requested):

flotho:
hi @legalsylvain already faced the same issue ?

pos_session_pay_invoice/models/pos_order.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import api, models
+
+
+class PosOrder(models.Model):
+    _inherit = "pos.order"
+    
+    @api.model
+    def search_paid_order_ids(self, config_id, domain, limit, offset):
+        domain = [("pos_reference", "!=", False), ("lines", "!=", False)]
```
legalsylvain: You should add domain to the current domain. Or did I missed something ?

legalsylvain (commented):
I dont use this module for the time being.

pos_session_pay_invoice/models/pos_order.py
flotho: hum right, we forgot to add instaed of replacing.

flotho:
@DorianMAG thanks for this first step,
Now you should use the AND() operator to combine the orginal domain with the one you want to apply otherwise your cod is breaking the flow of inheritance

pos_session_pay_invoice/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import pos_order
```
flotho: missing return here

pos_session_pay_invoice/__init__.py
```diff
@@ -1,3 +1,4 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 from . import wizard
+from . import models
```
flotho, line 4: missing return here

flotho (changes requested):
@DorianMAG thanks for this first step,
Now you should use the AND() operator to combine the orginal domain with the one you want to apply otherwise your cod is breaking the flow of inheritance

pos_session_pay_invoice/models/__init__.py
DorianMAG: thanks for your comments.
It s done now.
Regards

pos_session_pay_invoice/models/pos_order.py
DorianMAG: Hi @legalsylvain .
Thank you for your feedback.
It's kinda new to me

flotho:
thanks @DorianMAG

flotho (approved):

DorianMAG:
FIX already done and merged
https://github.com/OCA/pos/pull/1006

DorianMAG:
FIX already done and merged
https://github.com/OCA/pos/pull/1006

PR #922, opened 2023-01-25
[ADD] pos_stock_available_online: Add new module

GabbasovDinar:
Show the available quantity of products in the Point of Sale


ivs-cetmix (approved):
Functional is ok. Need to add a note regarding same company warehouses though.

pos_stock_available_online/models/pos_session.py
```diff
@@ -0,0 +1,22 @@
+from odoo import models
+
+
+class PosSession(models.Model):
+    _inherit = "pos.session"
+
+    def _process_pos_ui_product_product(self, products):
+        config = self.config_id
+        if config.display_product_quantity:
```
hugho-ad, line 9: return early and save a identation level
```suggestion
        if not config.display_product_quantity:
            return super()._process_pos_ui_product_product(products)
```

pos_stock_available_online/models/res_config_settings.py
```diff
@@ -0,0 +1,23 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    pos_display_product_quantity = fields.Boolean(
```
hugho-ad, line 7: res.configs.settings are per company

this fields should be on pos.config form view instead

pos_stock_available_online/models/stock_quant.py
```diff
@@ -0,0 +1,40 @@
...
+        pos_session_obj = self.env["pos.session"]
+        for quant in self:
+            warehouse_id = quant.warehouse_id.id
+            configs = pos_session_obj.search(
+                [
+                    ("state", "=", "opened"),
+                    ("config_id.display_product_quantity", "=", True),
+                    "|",
+                    ("config_id.additional_warehouse_ids", "in", [warehouse_id]),
+                    ("config_id.main_warehouse_id", "=", warehouse_id),
```
hugho-ad, line 31: also quant.product in producs inside categories setup on the pos.config?

pos_stock_available_online/models/stock_quant.py
```diff
@@ -0,0 +1,40 @@
...
+                ],
+            ).mapped("config_id")
+            if configs:
```
hugho-ad, line 41: you got a for loop inside the _notify_available_quantity, if not record on the set, then will not perform any 
```suggestion
            configs._notify_available_quantity(quant._prepare_pos_message())
```

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
```diff
@@ -0,0 +1,43 @@
...
+import Registries from "point_of_sale.Registries";
+import {format} from "web.field_utils";
+import utils from "web.utils";
+
+const StockProductItem = (ProductItem) =>
+    class StockProductItem extends ProductItem {
+        format_quantity(quantity) {
+            const unit = this.env.pos.units_by_id[this.props.product.uom_id[0]];
+            var formattedQuantity = `${quantity}`;
+            if (unit) {
```
hugho-ad, line 13:
```suggestion
            if (!unit) return `${formattedQuantity}`;
```

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductsWidget.esm.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+        async _handleNotification(payloads) {
+            if (this.env.isDebug()) {
+                console.log("Payloads:", payloads);
+            }
+            const db = this.env.pos.db;
+            const ProductIds = [];
+            for (const payload of payloads) {
+                for (const message of payload) {
+                    var product = db.get_product_by_id(message.product_id);
```
hugho-ad, line 39: you could avoid this if you sent from  backend only notifications used on the pos.config

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
```diff
@@ -0,0 +1,43 @@
+/** @odoo-module **/
```
hugho-ad, line 1: use 
odoo.define'module.class', function (require) {

hugho-ad: also rename the fiele
Productitem.esm.js > Productitem_esm.js

hugho-ad (changes requested):

hugho-ad:
Hello @GabbasovDinar 

thanks for the contribution! awsome!

how can we help you with the latest changes?

PhilDL:
Thanks for the contribution, the code looks okay to me, but I have some doubts about the UX, it looks very cluttered 
with 2 warehouses:

## Actual
CleanShot 2023-02-16 at 15 43 34

## Proposal
This is a suggestion of UX changes I tried, changing a few things. Feel free to follow it or not, depending on general feedback. It would look like that:

CleanShot 2023-02-16 at 16 51 02

### Changes made:
Use the Warehouse `code` instead of the name to gain some space:
```python
class StockWarehouse(models.Model):
    _inherit = "stock.warehouse"

    def _prepare_vals_for_pos(self, product):
        """
        Prepare warehouse info data to send a POS
        """
        self.ensure_one()
        return {
            "id": self.id,
            "name": self.name,
            "code": self.code,
            "quantity": product.with_context(warehouse=self.id).immediately_usable_qty,
            "product_id": product.id,
        }
````
**I changed the view with:**
- XPath inside "product-content" 
- Changed the order of the elements
- added font-awesome icon, also used the code instead of the warehouse name:


```xml
<t
        t-name="ProductItem"
        t-inherit="point_of_sale.ProductItem"
        t-inherit-mode="extension"
        owl="1"
    >
        <xpath expr="//div[hasclass('product-content')]" position="inside">
            <div t-if="env.pos.config.display_product_quantity" class="warehouse-info">
                <div class="flex">
                    <div t-if="warehouses.length > 1" class="warehouse total" title="Total">
                        <i class="fa fa-dropbox"/>
                        <span
                            t-attf-class="quantity {{total_quantity > env.pos.config.minimum_product_quantity_alert ? 'available' : 'not-available'}}"
                        >
                            <t t-esc="display_total_quantity" />
                        </span>
                    </div>
                </div>
                <div class="flex">
                    <t t-foreach="warehouses" t-as="warehouse" t-key="warehouse.id">
                        <div class="warehouse" t-att-title="warehouse.name">
                            <span class="warehouse-name">
                                <t t-esc="warehouse.code" />
                            </span>
                            <span
                                t-attf-class="quantity {{warehouse.quantity > env.pos.config.minimum_product_quantity_alert ? 'available' : 'not-available'}}"
                            >
                                <t t-esc="format_quantity(warehouse.quantity)" />
                            </span>
                        </div>
                    </t>
                </div>
            </div>
        </xpath>
    </t>
```

Removed the absolute positioning and made use of flex to display elements in a more compact way:

```css
.pos .product-list .warehouse-info {
    padding: 0 6px;
    font-weight: bold;
    display: flex;
    border-top: 1px solid #efefef;
    justify-content: space-between;
}
.pos .product-list .warehouse-info .warehouse {
    display: flex;
    flex-direction: row;
    justify-content: space-between;
    align-items: center;
}
.pos .product-list .warehouse-info .warehouse .quantity {
    color: black;
    padding: 1px 2px;
    font-size: 11px;
}
.pos .product-list .warehouse-info .warehouse .warehouse-name {
    display: block;
    color: #696969;
    font-size: 10px;
}
.pos .product-list .warehouse-info .warehouse.total .warehouse-name {
    font-weight: bold;
}
.pos .product-list .warehouse-info .warehouse .quantity.available {
    color: #32a868;
}
.pos .product-list .warehouse-info .warehouse .quantity.not-available {
    color: #ef5350;
}
```

pos_stock_available_online/models/pos_session.py
GabbasovDinar: Hello, @hugho-ad the negative condition is more difficult to read code I see no reason to change in this way

pos_stock_available_online/models/res_config_settings.py
GabbasovDinar: starting with Odoo 16 all settings specific to POS are defined in this model, you can check the original Odoo code. https://github.com/odoo/odoo/blob/b0df10c1e497f65b891f7141055e7f339d11b4b1/addons/point_of_sale/models/res_config_settings.py#L48-L105

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
GabbasovDinar: We should use no-negative condition: https://eslint.org/docs/latest/rules/no-negated-condition

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductsWidget.esm.js
GabbasovDinar: If a new product is added to the POS, this product will not be available in the POS until the POS product is read again (for example, until the page is updated), this check is necessary so that the POS does not process information about the product that is available in the POS but not actually displayed (because the pos db is not yet updated)

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
GabbasovDinar: 1. I see no reason to use the old approach of defining modules when there is OWL and the new approach, which also will not create problems in porting
2. Why? the file with this extension will be picked up by the OCA .eslint rule for ESM modules.:
see https://github.com/OCA/pos/blob/16.0/.eslintrc.yml#L11
and https://github.com/OCA/pos/pull/925#issuecomment-1429543607

pos_stock_available_online/models/pos_session.py
hugho-ad: not it's not, is called early return pattern
https://medium.com/swlh/return-early-pattern-3d18a41bba8

pos_stock_available_online/models/res_config_settings.py
hugho-ad: awesome, I didn't know that 

thanks!

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
hugho-ad: nice! there is a eslint about that!

pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
hugho-ad: @GabbasovDinar  thanks man, I need and update with v16.0

pos_stock_available_online/models/pos_session.py
GabbasovDinar: @hugho-ad we have to return "super" in any case. In your case, you offer to return only if this setting is not specified, which is not true.

pos_stock_available_online/models/stock_quant.py
GabbasovDinar: this check is added to prevent the `_prepare_pos_message` method for the current quant from triggering unnecessarily if no config is found.

GabbasovDinar:
@PhilDL Thanks for your suggestions, I applied these changes and it looks much better now

PhilDL (approved):

hugho-ad (approved):

hugho-ad:
@ivantodorovich 

what do you think?

ready to merge?

ivs-cetmix:
@OCA/pos-maintainers could you please have a look into this PR?

gregory-moka:
Why don't you use the Info tab ? You can display what you want without overloading the POS view.

> Thanks for the contribution, the code looks okay to me, but I have some doubts about the UX, it looks very cluttered with 2 warehouses:
> 
> ## Actual
> CleanShot 2023-02-16 at 15 43 34
> 
> ## Proposal
> This is a suggestion of UX changes I tried, changing a few things. Feel free to follow it or not, depending on general feedback. It would look like that:
> 
> CleanShot 2023-02-16 at 16 51 02
>

ivs-cetmix:
> Why don't you use the Info tab ? You can display what you want without overloading the POS view.

Which info tab do you mean? Could you please provide some picture?

liebana:
I don't know if I'm missing something, but this is something already featured in: https://github.com/OCA/pos/tree/16.0/pos_product_quick_info

(I think that's what @gregory-moka meant @isserver1 )

ivs-cetmix:
> I don't know if I'm missing something, but this is something already featured in: https://github.com/OCA/pos/tree/16.0/pos_product_quick_info
> 
> (I think that's what @gregory-moka meant @isserver1 )

Yup. This is also our module that we have contributed earlier. However the current one has completely different approach. 
Just imagine if you need to have a quick overlook at your current stock. Clicking each product to check quantity available might take tons of time.  And here you can easily see which products are available right now without doing any clicks.
Meanwhile this module was originally developed for Odoo 13 and our customer has been happily using it for 2 years already.
So we agreed to open source it for Odoo 16 and share with the community.

Issue #849, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] pos_access_right - By @geomer198 - #957
- [ ] pos_cash_move_reason - By @quentinDupont - #1420
- [ ] pos_customer_display - By @legalsylvain - #1159
- [x] pos_customer_tree_view_vat - By @chienandalu - #1221
- [x] pos_default_partner - By @alv97 - #853
- [ ] pos_edit_order_line - By @flachica - #926
- [x] pos_escpos_status - By @nwinkelstraeter - #910
- [ ] pos_event_sale - By @BT-dmoreno - #1022
- [ ] pos_event_sale_registration_qr_code
- [x] pos_hide_banknote_button - By @legalsylvain - #1173
- [ ] pos_hide_cost_price_and_margin
- [ ] pos_hide_empty_category - By @legalsylvain - #1511
- [x] pos_hide_partner_info - By @huguesdk - #1354
- [x] pos_journal_image - By @legalsylvain - #1021
- [ ] pos_lot_barcode - By @yankinmax - #934
- [ ] pos_lot_selection - By @yankinmax - #935
- [ ] pos_margin_account_invoice_margin - By @legalsylvain - #1128
- [ ] pos_order_remove_line - By @suker - #925
- [x] pos_order_to_sale_order - By @legalsylvain - #854
- [x] pos_partner_firstname - By @victor-champonnois - #1027
- [x] pos_partner_no_create - By @mihien - #1422
- [ ] pos_payment_change - By @julenfl - #929
- [x] pos_payment_method_cashdro - By @chienandalu - #1161
- [x] pos_payment_terminal - By @nwinkelstraeter - #908
- [x] pos_picking_delayed - By @em230418 - #1227
- [x] pos_picking_load - By @HeliconiaSolutions - #1266
- [x] pos_price_to_weight - By @legalsylvain - #1146
- [x] pos_pricelist_technical - By @legalsylvain - #1130
- [ ] pos_product_display_default_code - By @suker - #927
- [x] pos_product_mergeable_line - By @legalsylvain - #1145
- [x] pos_product_multi_barcode - By @nguyenminhchien - #1037
- [ ] pos_product_template - By @zurabkituashvili - #1204
- [x] pos_receipt_hide_price - By @danielduqma - #851
[...]

PR #925, opened 2023-02-13
[16.0] [MIG] pos_order_remove_line: migration to 16.0

suker:
PR for migration from 14.0 to 16.0


suker:
Related task: #849

flachica (approved):

chienandalu (changes requested):
Check and fix pre-commit errors and warnings

suker:
> Check and fix pre-commit errors and warnings

Yup I think It doesn't like to use import and export in .js files...

chienandalu:
> > Check and fix pre-commit errors and warnings
> 
> Yup I think It doesn't like to use import and export in .js files...

When using ES6 you have to rename the files to .esm.js

suker:
> > > Check and fix pre-commit errors and warnings
> > 
> > 
> > Yup I think It doesn't like to use import and export in .js files...
> 
> When using ES6 you have to rename the files to .esm.js

That's true...
I saw it in migration to v15.0 guidelines...
image

Gracias :+1:

flachica (commented):
If delete the last line of ticket, exception is launched because event pos_restaurant.Orderline.selectLine arrived

suker:
> Check and fix pre-commit errors and warnings



> If delete the last line of ticket, exception is launched because event pos_restaurant.Orderline.selectLine arrived

What do you mind?

suker:
> Check and fix pre-commit errors and warnings

pre-commit is failing if we do not update version of isort to 12.0

image

is this error [related](https://github.com/OCA/l10n-spain/pull/2783#issuecomment-1408349114)?

ivantodorovich:
Should be fixed here:
- https://github.com/OCA/pos/pull/928

pos_order_remove_line/static/src/js/orderline.js
```diff
@@ -0,0 +1,20 @@
+/*
+ *  Copyright 2023 LevelPrime
+ *  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+ */
+
+odoo.define("pos_order_remove_line.Orderline", function (require) {
```
chienandalu: I think it was ok to move to ESM module style as Odoo is transitioning it as well. The only issue was to rename the file so pre-commit would swallow it...

chienandalu (commented):
Can you check @flachica comment?

flachica:
> > Check and fix pre-commit errors and warnings
> 
> > If delete the last line of ticket, exception is launched because event pos_restaurant.Orderline.selectLine arrived
> 
> What do you mind?


What I mean is that if:

1 I add a single line to the ticket
2 I use the trash button to delete the only line of the ticket that is there

The POS throws an exception

image

pos_order_remove_line/static/src/js/orderline.js
suker: Sure, I will leave it as it was before.
Thanks.

suker:
> > > Check and fix pre-commit errors and warnings
> > 
> > 
> > > If delete the last line of ticket, exception is launched because event pos_restaurant.Orderline.selectLine arrived
> > 
> > 
> > What do you mind?
> 
> What I mean is that if:
> 
> 1 I add a single line to the ticket 2 I use the trash button to delete the only line of the ticket that is there
> 
> The POS throws an exception
> 
> image

I see... but that error is throwed by `pos_restaurant` module, It should be fixed there

pos_order_remove_line/static/src/js/orderline.js
suker: Changes done!

flachica:
IMHO I think the bug is in this development. The pos_restaurant module is maintained by Odoo SA and without this module it would not fail. I am not yet knowledgeable enough to suggest you the necessary modification. I am investigating and as soon as I can I will make the suggestion. Regards

chienandalu:
Taking a quick glance at the code it looks to me that a glue module would be needed to make it compatible with `pos_restaurant` returning to `super` in case there wouldn't be a line `id`: https://github.com/odoo/odoo/blob/880afb3006ece93ee6575c977233bfb61081bf3a/addons/pos_restaurant/static/src/js/Screens/ProductScreen/Orderline.js#L27-L39

It can go to the Roadmap section in any case, as not every instance use the restaurant features.

suker:
> IMHO I think the bug is in this development. The pos_restaurant module is maintained by Odoo SA and without this module it would not fail. I am not yet knowledgeable enough to suggest you the necessary modification. I am investigating and as soon as I can I will make the suggestion. Regards

Thanks you for your opinion, after that I did some modifications to avoid the error.
It also deletes reward lines when loyalty programs are active.

suker:
> Taking a quick glance at the code it looks to me that a glue module would be needed to make it compatible with `pos_restaurant` returning to `super` in case there wouldn't be a line `id`: https://github.com/odoo/odoo/blob/880afb3006ece93ee6575c977233bfb61081bf3a/addons/pos_restaurant/static/src/js/Screens/ProductScreen/Orderline.js#L27-L39
> 
> It can go to the Roadmap section in any case, as not every instance use the restaurant features.

I found another way to avoid the error. It's already pushed, can you take a look?

Thanks :+1:

flachica:
Tested and working. Thank you very much

pos_order_remove_line/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Roberto Fichera <roberto.fichera@levelprime.com>
+* Iván Todorovich <ivan.todorovich@gmail.com>
+* Foram Shah <foram.shah@initos.com>
```
chienandalu, line 3: Add yourself to contributors

pos_order_remove_line/static/src/js/orderline.esm.js
```diff
@@ -0,0 +1,30 @@
...
+            if (order) {
+                ev.stopPropagation();
+                ev.preventDefault();
+                this.selectLine();
+                order.remove_orderline(order.get_selected_orderline());
+                this.checkRewardLines(order);
+            }
+        }
+
+        checkRewardLines(order) {
```
chienandalu, line 23: Document it as a depency-less support to pos_loyalty

pos_order_remove_line/readme/CONTRIBUTORS.rst
suker: Done.

pos_order_remove_line/static/src/js/orderline.esm.js
suker: Done.

Thanks.

chienandalu (approved):
Thanks :)

PR #930, opened 2023-02-22
[WIP][MIG] pos_favorite_product_categories: Migration to 16.0

flachica:
- Simple migration of [kmee](https://github.com/kmee/pos/tree/feature/pos_favorite_product_categories) work /cc @luismalta

I have yet to retrieve the complete history to respect the authorship of the project.


pedrobaeza:
Don't put plurals in module name, and you should respect attribution in commit author.

PR #935, opened 2023-02-27
[16.0][MIG] pos_lot_selection: Migration to 16.0

yankinmax:
Tested functionally and compared with v15


Camille0907 (approved):
LGTM from a technical point of view, but not tested functionnaly

ivantodorovich (approved):
Technically LG

As the module has no test tours, have you manually tested it ? 🤔

PR #944, opened 2023-03-15
[ADD] [14.0] pos_stock_scrap

felipezago:
This module implements a popup for creating stock scrap movements on the POS.


liebana:
I have conducted functional testing on this feature, and it's a nice feature. However, I believe the scope could be enhanced by allowing the movement of items from the default location to any desired location, not just to a designated scrap area. For instance, this feature could be used to reserve stock for a customer by moving it to a specific, physically designated location.

Additionally, it would be more intuitive to place a button above the payment buttons, rather than having the menu in the top bar, as this is more consistent with other add-ons in the repository.

Lastly, the README documentation could benefit from further refinement to provide clearer guidance and enhance overall understanding. :)

felipezago:
> I have conducted functional testing on this feature, and it's a nice feature. However, I believe the scope could be enhanced by allowing the movement of items from the default location to any desired location, not just to a designated scrap area. For instance, this feature could be used to reserve stock for a customer by moving it to a specific, physically designated location.
> 
> Additionally, it would be more intuitive to place a button above the payment buttons, rather than having the menu in the top bar, as this is more consistent with other add-ons in the repository.
> 
> Lastly, the README documentation could benefit from further refinement to provide clearer guidance and enhance overall understanding. :)

Thanks for the feedback! I will work to improve it.

About the location, you can actually configure the source and the desired location on the POS settings! It was my bad not mentioning it on the README though.

You can find it here:
image

liebana:
> Thanks for the feedback! I will work to improve it.
> 
> About the location, you can actually configure the source and the desired location on the POS settings! It was my bad not mentioning it on the README though.
> 
> You can find it here: image

I've seen it! But you've got a restriction to choose only virtual locations.

felipezago:
> > Thanks for the feedback! I will work to improve it.
> > About the location, you can actually configure the source and the desired location on the POS settings! It was my bad not mentioning it on the README though.
> > You can find it here: image
> 
> I've seen it! But you've got a restriction to choose only virtual locations.

Oh, I see. Although this is a great ideia, unfortunately it is beyond the scope of our implementation for this module, which is precisely implementing stock scrap movements on POS.

However, I have realocated the button's position and improved the README documentation as requested. Thank you again for the feedback.

ygcarvalh (approved):
LGTM

PR #945, opened 2023-03-20
[ADD] pos_margin_accounting_role

jumeldi74:
Adding new module to limit access rule for field Margin and Margin Percent for Accounting User only


legalsylvain (approved):
Makes sense.

code review, no test.

jumeldi74:
> Makes sense.
> 
> code review, no test.

Thank you

PR #947, opened 2023-03-21
[14.0][FIX] pos_margin: Frontend crash with pos_restaurant

IvanHRN:
If you have pos_restaurant installed the frontend crashes when you try to transfer a pos order from one table to another.


legalsylvain (approved):
Thanks. 
Fast tracking trivial patch.

/ocabot merge patch

PR #949, opened 2023-03-23
[16.0][ADD] pos_order_exclude_attribute_combination

sandipvyas-osi:


diogocsc:
https://pm.opensourceintegrators.com/web#id=29348&cids=1&menu_id=176&action=1098&model=project.task&view_type=form

pos_order_exclude_attribute_combination/static/src/scss/ProductConfiguratorPopup.scss
```diff
@@ -0,0 +1,22 @@
+.no_combination {
```
ivantodorovich, line 1: It'd be nice to move these css classes under a global one, to limit their domain.

A rule of thumb when you're developing a PoS module, is to define the css classes under the `.pos` class, which is added to the topmost parent element in the PoS UI.

```scss
.pos {
    .no_combination {
        ...
    }
    ...
}
```

It's even better if you can specify even more.
In this case, you know the class will be used in the ProductConfiguratorPopup, whose top element has the `product-configurator-popup` class. So you could write it like this:

```scss
.pos {
    .product-configurator-popup {
        .no_combination {
            ...
        }
        ...
    }
}
```

ivantodorovich (approved):

ivantodorovich:
Unit tests would be nice, too ^^

PR #950, opened 2023-03-23
[16.0][ADD] pos_restrict_price_discount_buttons

geomer198:
This module allows to restrict access to "Discount" and "Price" buttons on POS. It allows to use those buttons only to selected employees.

### DIFF
--- a/pos_restrict_price_discount_buttons/__manifest__.py
+++ b/pos_restrict_price_discount_buttons/__manifest__.py
@@ -1,0 +1,16 @@
+{
+    "name": "POS Restrict Price Discount Buttons",
+    "summary": "POS Restrict Price Discount Buttons",
+    "author": "Cetmix, Odoo Community Association (OCA)",
+    "version": "16.0.1.0.0",
+    "category": "Point of Sale",
+    "website": "https://github.com/OCA/pos",
+    "maintainers": ["geomer198", "CetmixGitDrone"],
+    "depends": ["pos_sale"],
+    "data": [
+        "views/res_config_settings_views.xml",
+    ],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": False,
+}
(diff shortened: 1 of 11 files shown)


pos_restrict_price_discount_buttons/models/res_config_settings.py
```diff
@@ -0,0 +1,26 @@
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    restrict_price_button = fields.Boolean(
+        related="pos_config_id.restrict_price_button", readonly=False
+    )
+    restrict_price_users_ids = fields.Many2many(
```
ivs-cetmix: Must be employee not user

pos_restrict_price_discount_buttons/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "POS Restrict Price Discount Buttons",
+    "summary": "POS Restrict Price Discount Buttons",
```
ivs-cetmix: Restrict access to 'Price' and 'Discount' buttons in POS

ivs-cetmix (changes requested):
Restriction exclusions exceptions must apply to Employee not to User.
This will allow bypassing it by switching employee in POS during the session.

pos_restrict_price_discount_buttons/models/res_config_settings.py
geomer198: Resolved

pos_restrict_price_discount_buttons/__manifest__.py
geomer198: Resolved

GabbasovDinar:
@geomer198 Thanks for the PR, but I think you just need to port the https://github.com/OCA/pos/tree/14.0/pos_access_right module from 14.0 to 16.0, please check.

PR #952, opened 2023-03-27
[ADD] pos_product_mergeable_line

legalsylvain:


pos_product_no_merge/static/src/js/db.js
```diff
@@ -0,0 +1,13 @@
+/*
+Copyright (C) 2023-Today: GRAP (<http://www.grap.coop/>)
+@author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+*/
+
+odoo.define('pos_tare.db', function (require) {
```
ivantodorovich: 🚫 blocking: without this fix, you will get unexpected errors if `pos_tare` is installed, as only one js file will be loaded -- can't have two with the same name 😓 

```suggestion
odoo.define('pos_product_no_merge.models', function (require) {
```

This file should probably be merged with `models.js`, as they're both modifying the core models.

pos_product_no_merge/static/src/js/models.js
```diff
@@ -0,0 +1,27 @@
...
+    models.Orderline = models.Orderline.extend({
+
+        can_be_merged_with: function(orderline){
+            if (orderline.product.pos_no_merge) {
+                return false;
+            }
+            return moduleOrderlineParent.prototype.can_be_merged_with.apply(this, arguments);
+        },
+
+    });
```
ivantodorovich: 💡 I recommend this pattern to deal with "super". I find it easier to read, specially when there are multiple metohds being overriden:

```suggestion
    const OrderlineSuper = models.Orderline.prototype;
    models.Orderline = models.Orderline.extend({
        can_be_merged_with: function(orderline){
            if (orderline.product.pos_no_merge) {
                return false;
            }
            return OrderlineSuper.can_be_merged_with.apply(this, arguments);
        },
    });
```

pos_product_no_merge/static/src/js/models.js
```diff
@@ -0,0 +1,27 @@
...
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+******************************************************************************/
+
+odoo.define('pos_product_no_merge.model', function (require) {
+    "use strict";
+
+    var models = require('point_of_sale.models');
+    var core = require('web.core');
+    var session = require('web.session');
+    var _t = core._t;
```
ivantodorovich:
```suggestion
    const models = require('point_of_sale.models');
```

Only models seems to be used.
The others are imported but not used AFAICS

pos_product_no_merge/static/src/js/models.js
```diff
@@ -0,0 +1,27 @@
...
+        can_be_merged_with: function(orderline){
+            if (orderline.product.pos_no_merge) {
+                return false;
+            }
+            return moduleOrderlineParent.prototype.can_be_merged_with.apply(this, arguments);
+        },
+
+    });
+
+});
```
ivantodorovich:
```suggestion
    return models;
});
```

To play nice with other modules. Makes it slightly easier to import this file to apply more overrides on top of it

pos_product_no_merge/views/templates.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0"?>
+<!--
+    Copyright 2023 - Today Sylvain LE GAL (https://twitter.com/legalsylvain)
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="assets_backend" name="pos_product_no_merge" inherit_id="point_of_sale.assets">
```
ivantodorovich:
```suggestion
    <template id="point_of_sale_assets" name="pos_product_no_merge" inherit_id="point_of_sale.assets">
```

nitpicking: it's not for the backend ;)

ivantodorovich (changes requested):
Thanks! LG

1 blocking comment
the others are just suggestions

legalsylvain:
hi @ivantodorovich. Thanks a lot for your review. I have a doubt about the implementation. What is the best.

A) pos_mergeable_line = fields.Boolean(default=True)
B) pos_no_mergeable_line = fields.Boolean(default=False)

On reflection, I think that solution A avoids the negation. When do you think?

ivantodorovich:
Yes, option A is better IMO

pos_product_no_merge/static/src/js/db.js
ivantodorovich: ⚠️ this is still unresolved

legalsylvain: ow ! thanks !

pos_product_mergeable_line/static/src/js/db.js
```diff
@@ -0,0 +1,13 @@
...
+@author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+*/
+
+odoo.define("pos_product_mergeable_line.db", function (require) {
+    "use strict";
+
+    var models = require("point_of_sale.models");
+
+    models.load_fields("product.product", ["pos_mergeable_line"]);
```
ivantodorovich: IMO this line should be moved to `pos_product_mergeable_line/static/src/js/models.js`, and you can drop the file.

The name `pos_product_mergeable_line.db` suggests overrides on `point_of_sale.db`, but it's just "extending" models.js

ivantodorovich (approved):

pos_product_mergeable_line/static/src/js/db.js
legalsylvain: Ow thanks ! Totally right. Done.

Note : Oca modules requires some clean regarding that topic.

- https://github.com/OCA/pos/blob/12.0/pos_tare/static/src/js/db.js#L12
- https://github.com/OCA/pos/blob/12.0/pos_multi_ean/static/src/js/db.js#L7
- https://github.com/OCA/pos/blob/12.0/pos_supplierinfo_barcode/static/src/js/db.js#L7
- https://github.com/OCA/pos/blob/12.0/pos_supplierinfo_search/static/src/js/db.js#L7

pos_product_no_merge/static/src/js/db.js
legalsylvain: Fixed.

PR #954, opened 2023-03-29
[IMP] pos_event_sale: Add hook to store currency_id in database for each existing pos_order_line before installation

Camille0907:
This helps to avoid a Memory Error when installing this module on an instance with a big amount of existing pos_order_line

### DIFF
--- a/pos_event_sale/hooks.py
+++ b/pos_event_sale/hooks.py
@@ -2,7 +2,38 @@
 # @author Iván Todorovich <ivan.todorovich@gmail.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import logging
+
 from odoo import SUPERUSER_ID, api
+from odoo.tools.sql import column_exists, create_column
+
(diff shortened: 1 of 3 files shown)


pos_event_sale/hooks.py
```diff
@@ -2,7 +2,37 @@
 # @author Iván Todorovich <ivan.todorovich@gmail.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import logging
+
 from odoo import SUPERUSER_ID, api
+from odoo.tools.sql import column_exists, create_column
+
+_logger = logging.getLogger(__name__)
```
ivantodorovich, line 10:
```suggestion
from odoo import SUPERUSER_ID, api
from odoo.tools.sql import column_exists, create_column
```

`_logger` is imported but it doesn't seem to be used

ivantodorovich (approved):

yankinmax (approved):

PR #956, opened 2023-03-31
[14.0][ADD] pos_screen_lock

felipezago:
This module adds a PoS setting that if enabled, logs out of POS in case of inactivity after the configured time.


cvinh:
Thanks for the work but did you see that module ? https://github.com/OCA/pos/tree/12.0/pos_timeout

felipezago:
> Thanks for the work but did you see that module ? https://github.com/OCA/pos/tree/12.0/pos_timeout

Yes, but the functionality i'm implementing is a timeout for logging out of PoS automatically when the operator is idle. Thanks for the feedback though!

ygcarvalh (approved):

PR #962, opened 2023-04-04
[15.0] [FIX+IMP] `pos_event_sale`: compatibility with pos certification modules

ivantodorovich:
Negated lines are particularly useful with PoS modules that disable the removal
of pos orderlines through the use of ``disallowLineQuantityChange``.

This is the case of `l10n_fr_pos_cert` module, for example. But there are other
core modules doing it, too.

In this case, added order lines can't be removed. Instead, the cashier is
encouraged to add a negative to 'cancel' the original one.

e.g.:

Adds a line with 10 tickets.
Then, adds a line with -10 tickets.

This should be almost the same as removing the first line: no actual
registration should be confirmed in the end, as we're reverting this sale.

----

Easier to review commit by commit

### DIFF
--- a/pos_event_sale/models/event_registration.py
+++ b/pos_event_sale/models/event_registration.py
@@ -49,6 +49,13 @@ class EventRegistration(models.Model):
                 rec.payment_status = "not_paid"
         return res
 
+    def _check_auto_confirmation(self):
+        # OVERRIDE to disable auto confirmation for registrations created from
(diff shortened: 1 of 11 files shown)


yankinmax:
@ivantodorovich thanks for the great and necessary improvement :+1:

yankinmax (approved):

Camille0907 (approved):
LGTM, thanks

PR #965, opened 2023-04-09
[16.0][ADD] pos_margin_access_right: Add new module

jumeldi74:


cvinh:
Hello
Thank you for the module
I think that it would be worth improving pos_margin with the functionalities you introduced than creating a new module...
@legalsylvain @ivantodorovich what is your opinion ?

PR #968, opened 2023-04-12
[ADD] pos_disable_product_attribute module

ecino:
New module for disabling specific attributes from POS (and avoid annoying popups).


ecino:
I need help understanding what is failing in the pre-commit check? Do I have to fix a line too long or not? I thought the script is supposed to do it automatically or not?

PR #976, opened 2023-04-19
[14.0][MIG] pos_order_to_sale_order: Migration to 14.0

zamberjo:


chienandalu (commented):
Thanks for this @zamberjo :+1: 

A tiny functional issue I found is that when you convert the pos order to sale order it isn't flushed from the PoS order list, so they remain there until manually removed. May you have to destroy it.

belenaures:
Tested and found that everything works correctly. It is true that as for the orders created (as [chienandalu] says), could be deleted from the PoS order list once they have been created.

zamberjo:
hi  @chienandalu @belenaures  done, now the order is removed from the list once the sale.order is created

chienandalu (approved):
Thanks! :+1:

chienandalu:
@belenaures add a proper Github approval

belenaures (approved):
LGTM

zamberjo:
Hi @pedrobaeza Can you check it?

PR #977, opened 2023-04-19
[13.0][MIG] pos_user_restriction: Migration to 13.0

ghost:


pos_user_restriction/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Point of Sale - Restrict users",
     "summary": "Restrict some users to see and use only certain points of sale",
-    "version": "12.0.1.0.2",
+    "version": "13.0.1.0.2",
```
yankinmax: `"version": "13.0.1.0.0"`,

yankinmax: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#tasks-to-do-in-the-migration

yankinmax (approved):

Camille0907 (approved):

SilvioC2C (approved):
LGTM overall, remember to squash the fixup commit :wink:

gurneyalex (changes requested):
Tested the module, there are issues with the permissions:

some access which are given by the base odoo POS module to group "Point of Sale / User" need to be given also to the new group:

* point of sale payment method
* decimal precision
* point of sale payment
* account cash rounding

pos_user_restriction/models/pos_config.py
```diff
@@ -0,0 +1,12 @@
...
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+    assigned_user_ids = fields.Many2many(
+        "res.users",
+        string="Assigned users",
+        help="Restrict some users to only access their assigned points of sale. "
+        "In order to apply the restriction, the user needs the "
+        "'User: Assigned POS Only' group",
+    )
```
gurneyalex, line 13: you will also need to redefine the field group_pos_user_id (see https://github.com/odoo/odoo/blob/198977d40ea7f00f50ccc8dfb48026012654a595/addons/point_of_sale/models/pos_config.py#L188) the default value must be redefined to be the new group "pos_config_assigned_users" (otherwise it is not possible to open a session). 

I would also add a post install hook that updates all existing pos configurations to change the value of group_pos_user_id : if it is the "normal" default, then change it to the new default. Leave configurations using another value untouched.

gurneyalex (changes requested):

pos_user_restriction/models/pos_config.py
gurneyalex: update: use the version available in 14.0 https://github.com/OCA/pos/blob/6a3688f19773aa3f90e8328608a5dc4e75b16b47/pos_user_restriction/models/pos_config.py#L14-L29

More generally check this commit: https://github.com/OCA/pos/commit/d5ca514cfc2a12442468f19437d691e9acc2c3cf

Camille0907 (approved):
LGTM, thanks

legalsylvain:
Thanks for your contribution

/ocabot merge nobump

PR #980, opened 2023-04-21
[14.0][MIG] pos_tare: Migration to 14.0

baimont:
https://github.com/OCA/pos/issues/543


sanrav:
See my issue created for the V12 => https://github.com/OCA/pos/issues/984

In V14, I have the same issue with the tare barcode reading in the scaling screen but the difference with the V12 is that the product is added to the order : 

image

baimont:
BUG to fix : when scanning a normal product, it sets the quantity to 2

apineux (approved):

FrancoMaxime (approved):

pos_tare/static/src/js/Screens/ProductScreen/ProductScreen.js
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            async _barcodeTareAction(code) {
+                var last_orderline = this.currentOrder.get_last_orderline();
+                if (last_orderline) {
+                    last_orderline.set_tare(code.value, true);
+                }
+            }
+
+            async _getAddProductOptions(product) {
```
legalsylvain, line 24: rewriting all this function is not a good option.  
I think you can return weight + tare in the payload of the "ScaleScreen", and then overload _getAddProductOptions just to split the quantity field into the two weight and tare. What do you thing ?

legalsylvain (changes requested):

pos_tare/static/src/js/Screens/ProductScreen/ProductScreen.js
baimont: Hello @legalsylvain and thanks for your comment.
As I remember I really tried not to do that but I didn't have the choice. I cannot tell you why exactly as I currently work on other projects and haven't touch this for months. If you are sure of you, you can maybe propose some changes and I'll gladly take a look (and test).

PR #981, opened 2023-04-21
[14.0][FIX] pos_order_return: make returns properly

chienandalu:
Before this PR the returns weren't functional when a stock picking was involved.

cc @Tecnativa TT42730

please review @pedrobaeza @CarlosRoca13

### DIFF
--- a/pos_order_return/README.rst
+++ b/pos_order_return/README.rst
@@ -85,9 +85,10 @@ invoiced, a refund invoice will be made.
 Known issues / Roadmap
 ======================
 
-When migrating the module ``pos_order_return`` in version > 12.0 please merge
-both modules ``pos_order_return`` and ``pos_order_return_traceability`` into a
-single module.
+* When migrating the module ``pos_order_return`` in version > 12.0 please merge
+  both modules ``pos_order_return`` and ``pos_order_return_traceability`` into a
+  single module.
+* Stock update on session close is uncovered right now.
 
 Bug Tracker
 ===========
--- a/pos_order_return/models/pos_order.py
+++ b/pos_order_return/models/pos_order.py
@@ -2,10 +2,9 @@
 # Copyright 2018 David Vidal <david.vidal@tecnativa.com>
 # Copyright 2018 Lambda IS DOOEL <https://www.lambda-is.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-
-
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tests import Form
 
 
 class PosOrder(models.Model):
@@ -111,9 +110,13 @@ class PosOrder(models.Model):
 
     def _create_picking_return(self):
         self.ensure_one()
-        picking = self.returned_order_id.picking_ids
-        ctx = dict(self.env.context, active_ids=picking.ids, active_id=picking.id)
-        wizard = self.env["stock.return.picking"].with_context(ctx).create({})
+        return_form = Form(
+            self.env["stock.return.picking"].with_context(
+                active_id=self.returned_order_id.picking_ids.id,
+                active_model="stock.picking",
+            )
+        )
+        wizard = return_form.save()
         # Discard not returned lines
         wizard.product_return_moves.filtered(
             lambda x: x.product_id not in self.mapped("lines.product_id")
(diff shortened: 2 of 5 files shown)


pedrobaeza (approved):

CarlosRoca13 (approved):

pedrobaeza:
@chienandalu please check tests

legalsylvain (changes requested):
Hi. Using test framework (out of tests) is not a good idea in term of performance. 
In the meantime, i dont understand why it is needed in this case. Could you avoid it ?
Thanks !

chienandalu:
> Hi. Using test framework (out of tests) is not a good idea in term of performance.

That's normally true and we found it out the hard way time ago :) In this case, anyway, we're dealing with a single record and the overhead can be assumed IMO. In exchange, we cover onchange workflows by default without pulling extra deps like the onchange helper module.

legalsylvain:
>  In this case, anyway, we're dealing with a single record and the overhead can be assumed IMO

could you provide before / after performance ? 

could you also take a look on the CI that is red ?

chienandalu:
> > In this case, anyway, we're dealing with a single record and the overhead can be assumed IMO
> 
> could you provide before / after performance ?

The module was broken. So the gain is 100% :sweat_smile: Jokes aside in our experience is reliable for light singleton operations like this (run a wizard) 
 
> could you also take a look on the CI that is red ?

It should be fixed now

pedrobaeza:
Please push again, as there's no CI checks.

chienandalu:
All green

pedrobaeza (approved):

zamberjo:
Thanks for the fix, it seems that the error no longer occurs.

As far as I can see, the picking created in the return remains in `Ready` status, while the normal order picking is changed to `Done`. Is it possible to use the PR to transfer the picking? do you think that would be ok?

Note: this only occurs when we choose 'In real time', otherwise if we choose 'At the session closing' option all pickings are transferred.

chienandalu:
> As far as I can see, the picking created in the return remains in Ready status, while the normal order picking is changed to Done. Is it possible to use the PR to transfer the picking? do you think that would be ok?

Sure. It should be close. Check again, please

zamberjo (approved):
Thanks! :clap:

pedrobaeza:
@legalsylvain please unblock

pedrobaeza:
Merging as no further answer from @legalsylvain. If a serious performance problem is faced, we can proceed other way:

/ocabot merge patch

legalsylvain:
> Merging as no further answer from @legalsylvain.

@pedrobaeza, I was looking for the long debate with @rousseldenis , @Cedric-Pigeon, @lmignon,  which had concluded (except you) not to use odoo.test.Forms (outside the tests of course) and I could not find it.
I just found it again, so I put this link here to allow everyone to understand the issues of the debate "onchange_helper" vs odoo.test.Forms : https://github.com/OCA/sale-workflow/pull/1239

I din't changed my mind (follow OCA members conclusion, see link below). I have nothing to add.

Merging a PR in this context, because the OCA Representative of the repo (and original author of the module) who is blocking the PR has not responded for a few days is not correct.

> The module was broken. So the gain is 100% sweat_smile Jokes aside in our experience is reliable for light singleton operations like this (run a wizard)

@chienandalu : It would have been interesting to spend some time to analyze the execution time with / without ``odoo.test.Form``, to have some figures.
In [my previous benchmark](https://github.com/OCA/sale-workflow/pull/1239#issuecomment-690673471), the execution time was multiplied by 8 to 11, which is not insignificant. 

That's being said :

- Let's see the good side of things! this PR fixes a bug, it's a good news.
- _A priori_, return order is not a common operation, and as said by @chienandalu realized on a singleton. The degradation of the UX due to the slowness introduced by this PR, although undesirable, is not dramatic.
- This module is in the ``Alpha``  development status
- With the new "Refund" button present in Odoo 16.0 CE, this module will maybe not be usefull anymore, or at least, highly refactored.

so I will add soon in the ``ROADMAP.rst`` file of the module a text to indicate the need to replace the use of odoo.test.Forms by onchange_helper if we want to improve UX.

pedrobaeza:
That thread is because they want to do operations in batch. This is for only one record. And all that context was taken into account, but this was required to fix the problem, and it was paralyzed by you for 2 weeks, so I consider the merging correct. And your measures of 8x/11x of increase were on sale.order.line, where lots of computations/onchanges are done, not here in this specific wizard.

PR #996, opened 2023-05-08
[14.0][ADD] pos_return_voucher

zamberjo:
This module extends the functionality of the PoS by allowing the issuance of
discount vouchers (both fronted and backend) when a return is made. This means
that the customer receives a ticket with a barcode that can be redeemed for the
amount of the return during the time configured in the POS (in days).


belenaures (approved):
LGTM

zamberjo:
I just saw that there was an error when reprinting the return ticket, the barcode I was printing did not contain "Refund" but "Order". 

I have modified the code so that it extracts the uid from the order reference excluding that part.

I have also added the humanreadable flag in case the store does not have a barcode reader (or it does not work).

zamberjo:
@chienandalu  I would appreciate it if you would give your feedback on the module, thanks!

chienandalu (commented):
Hi @zamberjo Wouldn't it be better to backport the pos_coupon/pos_loyalty functionality on this regard from v15/16?

zamberjo:
> Hi @zamberjo Wouldn't it be better to backport the pos_coupon/pos_loyalty functionality on this regard from v15/16?

We already analyzed those modules and saw that they do not implement the same concept.

In this module the returned ticket is controlled as a voucher with a validity date for a next purchase, thus avoiding to return physical money to the customer. In this way both the return and the orders that redeem the voucher, are linked to the same payment method, being able to be accounted for.

In the modules that you propose as far as we understand a purchase of a gift card that can be redeemed is made, not relating it to the return.

zamberjo:
Rebase + supporting 
image

zamberjo:
Rebase + improvment:
 - it is allowed not to indicate days of validity of the vouchers, in this way they will never expire.

manuelcalerosolis:
Great module LGTM

manuelcalerosolis (approved):

zamberjo:
Hi, this module is being migrated to major versions (https://github.com/OCA/pos/pull/1140), can you keep an eye on this again? thanks.

Ping: @legalsylvain @chienandalu @rafaelbn @ivantodorovich

chienandalu:
Thanks @zamberjo . Merging as it is, although I still think it should be prepared to merge into the existing functionality in v16 (comenting now in the v16 migration)

/ocabot merge nobump

zamberjo:
thanks @chienandalu

PR #997, opened 2023-05-10
[16.0][ADD] pos_product_sequence

sandipvyas-osi:


pos_product_sequence/models/product_template.py
```diff
@@ -0,0 +1,28 @@
...
+        Every product should ideally have an unique sequence.
+        Default sequence (10000) should only be used for DB first product.
+        As we don't resequence the whole tree (as `sequence` does), this field
+        might have negative value.
+        """
+        self._cr.execute("SELECT MAX(pos_sequence) FROM %s" % self._table)
+        max_sequence = self._cr.fetchone()[0]
+        if max_sequence is None:
+            return 10000
+        return max_sequence + 5
```
dreispt, line 28: Do we need these computations? Was that copied from Website logic by chance?
I expected a default fixed sequence number, such as 1000.
Products with the same sequence number would just follow whatever default sort criteria is, if that's not a complication.

ivantodorovich: I totally agree. A simple default value should be enough here.

ivantodorovich (changes requested):

PR #1005, opened 2023-05-31
[14.0][IMP] pos_no_cash_bank_statement: Filter payment methods configured in the POS.

zamberjo:
In the case of having several payment methods and several POS (not all of them were configured with all the payment methods), when closing the session, 1 `account.bank.statement` was being created for each payment method even if it was not configured in the POS.

With this PR, we filtered those payment methods configured in the POS.

### DIFF
--- a/pos_no_cash_bank_statement/models/pos_session.py
+++ b/pos_no_cash_bank_statement/models/pos_session.py
@@ -23,12 +23,12 @@ class PosSession(models.Model):
         # I have to do that because the method
         # _create_cash_statement_lines_and_cash_move_lines()
         # reads payment_method_id.cash_journal_id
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

anmarmo1 (commented):
LGTM

chienandalu (approved):

chienandalu:
Does it need to be forward ported?

zamberjo:
Hi @chienandalu I don't know what you mean, this module is not migrated in later versions. As stated in the comment https://github.com/OCA/pos/issues/849#issuecomment-1367608294 it is not necessary to migrate it.

chienandalu:
Ok, that's what I meant :smile: Thanks for the heads up

PR #1011, opened 2023-06-09
[ADD] pos_loyalty_redeem_payment

danielduqma:
This module allows gift cards to be applied as a payment method when customer uses it. Provides the necessary flexibility to accomplish with [European Regulations](https://eur-lex.europa.eu/eli/dir/2016/1065/oj) (Council Directive (EU) 2016/1065 of 27 June 2016 amending Directive 2006/112/EC).

Related issue: https://github.com/OCA/l10n-spain/issues/2586


pedrobaeza:
For trying to accommodate to the usual naming scheme, I would put it as `pos_payment_from_loyalty`. What do you think? Do you like it?

Anyway, thanks for this useful one.

danielduqma:
Thanks for the feedback! It was named with `pos_loyalty` as prefix following guidelines, because it is an extension of this module and it's its only dependency. In any case, we are kind to rename it if its needed.

Regards.

danielduqma:
FYI @moduon @rafaelbn @fcvalgar @SoniaViciana

pos_loyalty_redeem_payment/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,18 @@
+To allow a loyalty program of type *Gift Card* to be redeemed as a payment
+method, first select *Payment Method* as *Redemption Method*:
+
+.. image:: static/img/redeemMethod.png
```
pedrobaeza: Please use relative paths with `..`.

pedrobaeza:
OK, I see. I was thinking in `area` + `model` +  `function`, but depending in the other module makes the prefix `pos_loyalty_` as mandatory, so it can be OK this way.

pos_loyalty_redeem_payment/readme/CONFIGURE.rst
danielduqma: Done!

fcvalgar (approved):
LGTM, thank you @danielduqma 


Sorry for my previous message, I hadn't seen the screenshot.

fcvalgar:
@danielduqma
The first sentence of the usage is a bit confusing: "In the payment screen of PoS interface, after selecting configured payment method a popup requests loyalty card code:"

Odoo-POS

Thank you.

SoniaViciana (approved):
LGTM. Functional review OK

PR #1016, opened 2023-06-20
[14.0][FIX] pos_order_return: Returns and sales are allowed on the same order.

zamberjo:
Currently if you use the `pos_order_return` module together with `pos_order_mgmt` you can perform order returns from the frontend.

From here, currently the frontend user can make a return and take advantage of the same order to make a sale (for example a replacement of a product, he returns one and we sell him another of the same, higher or lower price).

The current problem is that `pos_order_return` does not allow this, when the new order is persisted only the lines of the return are processed, ignoring those of the new sale.

With this change, we support these types of sales, so that an order can end up having 2 pickings, one with the returns of a previous order and another delivery note with the new sale.

### DIFF
--- a/pos_order_return/tests/test_pos_order_return.py
+++ b/pos_order_return/tests/test_pos_order_return.py
@@ -49,7 +49,17 @@ class TestPOSOrderReturn(common.SavepointCase):
                 "taxes_id": False,
             }
         )
+        cls.product_3 = cls.env["product.product"].create(
+            {
+                "name": "Test product 3",
+                "standard_price": 1.0,
+                "type": "product",
+                "pos_allow_negative_qty": True,
+                "taxes_id": False,
+            }
+        )
         cls.PosOrder = cls.env["pos.order"]
+        cls.PosOrderLine = cls.env["pos.order.line"]
         cls.pos_config = cls.env.ref("point_of_sale.pos_config_main")
         cls.pos_config.write(
             {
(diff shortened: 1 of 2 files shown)


zamberjo:
@chienandalu  I think you recently work on this part, can you take a look at it?

belenaures (approved):
After several tests done find it ok,  all orders are showed in the backend and it executes correctly the pickings.

pos_order_return/tests/test_pos_order_return.py
```diff
@@ -65,8 +75,8 @@ def setUpClass(cls):
                 "partner_id": cls.partner.id,
                 "pricelist_id": cls.partner.property_product_pricelist.id,
                 "amount_tax": 0,
-                "amount_total": 2700,
-                "amount_paid": 2700,
+                "amount_total": 1350,
```
chienandalu, line 78: I'm a little bit confused by this changes in the tests setup. Are they really necessary. I mean, it's hard to tell if the changes break the tests as they were before :thinking:

chienandalu (commented):
Thanks @zamberjo :smile: 

Tested functionally ok :+1: Some questions/doubts:

Do you know how does Odoo deal with this in next versions? (if they do at all)

pos_order_return/tests/test_pos_order_return.py
zamberjo: It is true that I should not have touched them, but when doing the new tests I realized that they were not consistent, qty * price_unit with respect to the total of the line (price_subtotal), so I ended up deciding to fix it, do you think we could try it with the previous values in __local__ and if they pass the tests and you see the correct changes leave these new ones?

zamberjo:
> Do you know how does Odoo deal with this in next versions? (if they do at all)

We have not yet started to analyze the POS in v15-16, I will take a look and comment on it as soon as I can.

pos_order_return/tests/test_pos_order_return.py
chienandalu: Ok, thanks for the explanations :)

chienandalu (approved):

PR #1019, opened 2023-06-29
16.0 fix eslint

legalsylvain:
Hi all.
rational : when I contribute to OCA, when I run pre-commit locally, there is a lot of useless warnings.
So this PR : 
- fix 2 trivials eslint warn in 2 modules
- remove no-shadow because in my opinion, the syntax is valid, and used a lot of time in the OCA/pos project. 

Any point of view, @ivantodorovich ?

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -126,7 +126,7 @@ rules:
   no-self-compare: warn
   no-sequences: warn
   no-shadow-restricted-names: warn
-  no-shadow: warn
+  # no-shadow: warn
   no-sparse-arrays: warn
   no-sync: warn
   no-this-before-super: warn
--- a/pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.esm.js
+++ b/pos_order_reorder/static/src/js/Screens/TicketScreen/ControlButtons/ReorderButton.esm.js
@@ -1,8 +1,8 @@
 /** @odoo-module **/
 
+import {Orderline} from "point_of_sale.models";
 import PosComponent from "point_of_sale.PosComponent";
 import Registries from "point_of_sale.Registries";
-import {Orderline} from "point_of_sale.models";
 
 class ReorderButton extends PosComponent {
     get isEmptyOrder() {
--- a/pos_payment_terminal/static/src/js/models.js
+++ b/pos_payment_terminal/static/src/js/models.js
@@ -25,7 +25,9 @@ odoo.define("pos_payment_terminal.models", function (require) {
             async after_load_server_data() {
                 for (var payment_method_id in this.payment_methods) {
                     var payment_method = this.payment_methods[payment_method_id];
-                    if (payment_method.use_payment_terminal == "oca_payment_terminal") {
+                    if (
+                        payment_method.use_payment_terminal === "oca_payment_terminal"
+                    ) {
                         this.config.use_proxy = true;
                     }
                 }


chienandalu:
Out of curiosity. Can you point to an example where you consider the no-shadow directive useless?

legalsylvain:
Hi @chienandalu. Thanks for your answer. If you check https://github.com/OCA/pos/actions/runs/5411682450/jobs/9834749032, you can see, that in many modules, the same code structure generate that warning. 
There warning are present in all the PoS module that includes javascript files that overload core PoS :  pos_global_discount_in_line, pos_lot_barcode, pos_lot_selection, pos_payment_terminal, pos_escpos_status, pos_lot_barcode, pos_lot_selection, pos_order_remove_line, pos_product_quick_info, pos_stock_available_online

I remember I tried to write differently my code to avoid this warning but without success. The "correct" syntax create error if many modules overload the same javascript object.
In that PR, I explicitely set ``no-shadow`` exception, but IMO, it adds extra useless lines.

https://github.com/OCA/pos/pull/859/files#diff-653d42f63ebb5228bce50e54e155771c3679aef4192d50ff1405700ffa103626R16

2 options : 
A) there is a good way to "write" PoS Overload, so it should be described, and we should set no-shadow as error. (and not warn)
B) There is not a good way to write Pos Overload, so it eslint should be disabled for that error. 

I propose option B because I don't think option A is possible. But I'm really not a javascript expert, so if you have a proposal, I'm OK to change the current PR.

```
/home/runner/work/pos/pos/pos_global_discount_in_line/static/src/js/GlobalLineDiscount.js
  7:33  warning  'DiscountButton' is already declared in the upper scope on line 4 column 11  no-shadow

/home/runner/work/pos/pos/pos_lot_barcode/static/src/js/Screens/ProductScreen.js
  14:41  warning  'ProductScreen' is already declared in the upper scope on line 8 column 11  no-shadow

/home/runner/work/pos/pos/pos_lot_selection/static/src/js/EditListPopup.js
  11:37  warning  'EditListPopup' is already declared in the upper scope on line 8 column 11  no-shadow

/home/runner/work/pos/pos/pos_lot_selection/static/src/js/ProductScreen.js
  11:38  warning  'ProductScreen' is already declared in the upper scope on line 8 column 11  no-shadow

/home/runner/work/pos/pos/pos_lot_selection/static/src/js/models.js
  11:38  warning  'PosGlobalState' is already declared in the upper scope on line 8 column 12  no-shadow

/home/runner/work/pos/pos/pos_payment_terminal/static/src/js/models.js
  22:47  warning  'PosGlobalState' is already declared in the upper scope on line 16 column 12                    no-shadow
  23:15  warning  'PosPaymentTerminalPosGlobalState' is already declared in the upper scope on line 22 column 11  no-shadow
  37:40  warning  'Payment' is already declared in the upper scope on line 16 column 28                           no-shadow
  38:15  warning  'PosPaymentTerminalPayment' is already declared in the upper scope on line 37 column 11         no-shadow

/home/runner/work/pos/pos/pos_escpos_status/static/src/js/ProxyStatus.js
  11:32  warning  'ProxyStatus' is already declared in the upper scope on line 9 column 9          no-shadow
  12:15  warning  'EscposProxyStatus' is already declared in the upper scope on line 11 column 11  no-shadow

/home/runner/work/pos/pos/pos_lot_barcode/static/src/js/Popups/EditListPopup.js
  12:41  warning  'EditListPopup' is already declared in the upper scope on line 8 column 11  no-shadow

/home/runner/work/pos/pos/pos_lot_selection/static/src/js/OrderWidget.js
  11:32  warning  'OrderWidget' is already declared in the upper scope on line 9 column 11  no-shadow

/home/runner/work/pos/pos/pos_order_remove_line/static/src/js/orderline.esm.js
  10:23  warning  'Orderline' is already declared in the upper scope on line 7 column 8  no-shadow

/home/runner/work/pos/pos/pos_product_quick_info/static/src/js/Screens/ProductScreen/ProductItem.js
  7:35  warning  'ProductItem' is already declared in the upper scope on line 4 column 11           no-shadow
  8:15  warning  'QuickInfoProductItem' is already declared in the upper scope on line 7 column 11  no-shadow

/home/runner/work/pos/pos/pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductItem.esm.js
  8:27  warning  'ProductItem' is already declared in the upper scope on line 3 column 8       no-shadow
  9:11  warning  'StockProductItem' is already declared in the upper scope on line 8 column 7  no-shadow

/home/runner/work/pos/pos/pos_stock_available_online/static/src/js/Screens/ProductScreen/ProductsWidget.esm.js
  6:30  warning  'ProductsWidget' is already declared in the upper scope on line 3 column 8       no-shadow
  7:11  warning  'StockProductsWidget' is already declared in the upper scope on line 6 column 7  no-shadow
```

chienandalu:
I see. I'd say that the warning is correct in those cases as the object in the arrow function argument doesn't have any effect as the extended object is already a global constant. But maybe I'm missing the js magic behind this type of declaration :sweat_smile: 

So for me this code:

```javascript
    const GlobalLineDiscount = (DiscountButton) =>
        class extends DiscountButton {
        ...
```

would be perfectly valid like this:

```javascript
    const GlobalLineDiscount = () =>
        class extends DiscountButton {
        ...
```

chienandalu:
Taking a look at the future v17, looks like they removed this class inheritance mode (so this won't be an issue in the future): https://github.com/odoo/odoo/commit/b119cca37b2f345e17ef62de9aacb638e1c075ab

And being that the standard enforces this way of inheritance up to v16, maybe the warning removal is right https://github.com/OCA/OCB/blob/2fcfc6cfa72f79700c00ea2c71543bd4706f2464/addons/point_of_sale/static/src/js/ClassRegistry.js#L12

(unless a more js savvy commenter says otherwise :))

legalsylvain:
>     const GlobalLineDiscount = () =>
>        class extends DiscountButton {

I remember that I tried that syntax, and that : 
- it was working for a module.
- but I developped two modules that extends the same object, and then, it was generating an error if both module was installed. For that reason, I reverted and set a ignore no-shallow in my PR.

However, I don't undestand why.

> And being that the standard enforces this way of inheritance up to v16, maybe the warning removal is right https://github.com/OCA/OCB/blob/2fcfc6cfa72f79700c00ea2c71543bd4706f2464/addons/point_of_sale/static/src/js/ClassRegistry.js#L12

what a clean change !

ivantodorovich:
> I remember that I tried that syntax, and that :
>
>it was working for a module.
>but I developped two modules that extends the same object, and then, it was generating an error if both module was installed. For that reason, I reverted and set a ignore no-shallow in my PR.

I don't understand why it fails, but I also have previous wars memories about this, and having to disable this `no-shadow` rule. IMO this linting rule goes against a pattern pushed by odoo core, and I'm happy to disable it.

Althuogh, disabling it here won't stick, right? Won't we loose it after a copier update?


-----

UPDATE: actually it's clear why it fails: the function is not evaluated right away. It's passed on to the registry, who's in charge of progresively evaluating it in chain for all the registered extensions

.eslintrc.yml
```diff
@@ -126,7 +126,7 @@ rules:
   no-self-compare: warn
   no-sequences: warn
   no-shadow-restricted-names: warn
-  no-shadow: warn
+  # no-shadow: warn
```
ivantodorovich, line 129: My only concern is this won't survive a copier update

ivantodorovich (approved):

legalsylvain:
Thanks all for your feedbacks ! 

> Won't we loose it after a copier update?

Indeed. Not a big deal, imo. When we copier update, we have to add this patch again. it doesn't happen every day.

> UPDATE: actually it's clear why it fails: the function is not evaluated right away. It's passed on to the registry, who's in charge of progresively evaluating it in chain for all the registered extensions

Do you think we have to remove this rule at the OCA template level ? (impacting web, website, etc...)

.eslintrc.yml
chienandalu: Then maybe the way to go would be to have an special eslint in copier for version 14-15 (something like) https://github.com/OCA/oca-addons-repo-template/blob/master/src/%7B%25%20if%20odoo_version%20%3E%2012%20%25%7D.eslintrc.yml%7B%25%20endif%20%25%7D

chienandalu:
> Do you think we have to remove this rule at the OCA template level ? (impacting web, website, etc...)

We can restrict it to pos, I think: https://github.com/OCA/oca-addons-repo-template/blob/master/copier.yml#L48

ivantodorovich:
> Do you think we have to remove this rule at the OCA template level ? (impacting web, website, etc...)

Yes, this pattern is also used for backend widgets.. so IMO it makes sense to remove it on all repos

chienandalu:
I wonder if we should simply use the standard patching pattern that is pointed here: https://github.com/odoo/odoo/commit/b119cca37b2f345e17ef62de9aacb638e1c075ab

I don't think nothing stops us from doing it at least in v16

ivantodorovich:
> I wonder if we should simply use the standard patching pattern that is pointed here: [odoo/odoo@b119cca](https://github.com/odoo/odoo/commit/b119cca37b2f345e17ef62de9aacb638e1c075ab)
> 
> I don't think nothing stops us from doing it at least in v16

I guess that's coming for 17.0, but until then I fail to see the point in enforcing an alternative pattern solely because of this linting rule. It'd only make things more confusing for people not aware of this, IMO.

legalsylvain:
@chienandalu 

Are you ok to merge this PR as is. (it removes a lot of annoying warning). 

Then we can improve some point, regarding CI, etc...

chienandalu (approved):
Yes, sure :+1:

PR #1025, opened 2023-07-10
[16.0][FIX] pos_order_to_sale_order

geomer198:
Added return for CreateOrderButton component. It is used for inheriting component.


legalsylvain (approved):
Thanks !

/ocabot merge patch

PR #1026, opened 2023-07-17
[FIX][16.0] pos_order_remove_line: clean toRefundLines when deleting a refunded orderline from remove-line-button

suker:
In a refund order, if we delete a refunded orderline from remove-line-button, it stills says that orderline is being refunding.

Example:
1. Refund an orderline line.
image

3. Remove it from orderline button
image

5. Check that refunded orderline is still in "refunding".
image

4. If we click in the order, line is already deleted.


legalsylvain (approved):
makes sense.

robyf70 (approved):
LGTM

danielduqma (approved):

PR #1028, opened 2023-07-31
[IMP] add a menu entry in account module to manage pos place for user that use place only on invoices

legalsylvain:


quentinDupont (approved):
Code review and works well in prod

legalsylvain (commented):
/ocabot merge nobump

PR #1029, opened 2023-08-02
[14.0] [FIX] pos_reset_search: wrong class name ProductsWidgetControlPanel

victor-champonnois:
@rousseldenis 
It seems the module pos_reset_search doesn't work currently (I tested on the runboat). I cannot say I understand all the logic in the JS code, but I suspect that the wrong class was called. It seems to work this way.


DemchukM (approved):
LGTM. I think that this class should be there

PR #1033, opened 2023-08-03
[ADD] new module 'pos_bank_statement_line_no_change'. prevent to unlink / write on bank statement lines of confirmed PoS Orders

legalsylvain:
Trivial module to prevent to update / delete bank statement lines that are related to a confirmed pos order.


quentinDupont (approved):
Code review :heavy_check_mark:

PR #1035, opened 2023-08-05
[IMP] pos_order_to_sale_order: Improvement for the ability to use the result of sales order creation after inheritance

GabbasovDinar:
These changes allow to use the data after sales order creation after inheritance.

Before this changes:
 - I cannot use sale_order_id

After this changes:
 - I can extend the _createSaleOrder method and use the result of creating a sales order


legalsylvain:
Hi. Could you take a take a look on precommit ?

GabbasovDinar:
@legalsylvain Done

legalsylvain (approved):
Thanks. 

/ocabot merge patch

PR #1042, opened 2023-08-14
[16.0][REF] pos_order_to_sale_order: create method optimization

GabbasovDinar:
Optimized the process of sales order creation. 

**Before this changes:**

Executing the create command creates an empty order, which can lead to unexpected effects if the create method has been extended by another module

**After this changes:**

The create method will return the order with the product lines included.

### DIFF
--- a/pos_order_to_sale_order/models/sale_order.py
+++ b/pos_order_to_sale_order/models/sale_order.py
@@ -2,7 +2,7 @@
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, models
+from odoo import Command, _, api, models
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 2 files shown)


pos_order_to_sale_order/models/sale_order.py
```diff
@@ -12,31 +12,27 @@ class SaleOrder(models.Model):
...
+            for line in order_data["lines"]
+        ]
         return {
             "partner_id": order_data["partner_id"],
             "origin": _("Point of Sale %s") % (session.name),
             "client_order_ref": order_data["name"],
             "user_id": order_data["user_id"],
             "pricelist_id": order_data["pricelist_id"],
             "fiscal_position_id": order_data["fiscal_position_id"],
+            "order_lines": order_lines,
```
legalsylvain:
```suggestion
            "order_line": order_lines,
```

https://github.com/odoo/odoo/blob/16.0/addons/sale/models/sale_order.py#L208

Should fix the CI

GabbasovDinar: @legalsylvain done

ivs-cetmix (approved):
LGTM

Aldeigja (approved):
Functional, ok.

PR #1045, opened 2023-08-18
[ADD] pos_order_zip

SMaciasOSI:


SMaciasOSI:
@max3903 @dreispt @AmmarOfficewalaSerpentCS

rafaelbn:
@fcvalgar please review 😄 

MT-4562

@moduon

pos_order_zip/models/order_zip_code.py
```diff
@@ -0,0 +1,72 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class OrderZipcode(models.Model):
+    _name = "order.zipcode"
+    _rec_name = "zip_code"
+
+    zip_code = fields.Char()
+    state_id = fields.Many2one("res.country.state", string="State", ondelete="restrict")
+    country_id = fields.Many2one("res.country", string="Country", ondelete="restrict")
```
Shide, line 15: Could be a related field to `state_id.country_id`

pos_order_zip/models/order_zip_code.py
```diff
@@ -0,0 +1,72 @@
...
+                zip_code.country_id = (
+                    zip_code.state_id.country_id and zip_code.state_id.country_id.id
+                )
+
+    # compute the number of pos orders with a zip code
+    def _compute_order_count(self):
+        for zip_code in self:
+            order_count = self.env["pos.order"].search_count(
+                [
+                    ("zip_code", "=like", zip_code.zip_code + "%"),
```
Shide, line 33: Why not "="?

```suggestion
                    ("zip_code", "=", zip_code.zip_code),
```

pos_order_zip/models/order_zip_code.py
```diff
@@ -0,0 +1,72 @@
...
+                    ("state", "!=", "cancel"),
+                ]
+            )
+            zip_code.order_count = order_count
+
+    # Check if the entered zipcode is valid
+    @api.model
+    def search_order_zipcode(self, zipcode):
+        valid_zip_code = self.search([("zip_code", "=", zipcode[:5])])
```
Shide, line 42: Why only first 5 positions?

pos_order_zip/models/order_zip_code.py
```diff
@@ -0,0 +1,72 @@
...
+        if valid_zip_code:
+            return True
+        return False
+
+    @api.constrains("zip_code")
+    def check_zip_code(self):
+        for zip_code in self:
+            if self.search(
+                [("zip_code", "=", zip_code.zip_code), ("id", "!=", zip_code.id)]
```
Shide, line 51: Here you are comparing with "=" instead "ilike" and "[:5]"

pos_order_zip/models/order_zip_code.py
```diff
@@ -0,0 +1,72 @@
...
+            ):
+                raise ValidationError(
+                    _("Another record with the same zip code already exists!")
+                )
+
+    def test_time(self):
```
Shide, line 57: What's the point of this function?

Shide (changes requested):
Whouldn't be easy to check if the selected customer has the `state_id` field set or add a new field `state_id` on the `pos.order` model?

I didn't see the point to create a new model to add a zip code.

fcvalgar:
@SMaciasOSI can you rebase , please?

pos_order_zip/models/order_zip_code.py
```diff
@@ -0,0 +1,72 @@
...
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+
+class OrderZipcode(models.Model):
+    _name = "order.zipcode"
+    _rec_name = "zip_code"
+
+    zip_code = fields.Char()
```
chienandalu, line 13: Why not simply build your module on top of the existing `base_location`: https://github.com/OCA/partner-contact/blob/16.0/base_location/models/res_city_zip.py

That way you ensure cross compatibility without pulling new data structures

fcvalgar (commented):
Functional review,

It is a fantastic functionality but I have some doubts for many use cases. 

It is very blocking that the validation is not optional because in some cases we will not have the data. Customers who want privacy for example. 
Entering a data that is not real will alter the sales reports. 

Can it be modified so that the field is optional?

rafaelbn:
Hello @SMaciasOSI @dreispt @max3903, could you please check and answer comments from other contributors please? 🙏🏼 😄 

- [ ] Make it optional or at least the option to make it optional. As said, not all customers will fill the zip and shouldn't be mandatory

> It is very blocking that the validation is not optional because in some cases we will not have the data. Customers who want privacy for example.

- [ ] Validate the ZIP, to be sure it's not fake

>  Entering a data that is not real will alter the sales reports.

- [ ] Depends of base_location: https://github.com/OCA/pos/pull/1045#pullrequestreview-1840781823

- [ ] Make possible to group by in reports https://github.com/odoo/odoo/pull/121594


Let us know please.

If you agree, maybe we could improve it. But I would like to know if you are already using un production.

Thank you! ❤️

PR #1049, opened 2023-09-06
[14.0][ADD] pos_takeout

felipezago:
This module implements an option on the POS order to set if the customer is eating here, or taking out.


pos_takeout/static/src/js/TakeOutButton.js
```diff
@@ -0,0 +1,41 @@
...
+            this.currentOrder.eat_here = !this.currentOrder.eat_here;
+            this.render();
+        }
+    }
+    TakeOutButton.template = "TakeOutButton";
+
+    ProductScreen.addControlButton({
+        component: TakeOutButton,
+        condition: function () {
+            return true;
```
ygcarvalh, line 34: wouldn't there need to be a check for the boolean you put in `pos.config`?

felipezago: No, i've made the check on the XML

ivantodorovich (commented):
Shouldn't this module depend on `pos_restaurant` ?
It kinda feels restaurant-related

pos_takeout/models/pos_order.py
```diff
@@ -0,0 +1,24 @@
...
+# @author: Felipe Zago Rodrigues <felipe.zago@kmee.com.br>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class PosOrder(models.Model):
+    _inherit = "pos.order"
+
+    eat_here = fields.Boolean(string="Is the customer eating here", default=True)
```
ivantodorovich, line 11: To be consistent with the module name and the rest of the terms, I'd invert this field like this:

```suggestion
    takeout = fields.Boolean(help="Is this a takeout order?")
```

pos_takeout/static/src/js/TakeOutButton.js
```diff
@@ -0,0 +1,41 @@
...
+        setup() {
+            useListener("click", this.onClick);
+        }
+
+        get currentOrder() {
+            return this.env.pos.get_order();
+        }
+
+        onClick() {
+            this.currentOrder.eat_here = !this.currentOrder.eat_here;
```
ivantodorovich, line 25: Same comment as above, I'd stick to the "takeout" term

```suggestion
            this.currentOrder.takeout = !this.currentOrder.takeout;
```

pos_takeout/static/src/js/TakeOutButton.js
ivantodorovich: This is exactly what `condition` is for. Better to put the condition here, so that the QWeb isn't even rendered in the first place

pos_takeout/static/src/xml/TakeOutButton.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template" xml:space="preserve">
+
+    <t t-name="TakeOutButton" owl="1">
+        <span
+            class="control-button js_take_out"
```
ivantodorovich, line 6: This class doesn't really seem useful, does it?

```suggestion
            class="control-button"
```

pos_takeout/static/src/xml/TakeOutButton.xml
```diff
@@ -0,0 +1,20 @@
...
+            t-if="env.pos.config.iface_show_takeout_option"
```
ivantodorovich, line 7: Put the condition in `addControlButton` instead

PR #1050, opened 2023-09-08
[14.0][NEW] pos_plate_number

felipezago:
This module implements the possibility of setting a plate number within the Point of Sale (POS).

The plate number may be set either manually or automatically through the order number.


legalsylvain:
Hi. thanks for your contribution. could you explicit what is "plate". The translation is not simple in french, I'm not sure to understand the use case.

pos_plate_number/models/pos_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    use_plate_number = fields.Boolean(string="Use Plate Number")
+
+    plate_number_generation = fields.Selection(
```
ygcarvalh, line 13: Is there a specific reason for having what type of plate number generation was or is it just to save the info?

ygcarvalh:
As far as I can understand, "plate number" would be those numbers that appear on the receipt so that you can pick up your order at a self-service terminal, for example.

@legalsylvain

pos_plate_number/models/pos_config.py
felipezago: Its just an option, to ask for the plate number or generate id based on the order uid.

felipezago:
> As far as I can understand, "plate number" would be those numbers that appear on the receipt so that you can pick up your order at a self-service terminal, for example.
> 
> @legalsylvain

Exactly, thanks @ygcarvalh!

ivantodorovich:
> As far as I can understand, "plate number" would be those numbers that appear on the receipt so that you can pick up your order at a self-service terminal, for example.

What's the difference with the "order number", then ? 🤔

PR #1052, opened 2023-09-12
[14.0][FIX] pos_customer_display: set use_proxy=true

alexis-via:
I found this bug on a setup where the only device using POSbox/pywebdriver was a Bixolon customer display (payment terminal and receipt printer didn't use the POSbox/pywebdriver).


ivantodorovich (approved):
Makes sense

PR #1058, opened 2023-09-28
[16.0] [IMP] pos_loyalty_redeem_payment Add parameters to disable auto-print giftcard

pfranck:
For some usecase we would like to disable the download of PDF gift cards. We added a config parameter allow_auto_print_giftcard to disable it

### DIFF
--- a/pos_loyalty_redeem_payment/static/src/js/PaymentScreen.esm.js
+++ b/pos_loyalty_redeem_payment/static/src/js/PaymentScreen.esm.js
@@ -156,7 +156,7 @@ export const CouponPosPaymentScreen = (OriginalPaymentScreen) =>
         }
 
         async _postPushOrderResolve(order, order_server_ids) {
-            if (order.has_redeem_payment_lines()) {
+            if ( this.pos.config.allow_auto_print_giftcard && order.has_redeem_payment_lines()) {
                 const payload = await this.rpc({
                     model: "pos.order",
                     method: "get_loy_card_reports_from_order",
--- a/pos_loyalty_redeem_payment/static/src/js/models.esm.js
+++ b/pos_loyalty_redeem_payment/static/src/js/models.esm.js
@@ -13,7 +13,7 @@ export const RedeemPaymentOrder = (OriginalOrder) =>
 
         wait_for_push_order() {
             let result = super.wait_for_push_order(...arguments);
-            result = Boolean(result || this.has_redeem_payment_lines());
+            result = Boolean(result ||  this.pos.config.allow_auto_print_giftcard && this.has_redeem_payment_lines());
             return result;
         }
     };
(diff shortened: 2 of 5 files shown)


pos_loyalty_redeem_payment/static/src/js/models.esm.js
```diff
@@ -6,7 +6,7 @@ import Registries from "point_of_sale.Registries";
 export const RedeemPaymentOrder = (OriginalOrder) =>
     class extends OriginalOrder {
         has_redeem_payment_lines() {
-            return this.paymentlines.some(
+            return this.pos.config.allow_auto_print_giftcard && this.paymentlines.some(
```
danielduqma: Practical config, but IMHO it would be better to leave `has_redeem_payment_lines` with its original behaviour (`true` if any payment line is linked to a voucher) and check `allow_auto_print_giftcard` in `_postPushOrderResolve` and `wait_for_push_order`

danielduqma (changes requested):

pos_loyalty_redeem_payment/static/src/js/models.esm.js
pfranck: yes make sense, I'll make the change

danielduqma (approved):

pos_loyalty_redeem_payment/static/src/js/PaymentScreen.esm.js
```diff
@@ -156,7 +156,7 @@ export const CouponPosPaymentScreen = (OriginalPaymentScreen) =>
         }
 
         async _postPushOrderResolve(order, order_server_ids) {
-            if (order.has_redeem_payment_lines()) {
+            if ( this.pos.config.allow_auto_print_giftcard && order.has_redeem_payment_lines()) {
```
danielduqma, line 159:
```suggestion
            if (this.env..pos.config.allow_auto_print_giftcard && order.has_redeem_payment_lines()) {
```

pos_loyalty_redeem_payment/static/src/js/models.esm.js
```diff
@@ -13,7 +13,7 @@ export const RedeemPaymentOrder = (OriginalOrder) =>
 
         wait_for_push_order() {
             let result = super.wait_for_push_order(...arguments);
-            result = Boolean(result || this.has_redeem_payment_lines());
+            result = Boolean(result ||  this.pos.config.allow_auto_print_giftcard && this.has_redeem_payment_lines());
```
danielduqma, line 16:
```suggestion
            result = Boolean(result ||  this.env..pos.config.allow_auto_print_giftcard && this.has_redeem_payment_lines());
```

danielduqma (changes requested):

PR #1064, opened 2023-10-13
[14.0][FIX] pos_pricelist_show_discount: remove fragile demo data

HekkiMelody:
Removing pricelist and pos configuration from the demo data because it creates issues when combined with l10n modules that modify the company's currency, due to inconsistencies in currencies.

The same demo data is now created programmatically during tests instead.


francesco-ooops:
@GSLabIt what do you think?

GSLabIt (approved):

PR #1065, opened 2023-10-15
[16.0][FIX] pos_order_to_sale_order

geomer198:
At creating SO from POS, description computed by default with adding customer note from POS.

### DIFF
--- a/pos_order_to_sale_order/models/res_config_settings.py
+++ b/pos_order_to_sale_order/models/res_config_settings.py
@@ -23,3 +23,13 @@ class ResConfigSettings(models.TransientModel):
     pos_iface_create_invoiced_sale_order = fields.Boolean(
         related="pos_config_id.iface_create_invoiced_sale_order", readonly=False
     )
+
+    sol_name_mode = fields.Selection(
+        selection=[
+            ("product_pos", "Product name + POS Comment"),
+            ("multiline", "Sale Multiline Description"),
+        ],
+        string="SO Line Name Mode",
+        default="product_pos",
+        config_parameter="pos_order_to_sale_order.sol_name_mode",
+    )
--- a/pos_order_to_sale_order/models/res_config_settings.py
+++ b/pos_order_to_sale_order/models/res_config_settings.py
@@ -2,7 +2,7 @@
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResConfigSettings(models.TransientModel):
@@ -33,3 +33,11 @@ class ResConfigSettings(models.TransientModel):
         default="product_pos",
         config_parameter="pos_order_to_sale_order.sol_name_mode",
     )
+
+    @api.model
+    def set_values(self):
+        result = super(ResConfigSettings, self).set_values()
+        ICPSudo = self.env["ir.config_parameter"].sudo()
+        value = self.sol_name_mode or "product_pos"
+        ICPSudo.set_param("pos_order_to_sale_order.sol_name_mode", value)
+        return result
--- a/pos_order_to_sale_order/models/sale_order_line.py
+++ b/pos_order_to_sale_order/models/sale_order_line.py
@@ -10,16 +10,22 @@ class SaleOrderLine(models.Model):
 
     @api.model
     def _prepare_from_pos(self, order_line_data):
-        ProductProduct = self.env["product.product"]
-        product = ProductProduct.browse(order_line_data["product_id"])
-        product_name = product.name
-        if order_line_data.get("customer_note"):
-            product_name += "\n" + order_line_data["customer_note"]
-        return {
+        vals = {
             "product_id": order_line_data["product_id"],
-            "name": product_name,
             "product_uom_qty": order_line_data["qty"],
             "discount": order_line_data["discount"],
             "price_unit": order_line_data["price_unit"],
             "tax_id": order_line_data["tax_ids"],
         }
+        if (
+            order_line_data.get("customer_note")
+            and self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("pos_order_to_sale_order.sol_name_mode", "product_pos")
+            == "product_pos"
+        ):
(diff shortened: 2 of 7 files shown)


legalsylvain (changes requested):
Hi. Don't understand that PR. 
I tested on runboat with the 16.0 module : 
- add a pos order line
- add a comment
- create a draft sale order

the comment is present in the name of the sale order line.

image

geomer198:
> Hi. Don't understand that PR. I tested on runboat with the 16.0 module :
> 
>     * add a pos order line
> 
>     * add a comment
> 
>     * create a draft sale order
> 
> 
> the comment is present in the name of the sale order line.
> 
> image


In default module behavior product internal reference in sol is not displayed. Now products Internal reference is displayed.

legalsylvain:
Thanks for the explanation ! 

Well, the problem IMO is that this PR introduce a useless field at sale.order.line level that will take space in the DB. We should consider another design.

I guess call ``_get_sale_order_line_multiline_description_sale`` in the prepare function should do the job. Don't you think ?

geomer198:
> `_get_sale_order_line_multiline_description_sale`

Hello @legalsylvain ! This function (_get_sale_order_line_multiline_description_sale) is called after created SOL, so at preparing value I think better solution adding new field for saving customer_note.
For the applying `_get_sale_order_line_multiline_description_sale` function then need create so and sol without preparing value. I think this solution is not good( 
What do you think about it?

legalsylvain:
I think you can pass information by context. It could make the job. 
but creating a useless field in database is an issue IMO.

ivs-cetmix:
> I think you can pass information by context. It could make the job. but creating a useless field in database is an issue IMO.

Hi @legalsylvain we have approached this problem from another angle) Could you have a look please?

legalsylvain:
Hi. I'm on holidays the next week. I'll take a look when I'm back to the office.

ivs-cetmix:
> Hi. I'm on holidays the next week. I'll take a look when I'm back to the office.

Thank you! Have a nice holidays)

pos_order_to_sale_order/readme/CONFIGURE.rst
```diff
@@ -1,6 +1,7 @@
 * Go to Point Of Sale / Configuration / Point of Sale
 * Check the box 'Create Sale Orders'
 * Select the desired default behaviour
+* Select the SO Line Name mode
```
ivs-cetmix: Select Sales Order Lina name composition mode

pos_order_to_sale_order/readme/newsfragments/2928.bigfix.rst
```diff
@@ -0,0 +1 @@
+Sale order line description computed by default with adding Customer note.
```
ivs-cetmix: This is not a bug. A feature is being added.
Add readable and clear feature description please

ivs-cetmix (changes requested):

pos_order_to_sale_order/models/res_config_settings.py
```diff
@@ -23,3 +23,13 @@ class ResConfigSettings(models.TransientModel):
...
         related="pos_config_id.iface_create_invoiced_sale_order", readonly=False
     )
+
+    sol_name_mode = fields.Selection(
+        selection=[
+            ("product_pos", "Product name + POS Comment"),
+            ("multiline", "Sale Multiline Description"),
+        ],
+        string="SO Line Name Mode",
+        default="product_pos",
```
ivs-cetmix, line 33: What happens if user will remove selection value? So no option is selected

ivs-cetmix (changes requested):

ivs-cetmix:
> Hi. I'm on holidays the next week. I'll take a look when I'm back to the office.

Hi @legalsylvain sorry for bothering you again) Would be great if you could have a look at this

pos_order_to_sale_order/models/sale_order_line.py
```diff
@@ -10,16 +10,22 @@ class SaleOrderLine(models.Model):
...
+            and self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("pos_order_to_sale_order.sol_name_mode", "product_pos")
+            == "product_pos"
+        ):
+            product = self.env["product.product"].browse(order_line_data["product_id"])
+            product_name = product.name
+            product_name += "\n" + order_line_data["customer_note"]
+            vals.update(name=product_name)
+        return vals
```
legalsylvain, line 31: Maybe I'm wrong, but it looks like if the previous condition is wrong, the vals returned doesn't contain any "name" value. Did I missed something ?

geomer198: > Maybe I'm wrong, but it looks like if the previous condition is wrong, the vals returned doesn't contain any "name" value. Did I missed something ?

Thank you for you question! 
Name field is auto complete after creating SOL. And this solution for the set custom value for SOL description.

legalsylvain:
Hi @geomer198.

> Hello @legalsylvain ! This function (_get_sale_order_line_multiline_description_sale) is called after created SOL, so at preparing value I think better solution adding new field for saving customer_note.

I don't think so.  ``_get_sale_order_line_multiline_description_sale`` is called by the ``_compute_name`` function, before the call during the call of create.

I made an alternative implementation here #1080.

Thanks for you review.

geomer198:
> Hi @geomer198.
> 
> > Hello @legalsylvain ! This function (_get_sale_order_line_multiline_description_sale) is called after created SOL, so at preparing value I think better solution adding new field for saving customer_note.
> 
> I don't think so. `_get_sale_order_line_multiline_description_sale` is called by the `_compute_name` function, before the call during the call of create.
> 
> I made an alternative implementation here #1080.
> 
> Thanks for you review.

Thank you for your solution!

geomer198:
> Hi @geomer198.
> 
> > Hello @legalsylvain ! This function (_get_sale_order_line_multiline_description_sale) is called after created SOL, so at preparing value I think better solution adding new field for saving customer_note.
> 
> I don't think so. `_get_sale_order_line_multiline_description_sale` is called by the `_compute_name` function, before the call during the call of create.
> 
> I made an alternative implementation here #1080.
> 
> Thanks for you review.

Thanks for your solution.
This solution looks better than my.  I close my PR in favor of your implementation.

PR #1067, opened 2023-10-26
[14.0][FIX] pos_order_mgmt: Return with positive amounts

zamberjo:
When making a return, the POS user can change the return quantity into positive. If the user does not realise this, the error message ["Please specify at least one non-zero quantity"](https://github.com/odoo/odoo/blob/14.0/addons/stock/wizard/stock_picking_return.py#L167) is displayed at the end of the order. leaving the order to be synchronised, always giving the same error.

I propose that when making the return from the POS, this line should always be forced to a negative amount. 

In addition, a ConfirmPopup has been added to make sure that the user is aware that he is making a return of more than the amount sold.

### DIFF
--- a/pos_order_mgmt/static/src/js/ProductScreen.js
+++ b/pos_order_mgmt/static/src/js/ProductScreen.js
@@ -1,0 +1,42 @@
+/* Copyright Aures Tic - Jose Zambudio <jose@aurestic.es>
+   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl) */
+odoo.define("pos_order_mgmt.ProductScreen", function (require) {
+    "use strict";
+
+    const ProductScreen = require("point_of_sale.ProductScreen");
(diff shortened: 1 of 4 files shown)


belenaures:
I've checked it and I've noticed that when trying to delete a line an error appears: 
-> what i do is to is press the delete button in the POS, first it remains as qty 0 and then it should delete the line, instead, the error shows.
Rastreo de error: Error: 'remove' no es un decimal correcto at Object.parseFloat [as float

zamberjo:
> I've checked it and I've noticed that when trying to delete a line an error appears: -> what i do is to is press the delete button in the POS, first it remains as qty 0 and then it should delete the line, instead, the error shows. Rastreo de error: Error: 'remove' no es un decimal correcto at Object.parseFloat [as float

Fixed

belenaures (approved):
Functionality checked, all ok

chienandalu (approved):
Looks good :+1:

PR #1071, opened 2023-11-02
[16.0][FIX] pos_partner_birthdate: fixing this.changes state

imbarbudiman:
use `useState` when add `birthdate_date` to `changes` state

PR #1072, opened 2023-11-02
[16.0][FIX] pos_partner_birthdate: Fixing conflict t-model vs t-att-value

imbarbudiman:
After merged this PR #1071 now `changes.birthdate_date` is reactive.
If we use `t-model`, there is no need to use the `t-att-value`, because this will cause abnormal behavior.

Before this changes: 
Before del t-att-value

After this changes: 
After del t-att-value


chienandalu (approved):
Thanks again @imbarbudiman :)

chienandalu:
Hi @ecino would you add your review?

ecino (approved):
Great! :+1:

PR #1074, opened 2023-11-03
[16.0] pos_partner_firstname: fix string

alexis-via:
The kind of bad string you only see when working on PO files !

### DIFF
--- a/pos_partner_firstname/static/src/js/PartnerDetailsEdit.js
+++ b/pos_partner_firstname/static/src/js/PartnerDetailsEdit.js
@@ -45,7 +45,7 @@ odoo.define("pos_partner_firstname.PartnerDetailsEdit", function (require) {
                         processedChanges.lastname === ""
                     ) {
                         return this.showPopup("ErrorPopup", {
-                            title: _t("Both Customer First and Last Name Are Required"),
+                            title: _t("Both customer first name and last name are required."),
                         });
                     }
                     if (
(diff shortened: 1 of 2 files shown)


pos_partner_firstname/static/src/js/PartnerDetailsEdit.js
```diff
@@ -45,7 +45,7 @@ odoo.define("pos_partner_firstname.PartnerDetailsEdit", function (require) {
                         processedChanges.lastname === ""
                     ) {
                         return this.showPopup("ErrorPopup", {
-                            title: _t("Both Customer First and Last Name Are Required"),
+                            title: _t("Both customer first name and last name are required."),
```
legalsylvain, line 48: are you sure on this one ? 

https://www.deepl.com/translator#fr/en/Le%20nom%20et%20le%20pr%C3%A9nom%20du%20client%20sont%20obligatoires

legalsylvain (commented):
One remark inline. Otherwise, LGTM.

DorianMAG:
Hi @alexis-via 
This issue https://github.com/OCA/pos/issues/1097 have relation with your PR?

pos_partner_firstname/static/src/js/PartnerDetailsEdit.js
cvinh: This is involved in v17 here https://github.com/OCA/pos/pull/1126

PR #1091, opened 2023-11-23
[ADD] pos_payment_method_change_policy : new module to add different ways to handle Change in Point of Sale

legalsylvain:
This module improves the 'Change' feature in Odoo Point of Sale,
when a customer pays more that the amount of the PoS Order:

cash_1_payment_screen

By default, a "Change" line is added on the receipt:

cash_2_bill

from an accounting point of view, it adds a second payment line in 'Cash' method
with a negative amount corresponding to the surplus:

cash_3_back_office

In some countries and for some payment methods, this is not the right way to proceed.

This module adds extra "Change Policy" option to manage differently the giving of change.

Configuration
=============

**Exception Profit Product**

* Create a dedicated 'No change' Product.
  The type of the product should be 'Service' and the product should available in PoS.
  Generally, no VAT is applied on such product.

* Go to the payment method form.

* set 'Change Option' as 'Exceptional Profit Product'

* set the new product in the 'Change Product field'

payment_screen_profit_policy

Usage
=====

**Exception Profit Product**

A dedicated warning will be displayed:

profit_policy_1_payment_screen

By default, a new sale order line is added on the receipt:

profit_policy_2_bill

So, from an accounting point of view, it adds a new line that will
impact a profit and loss account:

profit_policy_3_back_office


ivantodorovich (approved):
LGTM

Issue #1092, opened 2023-11-23
pos_report_session_summary Error when printing the session summary.


anmarmo1:
## Module
[pos_report_session_summary](https://github.com/OCA/pos/tree/15.0/pos_report_session_summary)

## Describe the bug
When printing the session summary for both an open and closed session, it produces a format error.

**Affected versions**:
15.0
## To Reproduce

1.Go to session
2. Click on print session summary
3. After clicking, this error occurs

Traceback (most recent call last):
  File "/opt/odoo/addons/web/controllers/main.py", line 2000, in report_download
    response = self.report_routes(reportname, docids=docids, converter=converter, context=context)
  File "/opt/odoo/odoo/http.py", line 546, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/addons/web/controllers/main.py", line 1929, in report_routes
    pdf = report.with_context(context)._render_qweb_pdf(docids, data=data)[0]
  File "/opt/odoo/addons/account/models/ir_actions_report.py", line 55, in _render_qweb_pdf
    return super()._render_qweb_pdf(res_ids=res_ids, data=data)
  File "/opt/odoo/odoo/addons/base/models/ir_actions_report.py", line 902, in _render_qweb_pdf
    html = self_sudo.with_context(context)._render_qweb_html(res_ids, data=data)[0]
  File "/opt/odoo/odoo/addons/base/models/ir_actions_report.py", line 946, in _render_qweb_html
    return self._render_template(self.report_name, data), 'html'
  File "/opt/odoo/odoo/addons/base/models/ir_actions_report.py", line 681, in _render_template
    return view_obj._render_template(template, values).encode()
  File "/opt/odoo/odoo/addons/base/models/ir_ui_view.py", line 1992, in _render_template
    return self.browse(self.get_view_id(template))._render(values, engine)
  File "/opt/odoo/addons/web_editor/models/ir_ui_view.py", line 29, in _render
    return super(IrUiView, self)._render(values=values, engine=engine, minimal_qcontext=minimal_qcontext)
  File "/opt/odoo/odoo/addons/base/models/ir_ui_view.py", line 2000, in _render
    return self.env[engine]._render(self.id, qcontext)
  File "/opt/odoo/odoo/tools/profiler.py", line 292, in _tracked_method_render
    return method_render(self, template, values, **options)
  File "/opt/odoo/odoo/addons/base/models/ir_qweb.py", line 76, in _render
    result = super()._render(template, values=values, **compile_options)
[...]

PR #1075, opened 2023-11-05
pos_report_session_summary: fix printing session report

sebastienbeau:
Remove useless quote

PR #1093, opened 2023-11-24
[15.0][FIX] pos_report_session_summary: fix error when printing

almumu:
Is not possible to print the report. This pr fixes the error: 'str' object has no attribute 'decimal_places' when printing session summary

### DIFF
--- a/pos_report_session_summary/views/report_session_summary.xml
+++ b/pos_report_session_summary/views/report_session_summary.xml
@@ -65,28 +65,16 @@
                                 <td><span t-field="statement.name" /></td>
                                 <td><span t-field="statement.journal_id" /></td>
                                 <td class="text-right">
-                                    <span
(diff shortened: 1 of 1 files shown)


almumu:
this solves https://github.com/OCA/pos/issues/1092

anmarmo1 (approved):
Functional review is ok.

zamberjo (approved):

chienandalu (approved):
As these fields are monetary we don't need indeed these `t-options`. The problem thou was that the `display_currency` value was a string instead of the value itself, as it was addressed in https://github.com/OCA/pos/pull/1075

/ocabot merge patch

ivantodorovich:
@almumu could you rebase your branch, please?

It seems there was an error during the last merge attempt

ivantodorovich:
/ocabot merge patch

Let's try again

PR #1096, opened 2023-11-26
[FIX] pos_payment_change : check installation fo existing modules.

### LINKED ISSUE #1004: pos_payment_change for France
I put this module for a French Retail (B2C) client. Very useful, it 's definitely an easier way to change payment in pos.

Just a question about the constrain which is present to check the value of Payment Change Policy. "If the module l10n_fr_certification is installed and if the current company has an inalterable accounting, it will not be possible to select the value 'Update Payments'". 

My client has "l10n_fr_pos_cert" but not  "l10n_fr_certification". I think it could be appropriate that the constrain checks whether one of these 2 modules is installed to prevent the "Update Payments" option to be chosen.

legalsylvain:
In old version, two certifications modules was present in odoo.  l10n_fr_pos_cert and l10n_fr_certification.  In odoo 16, only l10n_fr_pos_cert exists.

Fix : #1004 

@LESTRAT21, could you take a look on this one ?

### DIFF
--- a/pos_payment_change/models/pos_config.py
+++ b/pos_payment_change/models/pos_config.py
@@ -36,7 +36,7 @@ class PosConfig(models.Model):
         module_states = (
             self.env["ir.module.module"]
             .sudo()
-            .search([("name", "=", "l10n_fr_certification")])
+            .search([("name", "=", "l10n_fr_pos_cert")])
             .mapped("state")
         )
         if "installed" not in module_states:


LESTRAT21:
Thanks for your fix @legalsylvain I will re-use your PR

LESTRAT21 (approved):

legalsylvain:
thanks for your review ! 

/ocabot merge patch

PR #1098, opened 2023-11-28
[16.0][IMP] use right method to access action / remove an obsolete Warning

nayatec:


pos_payment_change/tests/test_module.py
```diff
@@ -23,6 +23,22 @@ def setUp(self):
         ]
         self.product = self.env.ref("product.product_product_3")
         self.pos_config = self.env.ref("point_of_sale.pos_config_main").copy()
+        self.pos_user = self.env["res.users"].create(
```
legalsylvain: we could use default current "admin" user, but just remove the "Administrator / Settings" group in that test. So it is more concise, don't you think ? 

Or directly use "base.user_demo". 

self.env = self.env.sudo(self.env.ref("base.user_demo")

legalsylvain (commented):
Hi olivier. 

A remark inline. Otherwise, LGTM. 
Thanks a lot !

pos_payment_change/tests/test_module.py
nayatec: will do !

legalsylvain (approved):
Thanks for this fix ! 

/ocabot merge patch

PR #1104, opened 2023-12-04
[17.0][MIG] pos_partner_birthdate

etobella:
Migration to 17.0

@cvinh


chienandalu (approved):

cvinh (approved):

cvinh:
@etobella do you know why the birthdate field in the pos is not sensitive to language ? I tested it with a user having english language and the date format is jj/mm/aaaa
it should be mm/dd/yyyy
In the backend, I have mm/dd/yyyy
does it depends on my browser language  (which is french) ?

cvinh (approved):

etobella:
@cvinh I followed the same logic we find in other date fields inside PoS.  

https://github.com/odoo/odoo/blob/17.0/addons/point_of_sale/views/pos_ticket_view.xml#L160-L162

When I was checking it, I had similar issues, but at the end I choose to follow the same logic applied to other date fields.

ivantodorovich (approved):

PR #1105, opened 2023-12-06
[17.0][MIG] pos_lot_barcode: Migration to 17.0

nguyenminhchien:
ref: BSRD-704


Fabrizio1305:
Not working. 
Tested flow:

1. activate lots on products
2. take a product available in POS, and update qty to create an entry with a LOT (make sure the lot number doesn't exist yet for any product), make sure the product doesn't have qty in stock without a LOT (to avoid inconsistencies)
(optional: print the lot barcode)
3. open POS
4. add that product to a cart the standard way, with the lot number you just created. 
5. Scan the lot barcode (or activate debug panel and enter the lot number manually in the debug panel and push scan)
6. the same product as point 4 should be added automatically to the cart.
7. finalize the transaction by paying in cash and closing the POS
8. and double check that the proper product and lot has been taken out of stock (inventory moves)

In my test, I'm stuck on point 5 as the product/lot is not found on scanning (both in debug mode and with the physical scanner).
tested only together with #1106

nguyenminhchien:
> Not working. Tested flow:
> 
>     1. activate lots on products
> 
>     2. take a product available in POS, and update qty to create an entry with a LOT (make sure the lot number doesn't exist yet for any product), make sure the product doesn't have qty in stock without a LOT (to avoid inconsistencies)
>        (optional: print the lot barcode)
> 
>     3. open POS
> 
>     4. add that product to a cart the standard way, with the lot number you just created.
> 
>     5. Scan the lot barcode (or activate debug panel and enter the lot number manually in the debug panel and push scan)
> 
>     6. the same product as point 4 should be added automatically to the cart.
> 
>     7. finalize the transaction by paying in cash and closing the POS
> 
>     8. and double check that the proper product and lot has been taken out of stock (inventory moves)
> 
> 
> In my test, I'm stuck on point 5 as the product/lot is not found on scanning (both in debug mode and with the physical scanner). tested only together with #1106

@Fabrizio1305 hi, what Lot Barcode did you test ? did it march to Barcode pattern of Default Nomenclature you defined in your POS?
fyi, here is the default Barcode pattern added by Odoo
image

Fabrizio1305:
Good point, double checked, and lot line was after the product line, moved it above and now it finds the product but doesn't add it to the cart, a manual confirmation is needed, and slowing downs the process. in v16 was adding it directly to the cart.
Also when you scan products with the product's barcode there isn't a confirmation step, and it's normal, as the process should flow without hiccups.

nguyenminhchien:
@Fabrizio1305 the popup is shown when you select the product (track=Lot), or scan its barcode (NOT Lot barcode). If you scan the Lot barcode, related product will be added to cart immediately. Check my screencast below:

[Screencast from 12-12-2023 09:51:36.webm](https://github.com/OCA/pos/assets/8724793/90c3c2de-7a37-4b95-b53f-99583ed970ca)

gurneyalex (approved):

PR #1110, opened 2023-12-15
[14.0][IMP] pos_disable_pricelist_selection: selectable pricelists

ilyasProgrammer:


ilyasProgrammer:
@pedrobaeza i use Binary here, same as in **sale_timesheet_order_line_no_update**. But here it does not work. Could you hint what might be the reason?
`Error: JSON.parse: unexpected character at line 1 column 1 of the JSON data`

pos_disable_pricelist_selection/__manifest__.py
```diff
@@ -7,6 +7,7 @@
     "license": "LGPL-3",
     "category": "Purchase",
     "website": "https://github.com/OCA/pos",
+    "maintainers": ["ilyasprogrammer"],
     "depends": ["point_of_sale"],
```
HekkiMelody, line 11: **chore**: missing dependency on `web_domain_field`

pos_disable_pricelist_selection/models/pos_config.py
```diff
@@ -1,9 +1,75 @@
...
+    @api.onchange("hide_pricelist_button")
+    def onchange_hide_pricelist_button(self):
+        self.update(
+            {"selectable_pricelist_ids": [(6, 0, self.allowed_pricelist_ids.ids)]}
+        )
+
+    def write(self, vals):
+        if vals.get("available_pricelist_ids"):
+            if self and not vals.get("selectable_pricelist_ids"):
+                selectable = set(self.selectable_pricelist_ids.ids)
```
HekkiMelody: **issue**: `self` could be a recordset containing multiple records.

pos_disable_pricelist_selection/models/pos_config.py
```diff
@@ -1,9 +1,75 @@
...
+            else:
+                selectable = set(vals["selectable_pricelist_ids"][0][2])
+            # leave only ids from available_pricelist_ids
+            diff = list(
```
HekkiMelody: **todo**: this variable name is IMO misleading, since it's the intersection, and not the difference.

pos_disable_pricelist_selection/models/pos_config.py
```diff
@@ -1,9 +1,75 @@
...
+        "hide_pricelist_button", "allowed_pricelist_ids", "selectable_pricelist_ids"
+    )
+    def _compute_pricelist_id_domain(self):
+        for rec in self:
+            if rec.hide_pricelist_button:
+                rec.pricelist_id_domain = [("id", "in", rec.allowed_pricelist_ids.ids)]
+            else:
+                rec.pricelist_id_domain = [
+                    ("id", "in", rec.selectable_pricelist_ids.ids)
+                ]
```
HekkiMelody, line 42: **chore**: Documentation of the `web_domain_field` recommends using `json.dumps` to make sure any edge case is handled properly.

pos_disable_pricelist_selection/static/src/js/SetPricelistButton.js
```diff
@@ -0,0 +1,37 @@
...
+                        title: this.env._t("Select the pricelist"),
+                        list: selectionList,
+                    }
+                );
+
+                if (confirmed) {
+                    this.currentOrder.set_pricelist(selectedPricelist);
+                }
+            }
+        };
```
HekkiMelody: **question (non-blocking)**: I don't know if this could be a better approach, but I wonder if, instead of overwriting the entire method, we could quietly replace `this.env.pos.pricelists` with `this.env.pos.selectable_pricelists`, then call super(), and finally put things back where they belong. My main concern is improving compatibility with other modules. What do you think?

HekkiMelody (changes requested):
Code review. Please add some tests as well, especially for the `pos_config` part.

pos_disable_pricelist_selection/__manifest__.py
ilyasProgrammer: web_domain_field is not required. Turns out compute domain thing works without web_domain_field. Just need to use Binary field.

HekkiMelody: Huh, that's good to know. Thanks!

pos_disable_pricelist_selection/static/src/js/SetPricelistButton.js
ilyasProgrammer: agree

HekkiMelody (approved):
Code review, LGTM

pos_disable_pricelist_selection/models/pos_config.py
HekkiMelody: Couldn't find a way to cause an error, so considering this non-blocking.

francesco-ooops (approved):
Functional is ok. 

@legalsylvain what do you think?

francesco-ooops:
@pedrobaeza can this be merged based on existing reviews? thanks!

PR #1112, opened 2023-12-22
[14.0][FIX] pos_*: avoid located CI issues

chienandalu:
Ensure test correctness when a different default localization is set on the CI as the existing pricelist could be in a different currency and that will cause all kind of test failures.

cc @Tecnativa TT46702

### DIFF
--- a/pos_cash_move_reason/tests/test_pos_cash_move_reason.py
+++ b/pos_cash_move_reason/tests/test_pos_cash_move_reason.py
@@ -13,8 +13,34 @@ class TestPosCashMoveReason(SavepointCase):
         cls.PosSession = cls.env["pos.session"]
         cls.WizardReason = cls.env["wizard.pos.move.reason"]
         cls.AccountMoveLine = cls.env["account.move.line"]
-
-        cls.config = cls.env.ref("point_of_sale.pos_config_main").copy()
+        cls.pricelist = cls.env["product.pricelist"].create(
+            {
+                "name": "Test pricelist",
+                "item_ids": [
+                    (
+                        0,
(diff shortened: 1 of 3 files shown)


legalsylvain (commented):
thanks ! 

still draft ?

chienandalu:
Ready for review. I was testing them in batch in a local CI :)

pos_cash_move_reason/tests/test_pos_cash_move_reason.py
```diff
@@ -13,8 +13,28 @@ def setUpClass(cls):
...
+                    )
+                ],
+            }
+        )
+        cls.config = cls.env.ref("point_of_sale.pos_config_main").copy(
+            {
+                "available_pricelist_ids": [(6, 0, cls.pricelist.ids)],
+                "pricelist_id": cls.pricelist.id,
+            }
+        )
```
legalsylvain, line 37: why not use core function ? 

https://github.com/odoo/odoo/blob/14.0/addons/point_of_sale/tests/common.py#L596

legalsylvain (commented):
naive question inline ?

pos_cash_move_reason/tests/test_pos_cash_move_reason.py
chienandalu: I've given it a shot but it looks that class pulls too many stuff to just use it out of the box without much refactor. As we're moving to future versions with these customer, I'd let the research for `16.0`

legalsylvain (approved):
OK !

/ocabot Merge patch

PR #1118, opened 2024-01-04
[IMP] pos_partner_birthdate: Adding birthdate on partner list

etobella:
image

### DIFF
--- a/pos_partner_birthdate/static/src/js/PartnerLine.esm.js
+++ b/pos_partner_birthdate/static/src/js/PartnerLine.esm.js
@@ -1,0 +1,13 @@
+/** @odoo-module **/
+import {PartnerLine} from "@point_of_sale/app/screens/partner_list/partner_line/partner_line";
+import {formatDate} from "@web/core/l10n/dates";
+import {patch} from "@web/core/utils/patch";
+const {DateTime} = luxon;
+
+patch(PartnerLine.prototype, {
+    get formatedBirthdate() {
+        return formatDate(
+            DateTime.fromJSDate(new Date(this.props.partner.birthdate_date))
+        );
+    },
+});
--- a/pos_partner_birthdate/static/src/js/PartnerLine.esm.js
+++ b/pos_partner_birthdate/static/src/js/PartnerLine.esm.js
@@ -7,7 +7,9 @@ const {DateTime} = luxon;
 patch(PartnerLine.prototype, {
     get formatedBirthdate() {
         return formatDate(
-            DateTime.fromJSDate(new Date(this.props.partner.birthdate_date))
+            DateTime.fromJSDate(new Date(this.props.partner.birthdate_date)).setZone(
+                "UTC"
+            )
         );
     },
 });
--- a/pos_partner_birthdate/static/src/js/PosDB.esm.js
+++ b/pos_partner_birthdate/static/src/js/PosDB.esm.js
@@ -1,0 +1,27 @@
+/** @odoo-module **/
+import {PosDB} from "@point_of_sale/app/store/db";
+import {formatDate} from "@web/core/l10n/dates";
+import {patch} from "@web/core/utils/patch";
+const {DateTime} = luxon;
+
+patch(PosDB.prototype, {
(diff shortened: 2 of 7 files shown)


pos_partner_birthdate/static/src/xml/screens.xml
cvinh, line 1: Rename file to screens.xml as declated in manifest

pos_partner_birthdate/static/src/js/PosDB.esm.js
```diff
@@ -0,0 +1,20 @@
...
+    _partner_search_string(partner) {
+        var result = super._partner_search_string(partner);
+        if (partner.birthdate_date) {
+            result =
+                result.substr(0, result.length - 1) +
+                "|" +
+                formatDate(DateTime.fromJSDate(new Date(partner.birthdate_date))) +
+                "\n";
+        }
+        console.log(result);
```
cvinh: console.log can be removed

cvinh (changes requested):

pos_partner_birthdate/static/src/js/PosDB.esm.js
```diff
@@ -0,0 +1,19 @@
...
+const {DateTime} = luxon;
+
+patch(PosDB.prototype, {
+    _partner_search_string(partner) {
+        var result = super._partner_search_string(partner);
+        if (partner.birthdate_date) {
+            result =
+                result.substr(0, result.length - 1) +
+                "|" +
+                formatDate(DateTime.fromJSDate(new Date(partner.birthdate_date))) +
```
cvinh: maybe you can improve search to allow user search by date without format...
For example mmddyyyy or mmddyy (en) or ddmmyyyy or ddmmyy (fr)

pos_partner_birthdate/static/src/js/PartnerLine.esm.js
```diff
@@ -0,0 +1,13 @@
+/** @odoo-module **/
+import {PartnerLine} from "@point_of_sale/app/screens/partner_list/partner_line/partner_line";
+import {formatDate} from "@web/core/l10n/dates";
+import {patch} from "@web/core/utils/patch";
+const {DateTime} = luxon;
+
+patch(PartnerLine.prototype, {
+    get formatedBirthdate() {
+        return formatDate(
```
cvinh, line 9: formatDate should not be used as birthdate does not depend on tz
Steps to reproduce : setup your tz to gmt-10 nd the birthdate changes to -1day

pos_partner_birthdate/static/src/js/PosDB.esm.js
etobella: Log removed :wink:

cvinh (changes requested):

pos_partner_birthdate/static/src/js/PartnerLine.esm.js
etobella: You are right, I was able to reproduce the error forcing the zone of the navigator. I fixed it forcing UTC timezone :wink:

pos_partner_birthdate/static/src/js/PosDB.esm.js
etobella: The problem with this configuration is that we do not have this settings. It is not even accepted by Odoo, making it hard to manage :sob: 

Odoo tries to parse the search parameters to datetime in order to do that, however, it only works on you locale standard format

cvinh: The user does not want to enter the '/' (same as in the backend)
Maybe you can add the '/' with some tricky js or use some code from web... ?

etobella: Ok, I can add a second filter with cleaned data (removing `.`, `-`and `/`)

cvinh: It works like a charm

cvinh (approved):

ecino (approved):

PR #1119, opened 2024-01-04
[14.0][FIX] pos_order_return: empty returns

chienandalu:
If by mistake a cashier creates an empty return, the order won't be synched as the backend code will raise an exception. The order will be stuck and everytime it tries to sync it will raise an error.

For the moment, avoid trying to create an empty picking return, although we should forbid it in UX in the first place.

cc @Tecnativa TT46400

please review @pedrobaeza 

fyi @zamberjo


legalsylvain (approved):

pedrobaeza (approved):

zamberjo (approved):
I have not been able to test it, but it seems consistent.

PR #1123, opened 2024-01-08
[ADD] pos_lot_expiry

etobella:
Evaluate expiry of product in order to check if we can deliver the product


FrancoMaxime (approved):
LGTM: code review

pos_lot_expiry/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "Pos Lot Expiry",
+    "summary": """
+        Evaluate expiry of lot""",
+    "version": "17.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Dixmit,INVITU,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/pos",
+    "depends": ["point_of_sale"],
```
cvinh: I think it should depend on product_expiry as well

etobella: Completly right. It was working without it installed, but it has no sense if it is not installed

cvinh (commented):
At the end, I would have named the module pos_product_expiry instead of pos_lot_expiry, as expiry is obviously  related to lot by design
and as it depends from product_expiry, the logical seems better to me
WDYT ?

etobella:
Completly true!

pos_product_expiry/static/src/js/app/models.esm.js
```diff
@@ -0,0 +1,52 @@
...
+            try {
+                const checked_lots_problem = await this.env.services.orm.call(
+                    "product.product",
+                    "check_pos_lots",
+                    [[this.product.id], lotsToCheck, this.env.services.pos.company.id]
+                );
+                if (checked_lots_problem) {
+                    await this.env.services.popup.add(ErrorPopup, {
+                        title: _t("Problem with lots"),
+                        body: checked_lots_problem + _t("No changes were applied."),
```
cvinh: a space or a dot is missing before "No changes..."

etobella:
I applied the fixes and it is functional, however, it would be great to make it work better and enable offline mode. To do that, it would be great to integrate it with https://github.com/nguyenminhchien/pos/pull/3

I will wait until it is merged to make it work even better

etobella:
Integrated with pos_lot_selection in order to make it work properly with offline mode.

Also, I added some tests

cvinh (approved):

PR #1122, opened 2024-01-08
[16.0][FIX]pos_partner_firstname: Create and Update partner

DorianMAG:
- Add useState and add missed fields
- Update saveChanges() method
- Update template
- Remove jquery code

PR #1126, opened 2024-01-10
[17.0][MIG] pos_partner_firstname: Migration to 17.0

etobella:
Refactoring to adapt to OWL 2.0

Changes applied to avoid an extra connection to server


flotho:
Hi @etobella , thanks for this.
Did you have a look to https://github.com/OCA/pos/pull/1122 ?
Regards

etobella:
Yes I checked it, but at the end I made it in my own way using their comments, as the approach was much better :smile:

flotho:
Hum thanks @etobella ,
@DorianMAG , could you have a look to the @etobella proposal ?

etobella:
Also, I removed a connection to the server that was unnecessary :smile:

cvinh (approved):

cvinh (changes requested):
If we want to add a new partner with only firstname or lastname (not both)
We have an error
TypeError: this.showPopup is not a function
    at PartnerDetailsEdit.saveChanges

etobella:
It works now. Thanks for the review!

etobella:
Actually, it has no sense to require both on PoS but only one on backend, I will keep the logic of backend.

One option would be to keep the same logic using a configuration parameter on backend and frontend. WDYT?

cvinh:
> Actually, it has no sense to require both on PoS but only one on backend, I will keep the logic of backend.
> 
> One option would be to keep the same logic using a configuration parameter on backend and frontend. WDYT?

I agree. BTW in v16, firstname AND lastname are mandatory... @robyf70  do you know if it was a requirement for anyone ?
@alexis-via

robyf70:
> > Actually, it has no sense to require both on PoS but only one on backend, I will keep the logic of backend.
> > One option would be to keep the same logic using a configuration parameter on backend and frontend. WDYT?
> 
> I agree. BTW in v16, firstname AND lastname are mandatory... @robyf70 do you know if it was a requirement for anyone ? @alexis-via

This module was done to give the possibility to create a contact as person from the PoS as well

pos_partner_firstname/static/src/app/PartnerDetailsEdit.esm.js
```diff
@@ -0,0 +1,61 @@
...
+    },
+    saveChanges() {
+        if (this.changes.is_company) {
+            this.changes.lastname = this.changes.firstname = undefined;
+        } else {
+            if (this.checkPartnerPersonName()) {
+                return this.popup.add(ErrorPopup, {
+                    title: _t("Missing information"),
+                    body: _t("Customer firstname or lastname are required"),
+                });
```
cvinh, line 38: It should be 'Customer firstname or lastname is required'

cvinh:
> > > Actually, it has no sense to require both on PoS but only one on backend, I will keep the logic of backend.
> > > One option would be to keep the same logic using a configuration parameter on backend and frontend. WDYT?
> > 
> > 
> > I agree. BTW in v16, firstname AND lastname are mandatory... @robyf70 do you know if it was a requirement for anyone ? @alexis-via
> 
> This module was done to give the possibility to create a contact as person from the PoS as well

Thanks @robyf70 we noticed that firstname AND lastname were mandatory in previous versions. The idea is to follow partner_firstname logical which allows partner with firstname OR lastname
@etobella if there's a configuration parameter then it's logical to add it in partner_firstname and inherit it
IMHO I'm approving as it is

etobella:
Well, adding the constrain might seem a good idea, but we must remember that some countries do not have lastname, so it makes sense to not make it required

cvinh (approved):
Functional test ok

kluna1998 (approved):

PR #1128, opened 2024-01-13
[16.0] mig pos_margin_account_invoice_margin

legalsylvain:
This module is a glue module between two OCA modules, ``pos_margin`` (same repository) and ``account_invoice_margin``
(OCA/margin-analysis).

With this module, invoices created from point of sale have correct margin.

**Note**

When installing this module, the margin will be computed from invoices that has been generated from point of sale.


legalsylvain:
/ocabot migration pos_margin_account_invoice_margin

CC : @quentinDupont 
original reviewers : @chienandalu, @ivantodorovich (https://github.com/OCA/pos/pull/783)

ivantodorovich (approved):

quentinDupont (commented):
Don't understand this glue module :-D

Function `_action_create_invoice_line` doesn't seems to exists anymore
Tested without this module and invoices created from POS are well handle for margin

legalsylvain:
Hum.... very nice catch @quentinDupont ! 
Indeed, the test of the module is green BUT, the code of this module is not executed...

image

maybe It means that the code is useless ;-) I'll take a look more deeply in the next days.

thanks for your review.

legalsylvain:
>  Tested without this module and invoices created from POS are well handle for margin

Indeed. Tested with point_of_sale and OCA/account_invoice_margin installed. The invoice created from PoS has correct margin and purchase_price, without any extra module. 

https://github.com/OCA/OpenUpgrade/pull/4450 to be reviewed.

Thanks @quentinDupont !

PR #1130, opened 2024-01-16
[16.0][MIG] pos_pricelist_technical

legalsylvain:
* migration from 12.0 version. (related PR)

### Description

This module extends the functionality of Point of Sale to prevent technical pricelists that are used for the computation of other pricelists from being selected.

 ### Configure
* First configure the main module ``sale_pricelist_technical``.

* Configure your pricelist on your pos config form.

pos_config_form

### Usage

pos_front_end_ui


quentinDupont (commented):
hey !
Why not hiding technical pricelist directly on pos.config ?
thanks for this PR and migration

legalsylvain:
>  Why not hiding technical pricelist directly on pos.config ? 

Do you mean, removing the pricelist from the field ``available_pricelist_ids`` ?  If yes : 

Considering 
- Pricelist 1
- Pricelist 2 (IS TECHNICAL = Should not be used on a sale for customer)
- Pricelist 3, based on pricelist 2

If you don't load the pricelist 2 in the pos.config, it will not be available in the PoS. As a result the computation of the pricelist 3 will fail, because information are not available.

quentinDupont:
> If you don't load the pricelist 2 in the pos.config, it will not be available in the PoS. As a result the computation of the pricelist 3 will fail, because information are not available.

Ok i didn't know !
Thanks for the explanation :)

quentinDupont (approved):
functionnal and code review, thank you for this migration :heavy_check_mark:

PR #1133, opened 2024-01-18
[FIX] eslint warnings

legalsylvain:
Rational : for the time being, when we want to contribute to the OCA 16.0 branch and if we run pre-commit run -a, a lot of warning are poluting the log. So it's not easy to see if there is a real problem.

This PR fix all warnings. (a commit per type)

- 'no-shadow' 
- 'no-param-reassign'
- 'eqeqeq'
- 'sort-imports'
- 'radix'

Thanks for your review ! 

CC : @OCA/pos-maintainers

### DIFF
--- a/pos_product_multi_barcode/static/src/js/db.js
+++ b/pos_product_multi_barcode/static/src/js/db.js
@@ -16,9 +16,6 @@ odoo.define("pos_product_multi_barcode.db", function (require) {
             var res = this._super(products);
             var self = this;
 
-            if (!products instanceof Array) {
-                products = [products];
-            }
             products.forEach(function (product) {
                 var barcodes = JSON.parse(product.barcodes_json);
 
--- a/pos_access_right/static/src/js/ActionpadWidget.js
+++ b/pos_access_right/static/src/js/ActionpadWidget.js
@@ -4,8 +4,8 @@ odoo.define("pos_access_right.ActionpadWidget", function (require) {
     const Registries = require("point_of_sale.Registries");
     const ActionpadWidget = require("point_of_sale.ActionpadWidget");
 
-    const PosActionpadWidget = (ActionpadWidget) =>
-        class extends ActionpadWidget {
+    const PosActionpadWidget = (OriginalActionpadWidget) =>
+        class extends OriginalActionpadWidget {
             get hasPaymentControlRights() {
                 return this.env.pos.user.hasGroupPayment;
             }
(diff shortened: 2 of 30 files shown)


chienandalu (approved):
Great!

(isn't it `eslint` anyway?)

pos_product_multi_barcode/static/src/js/db.js
```diff
@@ -16,9 +16,6 @@ odoo.define("pos_product_multi_barcode.db", function (require) {
             var res = this._super(products);
             var self = this;
 
-            if (!products instanceof Array) {
-                products = [products];
-            }
```
rvalyi, line 21: sorry not a Javascript expert here, but are you sure this can be simply removed like that?

legalsylvain: super is called before, and the original function does the job. So it is not possible to not have an array at this step or did I missed something ?

https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/db.js#L192C1-L194C1

chienandalu: Mmm, I'd say it isn't mutable

legalsylvain: Ow Thanks ! I didn't know. So what is the good way here ? Add an eslint exceptiont ?

chienandalu: Well, I'm wrong and is indeed mutable as is part of the `arguments` object: https://eslint.org/docs/latest/rules/no-param-reassign

legalsylvain: So it is OK like that. Right ?

chienandalu: I'd say so

rvalyi: > So it is OK like that. Right ?

yes sorry I didn't see the job was done already in super.

rvalyi (approved):

legalsylvain:
> (isn't it eslint anyway?)

Indeed ! typo fixed in commits texts.

pos_product_multi_barcode/static/src/js/db.js
ivantodorovich: > Strict mode code doesn’t sync indices of the arguments object with each parameter binding. Therefore, this rule is not necessary to protect against arguments object mutation in ESM modules or other strict mode functions.

I wouldn't remove it. According to that it's only mutable if running old code that hasn't been flagged with `"use strict";`

As we're slowly migrating to ESM, we can't rely on the argument being mutable (which is a bad thing, solved in "modern js")

chienandalu: Thanks @ivantodorovich I didn't get that nuance :) Wouldn't the lint rule be  a bit misguiding being that virtually all the code is run in strict mode?

ivantodorovich: Indeed, since we already have the `strict` rule, it seems we could remove the `no-param-reassign` one

PR #1134, opened 2024-01-18
[REF] copier Upate bump to v1.20 (from v1.17).

legalsylvain:
[REF] Do not use OCB version, to save resources

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,7 +1,8 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.17.2
+_commit: v1.20
 _src_path: gh:oca/oca-addons-repo-template
 ci: GitHub
+convert_readme_fragments_to_markdown: false
 generate_requirements_txt: true
 github_check_license: true
 github_ci_extra_env: {}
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
I don't think you should save resources this way, and more in a recent version. And in any case, you should opt for OCB, not Odoo.

legalsylvain:
> I don't think you should save resources this way, and more in a recent version. 

Hi @pedrobaeza, thanks for your quick answer.
- I didn't know that OCA policy. 
- I see the OCB project as abandonned by OCA. in V16, there is 3 merged commits (differents than in odoo/odoo) : https://github.com/OCA/OCB/pulls?q=is%3Apr+base%3A16.0 : One to flag the repo as OCB and two regarding autoinstallation of iap.

> And in any case, you should opt for OCB, not Odoo.

Well, we should ensure that OCA modules are working with odoo, not OCB. (even if it is preferable). odoo is used by millions of users, OCB, by some specific integrators.

pedrobaeza:
Well, it's not abandoned, it's simply that all the patches has landed on Odoo, but at the end OCB is more completed than Odoo, if you remove one of both. Take also into account that people has not yet landed too much on v16, so there's not a lot of PRs.

pedrobaeza (approved):

legalsylvain:
Hi @pedrobaeza. thanks for your review. I reverted the change regarding OCB/odoo, and I opened a dedicated issue in the right place : https://github.com/OCA/oca-addons-repo-template/issues/244

PR #1136, opened 2024-01-22
[ADD] pos_payment_usability : module to try to fix bad UX in payment PoS screen

legalsylvain:
This module extends the functionality of point of sale module, improving the display of the payment screen.

**Without the module**

without_module

**With the module**

with_module


quentinDupont (approved):
Works as described and code reviewed :heavy_check_mark: 
But i'm not fully convinced : which resolution screen size have you been using ? because it seems that button "validate" hide summary if the screen is quite very small.

legalsylvain:
>  But i'm not fully convinced : which resolution screen size have you been using ? because it seems that button "validate" hide summary if the screen is quite very small.

I'm not sure to understand : Step to reproduce. create 5 payment method and launch the PoS with Firefox with a simple shortcut navbar : 
- 1280 x **720** : No payment line visible ! :-1: 
- 1280 x **800** : No payment line visible ! :-1: 
- 1440 x **900** : 1 single payment line visible :-1: 
- 1280 x **1024** : 3 payment lines visible. -> :ok: 
- 1280 x **1050** : 4 payment lines visible. -> :ok: 
- 1280 x **1080** : 4 payment lines visible. -> :ok: 

So under a height of 1000 PoS is not working. I know that recent screen are > 1000, but to work with devices older, or people that want to have bigger text, this module is interesting.

PR #1138, opened 2024-01-23
[16.0][ADD] pos_product_warning

MS-OSI:
POS Product Warnings
==================

This module displays product warnings while processing payments on the POS screen.


MS-OSI:
@dreispt

PR #1139, opened 2024-01-23
[16.0][ADD] pos_product_warning

MS-OSI:
POS Product Warnings
==================

This module displays product warnings while processing payments on the POS screen.


MS-OSI:
@dreispt

legalsylvain:
Hi @MS-OSI. 

Thanks for sharing your module.

Design question : There is a ``sale_line_warn_msg`` field, present in sale module. We could reuse this field instead of adding a new one, don't you think ? 

Functional question : just tested on runboat. I think that the warning should be done when selecting the product (like in the sale and the purchase module)
- homogeneous behaviour
- if you go back on the product screen, you'll have twice the message
Don't you think ?

dreispt:
@legalsylvain I worked on the design for this, so I can share the thinking behind it.

We decided for a separate warning field, for flexibility.
This allows different warnings on different sales channels.
In case you want the same warning in both cases, it is not hard to automated the syncing between the two fields.

The main use case is to check for a person's ID in case of selling certain items, such as alcohol.
I don't want a check for each beer added.
I want to remind the operator to ask for ID because the sale includes alcoholic drinks.

pos_product_warning/static/src/js/ProductScreen.esm.js
```diff
@@ -0,0 +1,30 @@
...
+        async _onClickPay() {
+            const order = this.env.pos.get_order();
+            const orderlines = order
+                .get_orderlines()
+                .filter((line) => line.product.pos_warn_msg && line.quantity >= 0);
+            const productSaleWarningList = _.uniq(
+                _.pluck(_.pluck(orderlines, "product"), "pos_warn_msg")
+            );
+            if (productSaleWarningList.length > 0) {
+                const {confirmed} = await this.showPopup("ProductWarningPopup", {
```
ammarofficewla, line 17: Can we not use 'SelectionPopup' instead of creating a new popup?

murtuzasaleh: > Can we not use 'SelectionPopup' instead of creating a new popup?

SelectionPOPUP is used for the choice of item but in our case, it is a confirmation warning not choice of items that's why we create a new popup.

To confirm a message in SelectionPOPUP, you need to click on one of the message lines.

image

MS-OSI: > Can we not use 'SelectionPopup' instead of creating a new popup?

SelectionPOPUP is used for the choice of item but in our case, it is a confirmation warning not choice of items that's why we create a new popup.

To confirm a message in SelectionPOPUP, you need to click on one of the message lines.

image

legalsylvain:
> The main use case is to check for a person's ID in case of selling certain items, such as alcohol.
> I don't want a check for each beer added.
> I want to remind the operator to ask for ID because the sale includes alcoholic drinks.

Ow. I understand the use case. But in that case you describe, maybe if the message is at product category / pos.category level, it will be less configuration. In a "little" shop, you can have 200 products containing alcool. Set the message at product level can take time to configure. don't you think ?

cvinh:
> > The main use case is to check for a person's ID in case of selling certain items, such as alcohol.
> > I don't want a check for each beer added.
> > I want to remind the operator to ask for ID because the sale includes alcoholic drinks.
> 
> Ow. I understand the use case. But in that case you describe, maybe if the message is at product category / pos.category level, it will be less configuration. In a "little" shop, you can have 200 products containing alcool. Set the message at product level can take time to configure. don't you think ?

It would be another module pos_category_warning

PR #1140, opened 2024-01-24
[MIG] pos_return_voucher: Migration to 16.0

manuelcalerosolis:
Once the functionality of this new module has been verified in v 14.0 (https://github.com/OCA/pos/pull/996/files), and with the need to launch it in a version v 16.0 I start this PR.

Communicated with the author of the module @zamberjo, he invited me to begin the migration to v 16.0 to complete it among all those interested in this PR.

I think that the part that needs refactoring the most is Javascript, and I invite all those interested to help the migration of this module, which will be so interesting to have in OCA.


chienandalu (commented):
My main concern is the same I expressed to @zamberjo and in this version is more relevant than in the one this module was proposed: this should fill the gap (if there's any) between the existing `pos_loyalty`  functionality (coupons, ewallets) and this module functionality.

manuelcalerosolis:
If I thought that it would also be very convenient to hook up with the entire pos_loyalty system, when it came to collecting payments at POS it would already be solved without the need for anything else. 

But I have a question, can these rewards be created anonymously, without having to relate them to a client?

chienandalu:
> But I have a question, can these rewards be created anonymously, without having to relate them to a client?

Coupons at least can be. With eWallets I don't have that much experience

manuelcalerosolis:
@zamberjo and @chienandalu I'll tell you what I see in the module, the paython part is functional but where I'm stuck is when paying with a payment method "Vale" that has the "return_voucher" field activated in Javacript, it returns the value of "undfinded" and is not taking the value from the database correctly. 

In the PaymentScreen.js file the value of `paymentMethod.return_voucher` is not set correctly. 

Let's see if you can help me to continue advancing in the migration of the module.

Thanks

anmarmo1:
@chienandalu The problem with discount and loyalty modules is they don't cover the scenario where the customer doesn't receive a refund upon returning a product. Instead, a credit balance is retained for future purchases.
This could be managed using a gift card or an electronic wallet, but in Odoo, it's currently handled with a product that deducts from the total order amount. This case, however, is different because the voucher is not something that someone has purchased in advance. That's why we associate it with a payment method and a prior refund. I haven't found anything in v15, v16, or v17 that addresses this particular situation. 
I'm not sure if it's technically feasible to accommodate this use case by adding it as a program type to gift cards, utilizing some of that functionality @zamberjo What do you think regarding the last part? Do you consider it feasible?

zamberjo:
We could adapt the logic to the `loyalty.program` model by adding it as a new type:
image

And allow to be able to visualize there the vouchers applied from the POS.
image

@manuelcalerosolis , would you take care of the adaptation? If you don't see it clear we could take care of it as soon as we have less workload, I don't know how urgently you need the module.

@chienandalu , do you see this solution viable?

manuelcalerosolis:
@zamberjo what you propose may be an improvement and make the module a little lighter, but I think it should first work as it does in v 14.0. 

If you can see the JavaScript issue, I'll start trying to use the `loyalty.program` modules.

PR #1142, opened 2024-01-26
 [ADD] pos_sale_order_load: Importing sales orders at the POS

aayartsev:
This PR continue work from this PR https://github.com/OCA/pos/pull/841


DemchukM (approved):
LGTM

francesco-ooops (approved):
Functional ok!

@ygcarvalh @legalsylvain can you take a look?

legalsylvain (commented):
No review. Thanks for your contribution.

/ocabot merge patch

francesco-ooops:
@legalsylvain thanks! never seen the issue above before.

sbidoul:
@legalsylvain @francesco-ooops this is because https://pypi.org/project/odoo14-addon-pos-sale-order-load/ is owned by coopdevs on PyPI. Is it the same module?

francesco-ooops:
@sbidoul this is an odoo SA backporting, name was chosen as suggested by @legalsylvain

sbidoul:
In any case, we need to discuss with coopdevs, to decide if they can transfer the name to OCA, or move their module to OCA in place of this one, or if we need to rename the module here.

francesco-ooops:
I didn't know there were specific rules on how to name new modules, but it would be good to have such errors raised before merge in that case

sbidoul:
> I didn't know there were specific rules on how to name new modules, but it would be good to have such errors raised before merge in that case

I agree such a check would be useful. A difficulty is that neither PyPI nor the Odoo apps store have an API that return the owner of a package.

pos_sale_order_load/static/src/js/OrderManagementScreen/SaleOrderManagementScreen.js
```diff
@@ -0,0 +1,340 @@
...
+                    await this.env.pos.load_new_partners();
+                } catch (error) {}
+                const order_partner = this.env.pos.db.get_partner_by_id(
+                    sale_order.partner_id[0]
+                );
+                if (order_partner) {
+                    currentPOSOrder.set_client(order_partner);
+                } else {
+                    try {
+                        await this.env.pos._loadPartners([sale_order.partner_id[0]]);
```
vuwnevska, line 109: _loadPartners is not present in v14

PR #1144, opened 2024-01-26
[ADD] pos_product_packaging_container_deposit: Introduce new module in 16.0

nguyenminhchien:
Ref: BSALLO-607


pos_product_packaging_container_deposit/models/pos_order_line.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PosOrderLine(models.Model):
+    _inherit = "pos.order.line"
+
+    is_container_deposit = fields.Boolean()
```
santostelmo, line 10: The field is already present in `container.deposit.order.line.mixin`

nguyenminhchien: There is no field for the function `def _get_product_qty_delivered_received_field(self)`. So, i don't make pos.order.line inherits from that mixin model.

santostelmo: Indeed. Inherit mixin just to have the field `is_container_deposit` it's not worth it

santostelmo:
Functional test done.  ✅

santostelmo (approved):

santostelmo (approved):
LG

simahawk (approved):
LG. Minor remark for the commit msg: `[ADD] pos_product_packaging_container_deposit` is enough :wink:

nguyenminhchien:
> LG. Minor remark for the commit msg: `[ADD] pos_product_packaging_container_deposit` is enough 😉

i updated the commit message.

PR #1145, opened 2024-01-26
[16.0] mig pos_product_mergeable_line

legalsylvain:
Quite trivial migration of module pos_product_mergeable_line from 12.0 to 16.0

### Description

This module extends the Odoo Point of Sale to prevent to merge lines if product
is configured.

By default, new line can be merged with a previous one if the informations are the same. (same product, same restaurant note, etc...) and if the UoM Category allow it.

The new module add a boolean field 'Mergeable Line' (default True).

If unchecked, the product will never be merged into another line.

That's important in some context, as in the management of returnable products, where we want to have a record of all movements.

### Configure

* Go to "Point of Sale > Product" and edit a product

* Uncheck the checkbox "Mergeable line"

product_form


pos_product_mergeable_line/static/src/js/models.js
```diff
@@ -0,0 +1,28 @@
...
+odoo.define("pos_product_mergeable_line.models", function (require) {
+    "use strict";
+
+    const {Orderline} = require("point_of_sale.models");
+    const Registries = require("point_of_sale.Registries");
+
+    const PosProductMergeableLineOrderline = (OriginalOrderline) =>
+        class extends OriginalOrderline {
+            can_be_merged_with(orderline) {
+                console.log("==================");
```
quentinDupont: you can remove console log :)

quentinDupont (changes requested):
Otherwise functional review :heavy_check_mark:

pos_product_mergeable_line/static/src/js/models.js
legalsylvain: indeed ! Thanks ! 
Note : I thought that console.logs were forbidden and that the CI would be red, but that's not the case.

quentinDupont (approved):

PR #1151, opened 2024-02-07
[16.0] [ADD] pos_receipt_replace_user_by_trigram

trisdoan:
replace cashier's name by his trigram (first letter of firstname + first 2 letters of lastname) on POS receipt

### DIFF
--- a/pos_stock_available_online/i18n/es.po
+++ b/pos_stock_available_online/i18n/es.po
@@ -6,15 +6,15 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 16.0\n"
 "Report-Msgid-Bugs-To: \n"
-"PO-Revision-Date: 2023-03-31 08:15+0000\n"
-"Last-Translator: FLNerea <nerea.fernandez@factorlibre.com>\n"
+"PO-Revision-Date: 2024-03-11 09:52+0000\n"
+"Last-Translator: AdrianaSaiz <adrianasaizdelafuente@gmail.com>\n"
 "Language-Team: none\n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: \n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Weblate 4.14.1\n"
+"X-Generator: Weblate 4.17\n"
 
 #. module: pos_stock_available_online
 #: model:ir.model.fields,field_description:pos_stock_available_online.field_pos_config__additional_warehouse_ids
@@ -92,7 +92,7 @@ msgstr "Indicar la cantidad en los siguientes almacenes además de"
 #: code:addons/pos_stock_available_online/static/src/xml/Screens/ProductScreen/ProductItem.xml:0
 #, python-format
 msgid "Total"
-msgstr "Indicar la cantidad en los siguientes almacenes además de"
+msgstr "Total"
 
 #. module: pos_stock_available_online
 #: model:ir.model,name:pos_stock_available_online.model_stock_warehouse


pos_receipt_replace_user_by_trigram/models/pos_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    replace_user_by_trigram = fields.Boolean(
+        default=False, string="Replace User by trigram in POS receipt"
+    )
+
+
+class ResConfigSettings(models.TransientModel):
```
legalsylvain: nitpicking : could you split the file ? (one file per model).

pos_receipt_replace_user_by_trigram/models/res_users.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    trigram = fields.Char(compute="_compute_trigram")
```
legalsylvain:
```suggestion
    pos_trigram = fields.Char(compute="_compute_pos_trigram")
```
.It is used only in Point of Sale, to avoid possible collision

pos_receipt_replace_user_by_trigram/models/res_users.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @api.depends("firstname", "lastname")
+    def _compute_trigram(self):
+        for user in self:
+            user.trigram = ""
+            if user.firstname:
+                user.trigram += user.firstname[:1]
+            if user.lastname:
+                user.trigram += user.lastname[:2]
```
legalsylvain: This algorithm doesn't generate trigram in all cases.
Ex : 
- first Name : ""
- last name : "Doe"
- Result: trigram : "DO"

- first Name : "John"
- last name : ""
- Result: trigram : "J"

It deserve maybe a little test.

legalsylvain (commented):
Hi @trisdoan Thanks for your contribution.
some minor remarks inline.

otherwise. LGTM !

pos_receipt_replace_user_by_trigram/models/res_users.py
ivantodorovich: Alternatively, extract the trigram field and it's generation algorithm to a generic `base_user_trigram` and contribute to https://github.com/OCA/server-backend

pos_receipt_replace_user_by_trigram/models/res_users.py
ivantodorovich: > It deserve maybe a little test.

indeed

santostelmo:
@trisdoan there was a couple of comments in the OCA PR. => unit test, if you manage to add it will be perfect

pos_receipt_replace_user_by_trigram/models/res_users.py
simahawk: yes please

pos_receipt_replace_user_by_trigram/models/res_users.py
simahawk: for now I would simply move the computation to a util func that could be moved later out of the module w/ less impact

simahawk (changes requested):

pos_receipt_replace_user_by_trigram/models/res_users.py
simahawk:
```suggestion
            user.trigram = utils.get_trigram(user.firstname, user.lastname)
```
utils.py
```
def get_trigram(*pargs):
    valid = [x.strip() for x in pargs if x.strip()]
    # do your magic here
    return trigram
```

trisdoan:
Thank you so much @legalsylvain @simahawk @ivantodorovich for your in-depth reviews

I updated the PR to take into account your inputs, including utils.py and some test cases for the function

pos_receipt_replace_user_by_trigram/views/res_users.xml
```diff
@@ -0,0 +1,12 @@
+<odoo>
+    <record id="view_users_form_inherit" model="ir.ui.view">
+        <field name="name">Add Trigram</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="partner_firstname.view_users_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='firstname']" position="after">
+                <field name="pos_trigram" readonly="1" />
```
ivantodorovich:
```suggestion
                <field name="pos_trigram" />
```

nitpicking: redundant

pos_receipt_replace_user_by_trigram/views/res_config_settings.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record
+        id="pos_receipt_hide_user_res_config_settings_view_form_inherit"
+        model="ir.ui.view"
+    >
+        <field
+            name="name"
+        >pos.receipt.hide.user.res.config.settings.view.form.inherit</field>
```
ivantodorovich:
```suggestion
```

💡 tip: odoo automatically generates the view's name

unless you want to give a meaningful name, you can ignore this :)

ivantodorovich (approved):
Thanks!! LGTM 👍🏻 

My following comments are absolutely non-blocking

santostelmo (approved):

simahawk (commented):
LG. The build is red. Can you check?

trisdoan:
Hi @ivantodorovich, I pushed a new commit to support feature multi-employee-per-session.

Screenshot from 2024-03-11 10-36-28

### Context
- When the feature is checked, `get_cashier` is overridden by `set_cashier` in `pos_hr`

  - As consequences, `pos_trigram` is null
 
### This change

`pos_receipt_replace_user_by_triagram`

- `pos_trigram` is defined in `res.partner` instead of res.users.
  
  - `res.users` inherits `res.partner`
  - can link `res.partner` and `hr.employee` via related field `work_contact_id`

`pos_receipt_replace_user_by_triagram_hr`

- It is a glue module between `pos_receipt_replace_user_by_triagram` & `pos_hr`
- I take advantage of the module to pass the field `pos_trigram` into `_loader_params_hr_employee`, so it can be accessed on model Pos Order

pos_receipt_replace_user_by_trigram/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     "installable": True,
     "version": "16.0.1.0.0",
     "depends": [
-        "point_of_sale",
+        "pos_hr",
```
simahawk: this should go in a glue module IMO

simahawk (changes requested):
@ivantodorovich can you review again pls?

pos_receipt_replace_user_by_trigram/__manifest__.py
trisdoan: Hi @simahawk, thank you for your in-depth review 👍️

I separated the modules as you suggested.

trisdoan:
superseded by https://github.com/OCA/pos/pull/1167

PR #1153, opened 2024-02-09
[IMP] pos_partner_firstname: set name invisible instead of readonly when is_company=False

alexis-via:
From my experience with my end users, in POS partner form view, when is_company=False, they try to use the "name" field because it's not obvious that the field is readonly in the user interface. It's better to put that field invisible when is_company=false.


legalsylvain (approved):
code review / no test.

cvinh:
Hello can you please do a PR for v17 if needed ?

alexis-via:
If this PR is merged, then I'll forward port it to v17.

flotho (approved):
easy improvement
LGTM

cvinh:
> If this PR is merged, then I'll forward port it to v17.

I cherry-picked for you
https://github.com/OCA/pos/pull/1156

PR #1157, opened 2024-02-19
[16.0] pos_self_service_*

carmenbianca:
This is a heavily WIP module to use the Odoo POS as a self-service terminal for customers. Customers will be able to:

- Weigh their own products, like fruit at a supermarket.
- Tare their containers, like bringing a pot in which to put any amount of rice.
  - And subsequently, weigh that rice with the weight of the container removed.

It'll be a while before this is finished. Opening the PR early for early reviews from colleagues. Other OCA contributors needn't look at this PR for the moment.


carmenbianca:
Pinging @huguesdk for an early eye-over of the code to see if it passes the sniff test.

pos_self_service_weight_base/__manifest__.py
huguesdk, line 1: good idea to mention “weight” in the module name. it seems that i forgot about that. 👍

pos_self_service_weight_base/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    iface_self_service = fields.Boolean(
```
huguesdk: i’m wondering about `iface` in the name. i see other settings like that in odoo, but they all seem directly related to hardware devices. what do you think about `is_self_service_weight` or `enable_self_service_weight`?

pos_self_service_weight_base/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
...
+        string="Is Self-Service",
+        help="Use this POS as self-service point",
```
huguesdk: i think mentioning “weight” would be useful.

pos_self_service_weight_base/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    pos_iface_self_service = fields.Boolean(
```
huguesdk: same here: what about `pos_is_self_service_weight`?

pos_self_service_weight_base/static/src/css/pos.css
huguesdk, line 1: where are all these css rules coming from? is that self-made or copied from somewhere else?

pos_self_service_weight_base/static/src/js/Chrome.js
```diff
@@ -0,0 +1,29 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+odoo.define("pos_self_service_weight_base.Chrome", function (require) {
+    "use strict";
+
+    const Chrome = require("point_of_sale.Chrome");
+    const Registries = require("point_of_sale.Registries");
+
+    const SelfServiceChrome = (Chrome_) =>
```
huguesdk: any reason for the underscore in `Chrome_`? i know it is a local variable independent of the one defined above, but i see that most oca modules use the same name.

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/WeightWidget.js
```diff
@@ -0,0 +1,49 @@
...
+
+odoo.define("pos_self_service_weight_base.WeightWidget", function (require) {
+    "use strict";
+
+    const PosComponent = require("point_of_sale.PosComponent");
+    const Registries = require("point_of_sale.Registries");
+
+    const {onMounted, onWillUnmount, useState} = owl;
+
+    class WeightWidget extends PosComponent {
```
huguesdk: could this inherit from `point_of_sale.ScaleScreen` to avoid re-implementing similar methods?

huguesdk (changes requested):
thanks for this! it seems like a strong base. some things i noticed while testing it:

when it is not able to connect to the iotbox, a red dialog box appears saying:

> **Unknown Error**
> 
> Unable to show information about this error.

a new one appears on top of the previous one every few seconds. it would be nice to have a more informative message (or an icon somewhere showing the status).

moreover, it doesn’t seem to take the status of the scale into account: even if it is reported as disconnected, it will read the weight repeatedly.

with this module loaded, the normal (not self-service weight) point of sale interface is different (while it shouldn’t): to purple odoo top bar is not visible.

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/SelfServiceHomeButton.js
```diff
@@ -0,0 +1,38 @@
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+odoo.define("pos_self_service_weight_base.SelfServiceHomeButton", function (require) {
+    "use strict";
+
+    const SelfServiceControlButton = require("pos_self_service_weight_base.SelfServiceControlButton");
+    const Registries = require("point_of_sale.Registries");
+    const {useListener} = require("@web/core/utils/hooks");
```
robinkeunen: @huguesdk i don't understand the difference between these two syntaxes, can you enlighten me ?

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/WeightWidget.js
```diff
@@ -0,0 +1,49 @@
...
+        onWillUnmount() {
+            // Stop the scale reading
+            this.env.proxy_queue.clear();
+        }
+        _readScale() {
+            this.env.proxy_queue.schedule(this._setWeight.bind(this), {
+                duration: 500,
+                repeat: true,
+            });
+        }
```
robinkeunen: Do I understand correctly ? Once the weight widget is mounted, `_read_scale` schedules a call of `_set_weight` every 500ms ?

I do not understand the `bind` instruction.

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/SelfServiceHomeButton.js
huguesdk: the second syntax is called [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and in this case it is equivalent to:

```
    const useListener = require("@web/core/utils/hooks").useListener;
```

it avoids repeating the variable name, but it is even more useful when assigning multiple variables together, like this:

```
    const {useListener, useService} = require("@web/core/utils/hooks");
```

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/WeightWidget.js
huguesdk: unlike in python, in javascript a reference to a method is not bound to a `this` variable. `bind()` returns a new function that, when called, provides the `this` variable to the method. more information [on mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind).

i’m not familiar with the `schedule()` function, but as i understand it, it will call `this._setWeight()` every 500 ms, forever.

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/SelfServiceHomeButton.js
robinkeunen: thanks !

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/WeightWidget.js
robinkeunen: Thanks !

robinkeunen:
@huguesdk should i consider translating the code as a [Native JS Module](https://www.odoo.com/documentation/16.0/developer/reference/frontend/javascript_modules.html#native-javascript-modules) ?

robinkeunen:
squashed previous work

pos_self_service_weight_base/models/pos_config.py
robinkeunen: done in 4d2c0346eddc3af837829c0e32daba7ed754b46b

pos_self_service_weight_base/models/pos_config.py
robinkeunen: done in 4d2c0346eddc3af837829c0e32daba7ed754b46b

pos_self_service_weight_base/models/res_config_settings.py
robinkeunen: done in 4d2c0346eddc3af837829c0e32daba7ed754b46b

pos_self_service_weight_base/static/src/js/Chrome.js
robinkeunen: I like Carmen's way better but I changed it anyway for consistency.

cf 4d2c0346eddc3af837829c0e32daba7ed754b46b

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/WeightWidget.js
robinkeunen: I'm not sure it's a good idea : 
- the template displays the order button : should be hidden in both tare and product mode
- the template displays the price, it should be hidden in tare mode
- it needs to be loaded with a product
- it is usually called as a popup which returns a payload

If there is a clean way to do it, please point in the right direction :-)

pos_self_service_weight_base/static/src/css/pos.css
```diff
@@ -0,0 +1,136 @@
...
+.pos .self-service-action-buttons .button .ico {
+    display: inline-block;
+    background-size: contain;
+    width: 170px;
+    height: 170px;
+    order: 2;
+}
+
+.foodbox {
+    background-image: url("../../img/empty_foodbox_on_scale.svg");
```
huguesdk: where is this file?

is all this css needed at this point or is this mostly coming from the old `pos_self_service_base`?

pos_self_service_weight_base/static/src/js/Chrome.esm.js
```diff
@@ -0,0 +1,35 @@
+/** @odoo-module alias=pos_self_service_weight_base.Chrome **/
```
huguesdk: is the alias needed? (same remark for all modules)

pos_self_service_weight_base/static/src/js/Chrome.esm.js
```diff
@@ -0,0 +1,35 @@
...
+            // Calling super first because it logs an error in some cases.
+            var result = super.startScreen;
+            if (this.env.pos.config.is_self_service_weight_point) {
+                return {name: "SelfServiceWelcomeScreen"};
+            }
+            return result;
+        }
+
+        showTicketButton() {
+            return (
```
huguesdk: a why comment would help to understand what this method is for.

pos_self_service_weight_base/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    is_self_service_weight_point = fields.Boolean(
+        string="Is Self-Service",
```
huguesdk:
```suggestion
        string="Is a Self-Service Weighing Station",
```

pos_self_service_weight_base/views/pos_config_views.xml
```diff
@@ -0,0 +1,37 @@
...
+                    invisible="not context.get('pos_config_create_mode', False)"
+                >
+                    <div class="col-12 col-lg-6 o_setting_box" id="is_self_service_weight_point">
+                        <div class="o_setting_left_pane">
+                            <field name="is_self_service_weight_point" />
+                        </div>
+                        <div class="o_setting_right_pane">
+                            <label for="is_self_service_weight_point" />
+                            <div class="text-muted mb16" id="is_self_service_weight_point_text">
+                                Use your PoS as self-service
```
huguesdk:
```suggestion
                                Use this PoS as a self-service weighing station
```

pos_self_service_weight_base/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
...
+        help="Use this POS as self-service weight point",
```
huguesdk:
```suggestion
        help="Use this PoS as a self-service weighing station",
```

pos_self_service_weight_base/static/src/xml/Popups/CloseSelfServicePopup.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template" xml:space="preserve">
+
+    <t t-name="pos_self_service_weight_base.CloseSelfServicePopup" owl="1">
+        <div class="popup close-self-service-popup">
+            <header>
+                <div class="title">Closing Weighting Point</div>
+            </header>
+            <main class="body">
+                <div>Are you sure you want to close the weighting point ?</div>
```
huguesdk:
```suggestion
                <div class="title">Close Weighing Station</div>
            </header>
            <main class="body">
                <div>Are you sure you want to close the weighing station?</div>
```
weighing means to weigh something, while weighting means to add weight to something. (non-breaking) space before a question mark is a french typographic thing, not applicable in english.

pos_self_service_weight_base/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Configure a PoS to be a self-service station where customers can weigh things. As a base module, this covers primarily the welcome screen and some common functionalities.
```
huguesdk:
```suggestion
Configure a PoS to be a self-service station where customers can weigh products. As a base module, this covers primarily the welcome screen and some common functionalities.
```

pos_self_service_weight_base/__manifest__.py
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+{
+    "name": "Point of Sale weighing self-service base module",
```
huguesdk:
```suggestion
    "name": "Point of Sale self-service weighing base module",
```

pos_self_service_weight_base/README.rst
huguesdk, line 1: actually, the module should more correctly be named `pos_self_service_weighing_base`.

pos_self_service_weight_base/models/pos_config.py
```diff
@@ -0,0 +1,14 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    is_self_service_weight_point = fields.Boolean(
```
huguesdk:
```suggestion
    is_self_service_weighing_station = fields.Boolean(
```

pos_self_service_weight_base/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    pos_is_self_service_weight_point = fields.Boolean(
+        related="pos_config_id.is_self_service_weight_point", readonly=False
```
huguesdk:
```suggestion
    pos_is_self_service_weighing_station = fields.Boolean(
        related="pos_config_id.is_self_service_weighing_station", readonly=False
```

pos_self_service_weight_base/static/src/js/ChromeWidgets/HeaderButton.esm.js
```diff
@@ -0,0 +1,39 @@
...
+                const info = await this.env.pos.getClosePosInfo();
+                this.showPopup("CloseSelfServicePopup", {info: info, keepBehind: true});
+            } catch (e) {
+                if (isConnectionError(e)) {
+                    this.showPopup("OfflineErrorPopup", {
+                        title: this.env._t("Network Error"),
+                        body: this.env._t(
+                            "Please check your internet connection and try again."
+                        ),
+                    });
```
huguesdk: why should an internet connection error prevent to close the weighing station? ok, i just saw that this is a copy of the `HeaderButton.onClick()` method with the `showPopup()` argument changed. isn’t there a better way? why not override the `showPopup()` method instead? if this is really needed, a comment should explain that this is a copy and what has changed.

pos_self_service_weight_base/static/src/js/Chrome.esm.js
```diff
@@ -0,0 +1,35 @@
+/** @odoo-module alias=pos_self_service_weight_base.Chrome **/
+// SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import Chrome from "point_of_sale.Chrome";
+import Registries from "point_of_sale.Registries";
+
+const SelfServiceChrome = (Chrome) =>
```
huguesdk:
```suggestion
const SelfServiceWeighingChrome = (Chrome) =>
```

pos_self_service_weight_base/static/src/js/ChromeWidgets/HeaderButton.esm.js
```diff
@@ -0,0 +1,39 @@
+/** @odoo-module alias=pos_self_service_weight_base.HeaderButton **/
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import HeaderButton from "point_of_sale.HeaderButton";
+import Registries from "point_of_sale.Registries";
+import {isConnectionError} from "point_of_sale.utils";
+
+const SelfServiceHeaderButton = (HeaderButton) =>
```
huguesdk:
```suggestion
const SelfServiceWeighingHeaderButton = (HeaderButton) =>
```

pos_self_service_weight_base/static/src/js/Popups/CloseSelfServicePopup.esm.js
```diff
@@ -0,0 +1,13 @@
+/** @odoo-module alias=pos_self_service_weight_base.CloseSelfServicePopup **/
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import ClosePosPopup from "point_of_sale.ClosePosPopup";
+import Registries from "point_of_sale.Registries";
+
+class CloseSelfServicePopup extends ClosePosPopup {}
```
huguesdk:
```suggestion
class CloseSelfServiceWeighingPopup extends ClosePosPopup {}
```

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/AbstractSelfServiceScreen.esm.js
```diff
@@ -0,0 +1,14 @@
+/** @odoo-module alias=pos_self_service_weight_base.AbstractSelfServiceScreen **/
+// SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import PosComponent from "point_of_sale.PosComponent";
+import Registries from "point_of_sale.Registries";
+
+class AbstractSelfServiceScreen extends PosComponent {}
```
huguesdk:
```suggestion
class AbstractSelfServiceWeighingScreen extends PosComponent {}
```

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/SelfServiceControlButton.esm.js
```diff
@@ -0,0 +1,22 @@
+/** @odoo-module alias=pos_self_service_weight_base.SelfServiceControlButton **/
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import PosComponent from "point_of_sale.PosComponent";
+import Registries from "point_of_sale.Registries";
+
+class SelfServiceControlButton extends PosComponent {
```
huguesdk:
```suggestion
class SelfServiceWeighingControlButton extends PosComponent {
```

pos_self_service_weight_base/static/src/js/Screens/SelfServiceScreen/SelfServiceHomeButton.esm.js
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import {Gui} from "point_of_sale.Gui";
+import Registries from "point_of_sale.Registries";
+import SelfServiceControlButton from "pos_self_service_weight_base.SelfServiceControlButton";
+import {useListener} from "@web/core/utils/hooks";
+
+class SelfServiceHomeButton extends SelfServiceControlButton {
```
huguesdk:
```suggestion
class SelfServiceWeighingHomeButton extends SelfServiceWeighingControlButton {
```

pos_self_service_weight_base/static/src/js/Screens/SelfServiceWelcomeScreen/SelfServiceActionButton.esm.js
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import {Gui} from "point_of_sale.Gui";
+import PosComponent from "point_of_sale.PosComponent";
+import Registries from "point_of_sale.Registries";
+import {useListener} from "@web/core/utils/hooks";
+
+class SelfServiceActionButton extends PosComponent {
```
huguesdk:
```suggestion
class SelfServiceWeighingActionButton extends PosComponent {
```

pos_self_service_weight_base/static/src/js/Screens/SelfServiceWelcomeScreen/SelfServiceWelcomeScreen.esm.js
```diff
@@ -0,0 +1,14 @@
+/** @odoo-module alias=pos_self_service_weight_base.SelfServiceWelcomeScreen **/
+// SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import PosComponent from "point_of_sale.PosComponent";
+import Registries from "point_of_sale.Registries";
+
+class SelfServiceWelcomeScreen extends PosComponent {}
```
huguesdk:
```suggestion
class SelfServiceWeighingWelcomeScreen extends PosComponent {}
```

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceScreen/SelfServiceControlButton.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<templates id="template" xml:space="preserve">
+
+    <t t-name="pos_self_service_weight_base.SelfServiceControlButton" owl="1">
```
huguesdk:
```suggestion
    <t t-name="pos_self_service_weight_base.SelfServiceWeighingControlButton" owl="1">
```

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceScreen/SelfServiceScreen.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<templates id="template" xml:space="preserve">
+
+    <t t-name="pos_self_service_weight_base.AbstractSelfServiceScreen" owl="1">
```
huguesdk:
```suggestion
    <t t-name="pos_self_service_weight_base.AbstractSelfServiceWeighingScreen" owl="1">
```

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceWeightScreen/SelfServiceActionButton.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<templates id="template" xml:space="preserve">
+
+    <t t-name="pos_self_service_weight_base.SelfServiceActionButton" owl="1">
```
huguesdk:
```suggestion
    <t t-name="pos_self_service_weight_base.SelfServiceWeighingActionButton" owl="1">
```

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceWeightScreen/SelfServiceWelcomeScreen.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<templates id="template" xml:space="preserve">
+    <t t-name="pos_self_service_weight_base.SelfServiceWelcomeScreen" owl="1">
```
huguesdk:
```suggestion
    <t t-name="pos_self_service_weight_base.SelfServiceWeighingWelcomeScreen" owl="1">
```

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceWeightScreen/SelfServiceWelcomeScreen.xml
```diff
@@ -0,0 +1,16 @@
...
+        <div class="screen">
+            <div class="screen-full-width">
+                    <div class="self-service-action-buttons">
+                        <SelfServiceActionButton t-if="env.isDebug()" />
```
huguesdk: why is this button shown only in debug mode?

huguesdk (changes requested):
@robinkeunen thanks for all the improvements!

some remarks:

*   the red “unknown error” dialog box still appears.
*   the top bar is now visible again in the normal pos, but in self-service weighing mode as well, while it should be hidden, as the customer shouldn’t be able to easily exit the interface.
*   previously, the module was called `pos_self_service_base`, without a reference to weighing, while here it is specifically about weighing, so i suggest to rename all classes (and files) from `*SelfService*` to `*SelfServiceWeighing*`.
*   (not that important) it is possible to create a self-service weight station without checking the “electronic scale” checkbox is the iot box section. having something checking this would be a plus (only if it is quick and easy).

pos_self_service_weight_base/static/src/css/pos.css
robinkeunen: I did not look into it at all. Writing it down for later.

pos_self_service_weight_base/static/src/js/Chrome.esm.js
robinkeunen: it is needed if we want to allow other modules to inherit this one using the old Odoo _define_ syntax. so not strictly needed.

i also prefer importing module names than file names but that's maybe me coming from python and lack consistency.

do you think I should change it ?

huguesdk: since this module is extending classes that still use the old `define` syntax, i think t makes sense to keep it.

pos_self_service_weight_base/static/src/js/Chrome.esm.js
robinkeunen: done

pos_self_service_weight_base/models/pos_config.py
robinkeunen: done

pos_self_service_weight_base/views/pos_config_views.xml
robinkeunen: done

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceWeightScreen/SelfServiceWelcomeScreen.xml
robinkeunen: It exists to test the behavior of the base module when there are nos other modules installed. It can be removed in the future.

robinkeunen:
@huguesdk I restored the top bar because 

- i wanted to exit cleanly the interface
- i find it useful to have the information of the connections and the logged user
- i find the look more consistent with the other point of sale
- (i read the specification too fast)

To mitigate the risk of closing the point of sale by accident, I added the confirmation popup (suggestion from Rémy).

I tend to agree that the risk to have customer closing the weighing station is high so, keeping in mind the add benefits of the bar, do you agree to rather hide the "close" button and remove the popup altogether.

@victor-champonnois fyi

pos_self_service_weighing_base/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Coop IT Easy SC, Odoo Community Association (OCA) <https://github.com/OCA/pos>`_:
+
+  * Carmen Bianca BAKKER
```
carmenbianca, line 3: Add Robin Keunen

pos_self_service_weighing_base/static/src/xml/Chrome.xml
```diff
@@ -0,0 +1,21 @@
...
+        t-name="Chrome"
+        t-inherit="point_of_sale.Chrome"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//TicketButton" position="attributes">
+            <attribute name="t-if">showTicketButton()</attribute>
+        </xpath>
+    </t>
+
```
carmenbianca, line 23: There used to be this code here:

```xml
        <xpath expr="//div[hasclass('pos-topheader')]" position="attributes">
            <attribute
                name="t-att-class"
            >{ oe_hidden: state.uiState !== 'READY' || env.pos.config.iface_self_service }</attribute>
        </xpath>

        <xpath expr="//div[hasclass('pos-content')]" position="attributes">
            <t t-if="env.pos.config.ifaceself_service">
                <attribute name="class">pos-content positionzero</attribute>
            </t>
        </xpath>
```

This removed the bar at the top. Now, a user can click on 'close', and be sent back to the Odoo interface. This is bad for two reasons:

- Depending on how Odoo is configured, this allows the user access to things they shouldn't be able to see.
- It may be difficult/time-consuming to recover from such an error.

'Close' is also an incredibly normal button to press, so this may be very frequent.

pos_self_service_weighing_base/static/src/js/ChromeWidgets/HeaderButton.esm.js
```diff
@@ -0,0 +1,42 @@
...
+                        body: this.env._t(
+                            "An unknown error prevents us from getting closing information."
+                        ),
+                    });
+                }
+            }
+        }
+    };
+Registries.Component.extend(HeaderButton, SelfServiceWeighingHeaderButton);
+export default SelfServiceWeighingHeaderButton;
```
carmenbianca: Oof… This is 99% upstream code, except uses a custom implementation of the closing pop-up. That isn't great. Can we not edit the XML/template of `ClosePosPopup` to display something completely different if the POS is a weighing station?

Although, as mentioned in a previous comment, I'm not sure we _should_ be able to close from this screen.

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingControlButton.esm.js
```diff
@@ -0,0 +1,35 @@
...
+        super.setup();
+        useListener("click", this._onClick);
+    }
+
+    async _onClick() {
+        await this.showPopup("ErrorPopup", {
+            title: "Error",
+            body: "_onclick Not Implemented",
+        });
+    }
```
carmenbianca: This should probably be a public method.

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWelcomeScreen/SelfServiceWeighingActionButton.esm.js
```diff
@@ -0,0 +1,33 @@
...
+import {useListener} from "@web/core/utils/hooks";
+
+class SelfServiceWeighingActionButton extends PosComponent {
+    get name() {
+        return "Base Action";
+    }
+
+    get faSymbol() {
+        return "fa-barcode";
+    }
```
carmenbianca: What is this class? This seems very similar to `SelfServiceWeighingControlButton`, but with some strange defaults?

pos_self_service_weighing_base/static/src/xml/Screens/SelfServiceWeighingWelcomeScreen/SelfServiceWeighingWelcomeScreen.xml
```diff
@@ -0,0 +1,16 @@
...
+<!--
+SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<templates id="template" xml:space="preserve">
+    <t t-name="pos_self_service_weighing_base.SelfServiceWeighingWelcomeScreen" owl="1">
+        <div class="screen">
+            <div class="screen-full-width">
+                    <div class="self-service-action-buttons">
+                        <SelfServiceWeighingActionButton t-if="env.isDebug()" />
```
carmenbianca: Really nice!

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWelcomeScreen/SelfServiceWeighingActionButton.esm.js
carmenbianca: I think I've worked this out. It's the button on the welcome screen. The defaults are there because there's a button in the debug mode?

I think a few things need to happen to improve this:

1. Rename `SelfServiceWeighingActionButton` to `SelfServiceWeighingWelcomeButton`. This is much easier to disambiguate from `SelfServiceWeighingControlButton`.
2. Remove all the default values from `SelfServiceWeighingWelcomeButton`.
3. Create `SelfServiceWeighingDebugButton` with the removed defaults.

pos_self_service_weighing_tare/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Coop IT Easy SC, Odoo Community Association (OCA) <https://github.com/OCA/pos>`_:
+
+  * Carmen Bianca BAKKER
```
carmenbianca, line 3: add robin keunen

pos_self_service_weighing_tare/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Configure a PoS to be a self-service station where customers can weigh things. As a base module, this covers primarily the welcome screen and some common functionalities.
```
carmenbianca: This is incorrect

pos_self_service_weighing_tare/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingPrintTareButton.esm.js
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import Registries from "point_of_sale.Registries";
+import SelfServiceWeighingControlButton from "pos_self_service_weighing_base.SelfServiceWeighingControlButton";
+
+class SelfServiceWeighingPrintTareButton extends SelfServiceWeighingControlButton {
+    get name() {
+        return "Print Tare Label";
```
carmenbianca: This needs to be translatable.

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingHomeButton.esm.js
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import {Gui} from "point_of_sale.Gui";
+import Registries from "point_of_sale.Registries";
+import SelfServiceWeighingControlButton from "pos_self_service_weighing_base.SelfServiceWeighingControlButton";
+import {useListener} from "@web/core/utils/hooks";
+
+class SelfServiceWeighingHomeButton extends SelfServiceWeighingControlButton {
+    get name() {
+        return "Home";
```
carmenbianca: This needs to be translatable.

pos_self_service_weighing_tare/static/src/xml/Screens/SelfServiceWeighingTareScreen/SelfServiceWeighingPrintTareButton.xml
```diff
@@ -0,0 +1,16 @@
...
+<templates id="template" xml:space="preserve">
+
+    <t t-name="pos_self_service_weighing_tare.SelfServiceWeighingPrintTareButton" owl="1">
+        <button class="button">
+            <i t-attf-class="fa {{ faSymbol }}" />
+            <t t-esc="name" />
+        </button>
+    </t>
+
+</templates>
```
carmenbianca: This file is unnecessary.

pos_self_service_weighing_tare/static/src/js/Screens/SelfServiceWeighingWelcomeScreen/SelfServiceWeighingTareScreenButton.esm.js
```diff
@@ -0,0 +1,27 @@
...
+        return "fa-barcode";
+    }
+
+    async _onClick() {
+        Gui.showScreen("SelfServiceWeighingTareScreen");
+    }
+}
+
+SelfServiceWeighingTareScreenButton.template =
+    "pos_self_service_weighing_tare.SelfServiceWeighingTareScreenButton";
```
carmenbianca: You can set the template to `pos_self_service_weighing_base.SelfServiceWeighingActionButton`.

pos_self_service_weighing_base/static/src/xml/Screens/SelfServiceWeighingScreen/SelfServiceWeighingScreen.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+SPDX-FileCopyrightText: 2023 Coop IT Easy SC
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<templates id="template" xml:space="preserve">
+
+    <t
+        t-name="pos_self_service_weighing_base.AbstractSelfServiceWeighingScreen"
```
carmenbianca: Name does not match file name.

pos_self_service_weighing_tare/static/src/xml/Screens/SelfServiceWeighingTareScreen/SelfServiceWeighingTareScreen.xml
```diff
@@ -0,0 +1,27 @@
...
+                    </div>
+                </div>
+                <div class="rightpane">
+                    <div class="self-service-action-buttons">
+                        <SelfServiceWeighingPrintTareButton />
+                    </div>
+                </div>
+            </div>
+        </div>
+    </t>
```
carmenbianca, line 25: This completely re-implements `pos_self_service_weighing_base.AbstractSelfServiceWeighingScreen`, which I'm not sure is intentional. Can you inherit the latter template without changing it for other inheritors?

pos_self_service_weighing_tare/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingTareScreen.esm.js
```diff
@@ -0,0 +1,14 @@
+/** @odoo-module alias=pos_self_service_weighing_tare.SelfServiceWeighingTareScreen **/
+// SPDX-FileCopyrightText: 2024 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import PosComponent from "point_of_sale.PosComponent";
+import Registries from "point_of_sale.Registries";
+
+class SelfServiceWeighingTareScreen extends PosComponent {}
```
carmenbianca: Why does this not extend `AbstractSelfServiceWeighingScreen`?

pos_self_service_weighing_tare/static/src/xml/Screens/SelfServiceWeighingWelcomeScreen/SelfServiceWeighingTareScreenButton.xml
```diff
@@ -0,0 +1,19 @@
...
+
+    <t
+        t-name="pos_self_service_weighing_tare.SelfServiceWeighingTareScreenButton"
+        owl="1"
+    >
+        <button class="button">
+            <t t-esc="name" />
+            <i t-attf-class="fa {{ faSymbol }}" />
+        </button>
+    </t>
```
carmenbianca: This file is not needed.

pos_self_service_weighing_tare/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingPrintTareButton.esm.js
```diff
@@ -0,0 +1,30 @@
...
+        const barcode = "1234567890";
+        this.showPopup("SelfServiceWeighingBarcodePopup", {
+            barcode: barcode,
+            keepBehind: true,
+        });
+    }
+}
+
+SelfServiceWeighingPrintTareButton.template =
+    "pos_self_service_weighing_tare.SelfServiceWeighingPrintTareButton";
```
carmenbianca: You can set the template to `pos_self_service_weighing_base.SelfServiceWeighingControlButton`.

carmenbianca (commented):
Quite a big review. Some of the things in the review are my own bad/questionable decisions.

pos_self_service_weighing_base/readme/CONTRIBUTORS.rst
robinkeunen: done

pos_self_service_weighing_tare/readme/CONTRIBUTORS.rst
robinkeunen: done

pos_self_service_weighing_tare/readme/DESCRIPTION.rst
robinkeunen: done

pos_self_service_weighing_base/static/src/xml/Screens/SelfServiceWeighingScreen/SelfServiceWeighingScreen.xml
robinkeunen: done

pos_self_service_weighing_base/static/src/xml/Chrome.xml
carmenbianca: I talked with @robinkeunen after forgetting to read his comment from earlier today → hiding the close button would also be a satisfactory solution. This would also allow getting rid of some code elsewhere related to the close pop-up.

pos_self_service_weighing_tare/static/src/xml/Screens/SelfServiceWeighingTareScreen/SelfServiceWeighingTareScreen.xml
carmenbianca: Talked with @robinkeunen about this. We can do primary inheritance here as described in my previous comment, _but_ there's a chance that the primary inheritance would be more lines of code in a much less transparent way. e.g. insert a button here, insert a button there, etc.

For developer ease, it may be nice to just see the simple hierarchy like in the quoted code. The hierarchy is sufficiently small that the repetition is barely a problem. And there would be repetition, because there will be other screens that re-implement `pos_self_service_weighing_base.AbstractSelfServiceWeighingScreen`. Furthermore, although we _currently_ think that other implementations of `AbstractSelfServiceWeighingScreen` will look incredibly similar, this may not always be the case.

But… For the time being, it's still repetition. And because this hierarchy kind of depends on `pos.css` from the main module to display well, it's probably better to do primary inheritance?

Your thoughts @huguesdk ?

robinkeunen: @huguesdk we discussed this with Carmen. It is now obvious to us wether primary inheritance is better (less code - maybe) or worse (lacks clarity) then redefining a template. What is your point of view ?

pos_self_service_weighing_tare/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingTareScreen.esm.js
robinkeunen: done

pos_self_service_weight_base/static/src/xml/Screens/SelfServiceWeightScreen/SelfServiceWelcomeScreen.xml
robinkeunen: done for all

pos_self_service_weighing_base/static/src/xml/Chrome.xml
robinkeunen: done

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWelcomeScreen/SelfServiceWeighingActionButton.esm.js
robinkeunen: done

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingControlButton.esm.js
robinkeunen: I'm not sure I understand, do you mean exporting this method ? It is accessible from child classes inheriting it :thinking:

pos_self_service_weighing_tare/static/src/xml/Screens/SelfServiceWeighingWelcomeScreen/SelfServiceWeighingTareScreenButton.xml
robinkeunen: done (and other similar)

pos_self_service_weighing_base/static/src/js/Screens/SelfServiceWeighingScreen/SelfServiceWeighingControlButton.esm.js
carmenbianca: Removing the `_` prefix :)

robinkeunen: aaah, that I can do !

PR #1161, opened 2024-02-29
[16.0][MIG] pos_payment_method_cashdro: migration to 16.0

chienandalu:
@Tecnativa TT45180

please review @CarlosRoca13 @victoralmau 

It can be tested with this mock backend: drde5.wiremockapi.cloud


pos_payment_method_cashdro/models/pos_payment_method.py
```diff
@@ -13,10 +13,11 @@ def _get_payment_terminal_selection(self):
...
+    cashdro_user = fields.Char()
+    cashdro_password = fields.Char()
 
-    def _onchange_is_cash_count(self):
+    def _onchange_journal_id(self):
+        res = super()._onchange_journal_id()
         if self.use_payment_terminal == "cashdro":
-            return
-        return super()._onchange_is_cash_count()
+            self.use_payment_terminal = True
```
pedrobaeza: Don't get the intention of this changing the value from `cashdro` to `True`, and there's no docstring.

chienandalu: It wasn't right indeed. Fixed and added docstring

pos_payment_method_cashdro/models/pos_payment_method.py
```diff
@@ -13,10 +13,13 @@ def _get_payment_terminal_selection(self):
...
-    def _onchange_is_cash_count(self):
-        if self.use_payment_terminal == "cashdro":
-            return
-        return super()._onchange_is_cash_count()
+    def _onchange_journal_id(self):
+        """Cash payment method force the `use_payment_terminal` to `False` as
+        it's assumed that a cash journal can't have a payment terminal"""
+        if self.use_payment_terminal != "cashdro":
+            return super()._onchange_journal_id()
+        self.use_payment_terminal = "cashdro"
```
pedrobaeza, line 26: Shouldn't this be done only for cash journals? Or let decide the user and do nothing

chienandalu: I just improved the usability. This is done to avoid the change of the journal drop the terminal config.

CarlosRoca13 (commented):
Thanks! :+1:

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #1180, opened 2024-04-22
[16.0] pos_membership_extension : change membership category

legalsylvain:
**Use case :** 
- select a customer that belong to a membership category
- select then MANY products that are reserved to this membership category
- switch then to a "normal" customer that doesn't belong to the membership category

**Before :** 

- Not all lines are removed
- the customer is not changed.

**After**
- the customer is correctly changed
- all the lines are correctly removed.

Ref : https://stackoverflow.com/a/9882349


ivantodorovich (approved):
LG

legalsylvain (commented):
/ocabot merge patch

PR #1184, opened 2024-04-29
[17.0] copier repo update

trisdoan:


.copier-answers.yml
```diff
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.19
+_commit: v1.20-4-g56d206b
```
legalsylvain: Why it is not just 1.20 here ?

trisdoan: Hello @legalsylvain, I updated as you suggest :pray:

legalsylvain (approved):
/ocabot merge nobump

.copier-answers.yml
nilshamerlinck: Hi @legalsylvain it can happen that some changes have been merged in template while a new version has not been released; that was the case here, updating repo with template v1.20, but commit 56d206b
anyway this PR can now be closed as the repo as been [mass-updated meanwhile for another reason](https://github.com/OCA/oca-addons-repo-template/issues/250) :)

PR #1194, opened 2024-05-08
[16.0] [ADD] pos_product_available - Available products for each POS

cfrancoae:


cfrancoae:
In the product configuration, a field has been added where it can be assigned to the point of sale where its sale is allowed.

This functionality serves to prevent all products with the configuration available in the POS from appearing in all points of sale.

PR #1198, opened 2024-05-22
[17.0][ADD] Module pos_order_split_invoice_attachment_sync

cvinh:
This module synchronizes attachments between pos orders and splitted invoices
It depends on pos_order_split_invoice and pos_order_attachment


cvinh:
@etobella could you review please ?

PR #1206, opened 2024-06-19
[14.0][MIG] pos_tare: Migration to 14.0

robinkeunen:
superseed #980 integrating @legalsylvain 's comment.


ivantodorovich (approved):

legalsylvain (commented):
Hi. just tested. a lot of things working well. Thanks ! 
however, I have a question regarding some configuration. If I set such configuration :
image

I can not set the gross weight manually.

Can you reproduce the problem ? 

image

robinkeunen:
Yes, I get it too. The overload of ScaleScreen does not work in that case. I'll focus on the v16 for the time being as we won't use this version...

PR #1209, opened 2024-06-29
[16.0][ADD] new module pos_check_accounting_settings to avoid launching new PoS session with incorrect configuration.

legalsylvain:
This module extends the functionnality of Odoo PoS adding extra checks, when opening session, to ensure that accounting configation is correct.

### Use case 1

When we create a payment method and if :
- the field 'Outstanding Payment' is not set on payment method
- AND the field 'Outstanding Receipts Account' is not on company

if we use the payment method in the Point of sale, it is not possible to close the session after.

error_oustanding_payment

As it not possible to change a payment method when a session is not closed, this bad configuration has to be fixed by SQL.

This module adds a check when opening session, to avoid such situation.


miquelalzanillas:
@legalsylvain could you rebase please?

flotho:
any chance to rebase @legalsylvain  ?

legalsylvain:
@miquelalzanillas, @flotho : done !

quentinDupont:
Very useful module.
We could add the check for `account_default_pos_receivable_account_id` which causes the same problem

yakafodon as we say in french

quentinDupont:
And also `default_cash_payment_method` if someone has the bad idea to remove Cash journal in Payment Methods

legalsylvain:
superseded by https://github.com/OCA/pos/pull/1575

PR #1158, opened 2024-02-20
[16.0][ADD] pos_screen_element_custom_size

geomer198:
This module allows to define custom size values for certain POS screen elements.

PR #1211, opened 2024-07-01
[16.0] [IMP] pos_order_to_sale_order: use test helpers

ivantodorovich:
Hi @legalsylvain 

This should fix the issue described here: https://github.com/OCA/pos/pull/1158#issuecomment-2152750510

It's a race condition when closing the POS on tours.. the tour finished way too quickly some times, not giving time for the backend request to complete and commit the changes.

I re-wrote the test tours to use the "new" helpers, which is a nice, and added a specific step to wait for the popup to close right after executing the rpc request.


ivantodorovich:
Okay, switching back to draft because there's the error again on the CI, which wasn't there on my local
Sorry for the noise

At least this time it seems to be consistent, probably a compatibility issue with some other module.
Investigating...

legalsylvain (commented):
LGTM. Nice simplification ! and thanks for the investigation ! 

Let's see if it is solving the random CI problem ! 

/ocabot merge patch

legalsylvain:
11 hours later ...

/ocabot merge patch

legalsylvain:
@sbidoul : is the bot down ? 

(not sure this is you to be pinged ...)

sbidoul:
> is the bot down ?

Yeah, I got the alert since a few days that the bot is consuming a lot of RAM. That's something new... sigh.

It's slowly going through the queue, at the moment.

legalsylvain:
> Yeah, I got the alert since a few days that the bot is consuming a lot of RAM. That's something new... sigh.

Hum annoying. Let us know if we have to organize some code sprint on that topic in the next OCA days.

regards.

PR #1210, opened 2024-07-01
[ADD] pos_receipt_usability : new module to improve the display of the receipt screen

legalsylvain:
### pos_receipt_usability

This module extends the functionality of point of sale module, improving the display of the receipt screen.

**Without the module**

without_module

**With the module**

with_module

PR #1212, opened 2024-07-02
16.0 add pos_restaurant_receipt_usability

legalsylvain:
depends on https://github.com/OCA/pos/pull/1210

This module extends the functionality of Point of Sale Restaurant module,
improving the display of the receipt screen, in a "Split receipt" use case.

By default, at the end of a split-ticket cashing, on the receipt page,
the user only has the option of clicking on "Next order",
which creates a new blank order.

In this use case, it is preferable to return to the screen which allows
to continue "splitting" the same ticket, until all the order is paid.

**Without the module**

without_module

**With the module**

with_module


quentinDupont (approved):
Tested locally, works like a charm :heavy_check_mark: 
Great module to avoid click, should be native !  💫

PR #1214, opened 2024-07-02
[ADD] New module pos_receipt_vat_detail to display on receipt, the detail of the taxes. (Tax amount and base amount).

legalsylvain:
This module extends the Odoo Point of Sale module to explicit the fiscal information present in the Point of sale Receipt.

By default, in odoo, there is not the base amount of the vat displayed on the receipt.

without_module

With this module, the base amount is displayed. It is usefull, for exemple, if receipt is asked by customer to make an HR expense because such details can be asked by the accountant.

with_module

The informations are similar to the informations present in the invoice report.


legalsylvain:
used in production since 6 monthes.

/ocabot merge nobump

PR #1215, opened 2024-07-03
[14.0][FIX] pos_sale_order_load: Fix ValueError Invalid field 'pos_allow_negative_qty'

carolinafernandez-tecnativa:
- Field pos_allow_negative_qty is set in pos_order_return and this module not depends on it. 
- Tests are failing on environments without pos_order_return installed. 
- Fix ValueError: Invalid field 'pos_allow_negative_qty' on model 'product.product'

@Tecnativa
TT49941
@pedrobaeza @victoralmau


legalsylvain:
thanks for this patch.

/ocabot merge patch

PR #1217, opened 2024-07-09
[16.0][ADD] pos_loyalty_exclude

trisdoan:


cyrilmanuel (approved):

pos_loyalty_exclude/models/loyalty_rule.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+        # this check ensures that it only runs when there are loyalty-excluded products
+        if excluded_products:
+            for rule in self:
+                # exclude product when no config in loyalty rule
+                if not (
+                    rule.product_ids
+                    or rule.product_category_id
+                    or rule.product_tag_id
+                    or rule.product_domain not in ("[]", "[['sale_ok', '=', True]]")
```
Ricardoalso, line 25: is [['sale_ok', '=', True]] intended or do you meant [('sale_ok', '=', True)]

Ricardoalso (approved):

pos_loyalty_exclude/models/loyalty_rule.py
```diff
@@ -0,0 +1,36 @@
...
+                ):
+                    rule.any_product = False
+                    rule.valid_product_ids = self.env["product.product"].search(
+                        [("loyalty_exclude", "=", False)]
+                    )
```
nilshamerlinck:
```suggestion
                    rule.valid_product_ids = self.env["product.product"].search(
                        [("loyalty_exclude", "=", False), ("available_in_pos", "=", True)], order="id"
                    )
```

pos_loyalty_exclude/models/loyalty_rule.py
trisdoan: Hello @Ricardoalso, it is intended as it is in https://github.com/odoo/odoo/blob/def7c93c6c05e1814d6221cf90e93bdd8867e473/addons/pos_loyalty/models/loyalty_rule.py#L29

simahawk:
https://github.com/OCA/pos/issues/1236

PR #1237, opened 2024-09-24
[FIX][14.0] get_purchase_price pos_margin

sergiocorato:


pos_margin/models/pos_order_line.py
```diff
@@ -37,12 +37,12 @@ def _get_purchase_price(self, line):
         """Overwrite the method from the module "sale_margin"
         (https://github.com/odoo/odoo/blob/14.0/addons/sale_margin/models/sale_order.py#L20)"""
         if not line.product_id:
-            line.purchase_price = 0.0
+            return 0.0
```
legalsylvain, line 40: I don't get it. It looks that the comment is obsolete.
https://github.com/odoo/odoo/blob/14.0/addons/sale_margin/models/sale_order.py#L20 doens't mention a _get_purchase_price function, but a _compute_purchase_price function.

It should be bette "copy the method from sale_margin module". Don't you think ? 

otherwise, LGTM.

PR #1238, opened 2024-09-24
[ADD][16.0] pos_default_quantity

carmenbianca:
Migrated from v12 @ https://github.com/coopiteasy/cie-pos/tree/12.0/pos_default_quantity

This module adds a field for a default quantity to Unit of Measure Categories.
When this is filled in, and when this functionality is enabled in the Point of
Sale configuration, then adding that product to an order in the Point of Sale
will set the quantity to that value by default.

Internal task T12149


legalsylvain (commented):
Hi @carmenbianca . Thanks for sharing.

- could you provide a use case, when this module is interesting for end user. I'm not sure to understand.
- naive question : when you port a module from 12 to 16, why do you make migration to 13 / 14 /15 and not directly to 16 ? 

thanks !

victor-champonnois:
@legalsylvain Regarding the use case, it's an old module do I don't know the original intention. From what supermarkets told me, they use it on weighted products by setting a very high default minimal amount. 
This prevents the cashier from forgetting to put a weight on the product (because the total of the order is then very high if the weight is not manually set). 

I don't know if it's the original intent of the module but some supermarkets are attached to it. I am sure there are better way to prevent forgetting to put the weight though.

legalsylvain:
> I don't know if it's the original intent of the module but some supermarkets are attached to it. I am sure there are better way to prevent forgetting to put the weight though.

;-)

victor-champonnois:
@carmenbianca does the module works on your side ? I tested it on products with a "kg" and a "g" uom but the default quantity is not set on the POS.

carmenbianca:
@victor-champonnois yes the module works for me. see https://github.com/OCA/pos/blob/f9c012809be12e1948c060a89e0bf92fdb0b648c/pos_default_quantity/readme/CONFIGURE.rst for the steps needed to make it work.

carmenbianca:
@legalsylvain 

> * could you provide a use case, when this module is interesting for end user. I'm not sure to understand.

the use-case is a little beyond me. doing this by UoM category instead of, say, product category is a strange choice, but I just ported it as-is. I could 100% understand keeping this as an internal module. I'm just trying to lower the amount of internal modules where I can to be nicer members of the community.

> * naive question : when you port a module from 12 to 16, why do you make migration to 13 / 14 /15 and not directly to 16 ?

[Those are the OCA contribution guidelines](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0). Now technically I didn't do a good job of this, and this wasn't an OCA module to begin with, so that history wasn't super important.

victor-champonnois:
@carmenbianca I test on the runboat, following the steps in the documented configuration, but I don't see it working in the POS. Could you have a look at the runboat ? (I test with the product called "kg product", with the "weight" uom category).

carmenbianca:
@victor-champonnois you're right, it doesn't work (any longer?). i was also able to reproduce the dysfunction locally.

also, apparently the 'technical' section is hidden unless you use debugging mode. that isn't great; to fix.

carmenbianca:
@victor-champonnois fixed. there was a typo in the javascript that mustn't have been there before.

i also moved the setting to the 'inventory' section. not a perfect match, but it's the best match without creating a new section.

victor-champonnois (approved):
Thank you for the fix, LGTM functionnally!

victor-champonnois:
@carmenbianca I am not sure it's a good OCA fit though, since the known use case is as bit of a hack.

carmenbianca:
@victor-champonnois agreed, closing in favour of https://github.com/coopiteasy/cie-pos/pull/27

PR #1242, opened 2024-09-24
[FIX][16.0] pos_lot_selection rounding in lots search

sergiocorato:
Simple fix for a method call bugged.


sergiocorato:
I don't think it's related to this PR errors like this one: https://github.com/OCA/pos/actions/runs/11027459923/job/30625801358?pr=1242#step:8:356

PR #1243, opened 2024-09-25
[16.0][ADD] pos_partner_pricelist_load_background

santostelmo:
This module allow you to load a pricelist for a customer in the background.

**Context**
In the POS, we can configure a list of available pricelists. These price lists are loaded during start-up and only these can be chosen for any order.
When a customer is selected, the POS will try to find the customer's pricelist (property_pricelist_id) among the available pricelists. 
If it's found, the pricelist is selected otherwise a default one will be selected instead.

**With this module**
When a customer is selected, their pricelist (property_pricelist_id) is loaded and available to be chosen.


cyrilmanuel (approved):
not an expert but LGTM

pos_partner_pricelist_load_background/models/pos_session.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2024 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import models
+
+
+class POSSession(models.Model):
+    _inherit = "pos.session"
+
+    def get_pos_ui_partner_pricelist_background(self, pricelist_id, product_ids):
```
simahawk, line 9: would be nice to see a docstring and a test for this

santostelmo: Added doc string and unit test.

gurneyalex (approved):

cyrilmanuel (approved):

cyrilmanuel:
hi @OCA/pos-maintainers  could we merge this PR please :)

PR #1245, opened 2024-09-26
[FIX] pos_product_label: Avoid to show extra content field in different formats at PoS

miguelcb2003:
In backend, this field dissapear automatically when you click any other format that is not '2x7xprice'. This is because it's the only format that allows to have extra content in its label. This is a video that shows this functionallity at backend:

[backend.webm](https://github.com/user-attachments/assets/4d461952-a928-4231-844a-49faf3cc050e)

Before this changes, that behaviour didn't work at PoS, making this field editable in all formats.

[pos.webm](https://github.com/user-attachments/assets/d664a058-3fab-4406-a6b6-236f1185e516)

Now, with this changes the behaviour at PoS is the same than in backend, avoiding to modify the extra content field when it's not possible.

FL-556-3627

### DIFF
--- a/pos_product_label/static/src/PrintLabelPopup.xml
+++ b/pos_product_label/static/src/PrintLabelPopup.xml
@@ -35,6 +35,7 @@
                                         t-model="data.format"
                                         t-att-id="format"
                                         t-att-value="format"
+                                        t-on-change="updateFormat"
                                     />
                                     <label
                                         t-att-for="format"
@@ -45,7 +46,7 @@
                         </div>
                     </section>
                     <section>
-                        <div>
+                        <div t-if="data.format == '2x7xprice'">
                             <label for="extraContent">Extra Content</label>
                             <textarea
                                 rows="4"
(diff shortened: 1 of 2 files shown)


Pablocce (approved):
LGTM

pos_product_label/static/src/PrintLabelPopup.xml
```diff
@@ -35,6 +35,7 @@
                                         t-model="data.format"
                                         t-att-id="format"
                                         t-att-value="format"
+                                        t-on-change="updateFormat"
```
ivantodorovich, line 38: are you sure this is needed? I thought it'd be handled by the widget through the `t-model="data.format"` attribute

Alber1998 (approved):
LGTM

miguelcb2003:
Hi @ivantodorovich ! I add that because at PoS, this value does not update until confirm the action. If we want to hide or show the extra content dinamically we need the onchange function.
Anyway, if you have any suggest to improve this let me know and I'll implement it. Thank you for the validation!

ivantodorovich (approved):

ivantodorovich:
/ocabot merge patch

thanks!

PR #1041, opened 2023-08-13
[16.0][FIX] pos_stock_available_online: FIX For After refund of 1 qty, POS qty display increased by 2 qty

kothiyarajesh:
[After refund of 1 qty, POS qty display increased by 2 qty](https://github.com/OCA/pos/issues/1039l)#1039
Found same issue on refund time
Here added few fix for that, It's work for me please review @GabbasovDinar @chienandalu

Issue #1039, opened 2023-08-09
After refund of 1 qty, POS qty display increased by 2 qty


arjungitub:
## Module
The name of the module that has a bug.
pos_stock_available_online
stock_available

## Describe the bug
When we are refunding 1 qty on the POS machine, qty of product is increased by 2. Please check this error.

## To Reproduce
I have tried on odoo16 only.

1. Sell a product in POS
2. Refund it.
3. Then check the qty displayed on POS

**Expected behavior**
Exact qty should be added after refund

chienandalu:
@GabbasovDinar @francesco-ooops Maybe you're interested in checking this

ivs-cetmix:
Hi @arjungitub , thank you for reporting! Will check this

francesco-ooops:
@chienandalu @isserver1 I need to ask for confirmation to @GSLabIt who analyzed the issue internally a few months ago, but as far as I remember the issue was related to standard odoo pos return/refund feature (this info has to be verified)

PR #1250, opened 2024-10-01
[16.0][FIX] pos_stock_available_online: stock move updates

danielduqma:
Make stock updates not only on quant updates, but also when a stock move changes its state.

This fixes two bugs:

- Stock is not updated when a new incoming move is ready, or when a ready one is cancelled. This triggers notification in both outgoing and incoming warehouses

- Stock is decreased twice because when quant is updated, move is taken in account because it is not yet on 'done' state. This was reported in https://github.com/OCA/pos/issues/1039, but the proposed solution in https://github.com/OCA/pos/pull/1041 was local and only concerned moves from PoS orders

FL-556-4139

### DIFF
--- a/pos_stock_available_online/models/stock_notifier_pos_mixin.py
+++ b/pos_stock_available_online/models/stock_notifier_pos_mixin.py
@@ -1,0 +1,51 @@
+from odoo import models
+
+
+class StockNotifierPosMixin(models.AbstractModel):
+    _name = "stock.notifier.pos.mixin"
+    _description = "Stock Notifier POS Mixin"
+
+    def _prepare_pos_message(self):
+        """
+        Return prepared message to send to POS
+        """
+        self.ensure_one()
+        return self.warehouse_id._prepare_vals_for_pos(self.product_id)
+
+    def _skip_notify_pos(self):
+        """
+        Skip notification to POS
+        """
+        return False
+
+    def _get_warehouses_to_notify(self):
+        self.ensure_one()
+        return self.env["stock.warehouse"]
+
+    def _notify_pos(self):
+        """
+        Send notification to POS
+        """
+        pos_session_obj = self.env["pos.session"]
+        for record in self:
+            if record._skip_notify_pos():
+                continue
+            for warehouse in self._get_warehouses_to_notify():
+                configs = pos_session_obj.search(
+                    [
+                        ("state", "=", "opened"),
+                        ("config_id.display_product_quantity", "=", True),
+                        "|",
+                        ("config_id.additional_warehouse_ids", "in", [warehouse.id]),
+                        ("config_id.main_warehouse_id", "=", warehouse.id),
+                        "|",
+                        ("config_id.iface_available_categ_ids", "=", False),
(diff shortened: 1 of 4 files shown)


miguelcb2003 (approved):
LGTM!

Pablocce (approved):
LGTM

pos_stock_available_online/models/stock_notifier_pos_mixin.py
```diff
@@ -0,0 +1,54 @@
+from odoo import models
+
+
+class StockNotifierPosMixin(models.AbstractModel):
+    _name = "stock.notifier.pos.mixin"
+    _description = "Stock Notifier POS Mixin"
+
+    def _prepare_pos_message(self, warehouse=None):
```
GabbasovDinar: The `_prepare_pos_message` is called directly in the warehouse loop in the `_notify_pos` method, which gets the list of warehouses using the `_get_warehouses_to_notify` method in this case warehouse will always be in the argument values, so it can be left as a mandatory argument.

```python
def _prepare_pos_message(self, warehouse):
     """
     Return prepared message to send to POS
     """
     self.ensure_one()
     return warehouse._prepare_vals_for_pos(self.product_id)
```

pos_stock_available_online/models/stock_notifier_pos_mixin.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def _skip_notify_pos(self):
+        """
+        Skip notification to POS
+        """
+        return False
+
+    def _get_warehouses_to_notify(self):
+        self.ensure_one()
+        return self.warehouse_id
```
GabbasovDinar: Since the current model is a mixin and can be plugged into any model this field may be missing which will cause an error, possible solution is to either return the model object `return self.env["stock.warehouse"]` or throw an exception `raise NotImplementedError()`

In the models where this mixin is connected, implement a method that returns the necessary warehouses

GabbasovDinar (changes requested):
@danielduqma Thank you for your contribution, please check out the small comments

sheragar:
LGTM

pos_stock_available_online/models/stock_notifier_pos_mixin.py
danielduqma: This mixin was intended just to be inherited by `stock.move` and `stock.quant`, but I modified it as you proposed to be more extendable

pos_stock_available_online/models/stock_notifier_pos_mixin.py
danielduqma: You're right, this was a first approach but the logic was moved to `_get_warehouses_to_notify`

GabbasovDinar (approved):
Code LGTM

danielduqma:
Hi @OCA/pos-maintainers can this be merged please? The only checks in red are related to tests, but module hasn't got any since the beginning. Thanks!

ivantodorovich:
thanks!

/ocabot merge major

PR #1252, opened 2024-10-07
[17.0][ADD]: Improve pos_order_to_sale_order

victorvermot:
# Before Change:
- The lang used in the SO is not the one set on the partner
- The pricelist is not the one set on the partner

# After Merge:
- The lang is the one used in the partner -> a german partner will have the article description in german
- The pricelist is the correct one by retriggering compute methods

### DIFF
--- a/pos_order_to_sale_order/models/sale_order.py
+++ b/pos_order_to_sale_order/models/sale_order.py
@@ -13,6 +13,14 @@ class SaleOrder(models.Model):
         PosSession = self.env["pos.session"]
         session = PosSession.browse(order_data["pos_session_id"])
         SaleOrderLine = self.env["sale.order.line"]
+        partner_pricelist = (
+            order_data["pricelist_id"]
+            if order_data["pricelist_id"]
+            else self.env["res.partner"]
+            .browse(order_data["partner_id"])
+            .property_product_pricelist.id
+        )
+        self.pricelist_id = partner_pricelist
         order_lines = [
             Command.create(SaleOrderLine._prepare_from_pos(sequence, line_data[2]))
             for sequence, line_data in enumerate(order_data["lines"], start=1)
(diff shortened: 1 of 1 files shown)


pos_order_to_sale_order/models/sale_order.py
```diff
@@ -13,6 +13,14 @@ def _prepare_from_pos(self, order_data):
...
         session = PosSession.browse(order_data["pos_session_id"])
         SaleOrderLine = self.env["sale.order.line"]
+        partner_pricelist = (
+            order_data["pricelist_id"]
+            if order_data["pricelist_id"]
+            else self.env["res.partner"]
+            .browse(order_data["partner_id"])
+            .property_product_pricelist.id
+        )
+        self.pricelist_id = partner_pricelist
```
legalsylvain, line 23: pricelist is selected on pos.order. Why the pricelist is not selected correctly in the point of sale ?

legalsylvain:
> The lang is the one used in the partner -> a german partner will have the article description in german

I don't get it. why the data are not in german in the point of sale ?

victorvermot:
> > The lang is the one used in the partner -> a german partner will have the article description in german
> 
> I don't get it. why the data are not in german in the point of sale ?

In our case, a French vendor is using the pos to scan the articles. His/her system is in french. When creating the invoice for the partner, we want to use their language.

pos_order_to_sale_order/models/sale_order.py
```diff
@@ -13,6 +13,14 @@ def _prepare_from_pos(self, order_data):
...
         order_lines = [
             Command.create(SaleOrderLine._prepare_from_pos(sequence, line_data[2]))
```
legalsylvain, line 25: I guess adding lang context over there should made the work.

PR #1253, opened 2024-10-14
[ADD] pos_full_refund: New module for quantity autocomplete in ticket view

LorenzoC0:
This module adds a button on the ticket screen, when this button is clicked it autocompletes all the rows of the selected order with the correct quantity and returns to the main screen of the point of sale.


eLBati:
Porting to 18 with https://github.com/OCA/pos/pull/1407

PR #1010, opened 2023-06-08
[15.0][MIG] pos_user_restriction: Migration to 15.0

ghost:

PR #1255, opened 2024-10-25
[DO NOT MERGE] Testing pos_user_restriction without hack

em230418:
Just testing how it works without hack


em230418:
To be honest, I don't know why hack is that used. It works fine without it. Unfortutaly I can't ask author of this hack, since author's account is wiped.
https://github.com/OCA/pos/pull/1010

em230418:
@ivantodorovich maybe you are aware?

PR #1258, opened 2024-10-30
[16.0][ADD] pos_auto_invoice

huguesdk:
module to set point-of-sale orders as to-invoice by default.


legalsylvain:
hi @huguesdk. Thanks for this contribution.

At a first look, LGTM to me.
some remarks : 
- Is it possible to add a boolean auto_invoice_by_default on pos.config and base the change in the PoS regarding that new flag ? Pro : 
- -> it will be installable on multi-company or multi-pos-config context, because such default auto_invoice is great but can be limited to some pos.config.
- ->  we can disable it on demo data, to avoid "excude" mechanism on CI.

What do you think ?

huguesdk:
i completely agree. like this?

do you think that it should be enabled by default and disabled by demo data? i think that disabled by default makes sense, especially since having a different default behavior with and without demo data could be confusing for users.

pos_auto_invoice/static/src/js/models.js
```diff
@@ -0,0 +1,20 @@
+odoo.define("pos_auto_invoice.models", function (require) {
+    "use strict";
+
+    const {Order} = require("point_of_sale.models");
+    const Registries = require("point_of_sale.Registries");
+
+    // eslint-disable-next-line no-shadow
+    const AutoInvoiceOrder = (Order) =>
+        // eslint-disable-next-line no-shadow
+        class AutoInvoiceOrder extends Order {
```
legalsylvain:
```suggestion
    const AutoInvoiceOrder = (OriginalOrder) =>
        class extends OriginalOrder {
```

pos_auto_invoice/static/src/js/models.js
```diff
@@ -0,0 +1,20 @@
...
+            constructor() {
+                super(...arguments);
+                if (this.pos.config.invoice_by_default) {
+                    this.to_invoice = true;
+                }
+            }
+        };
```
legalsylvain: Nice to have : not blocking : consider the following use case : 
- create a new order (invoice is set to True)
- user uncheck the "to invoice"
- reload the PoS
- the order is checked again

(not tested, but I guess)

overload export_from_JSON / load_from_JSON should be fix the problem I guess.

legalsylvain:
> do you think that it should be enabled by default and disabled by demo data? i think that disabled by default makes sense, especially since having a different default behavior with and without demo data could be confusing for users

totally agree!

huguesdk:
thanks a lot for the review! i added your suggestion to avoid variable shadowing.

i also ensured that the value of `to_invoice` is kept for existing orders. strangely, [the default behavior in `init_from_JSON()`](https://github.com/OCA/OCB/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2584) is to always set it to `false`, although it is correctly handled in `export_as_JSON()`.

legalsylvain (approved):
Thanks !

polchampion (approved):
successful functional test on rotor-v16-test

huguesdk:
@OCA/pos-maintainers would someone please merge this?

huguesdk:
@legalsylvain thanks! but it doesn’t seem to work. is ocabot sleeping? ☺️

legalsylvain:
hi @huguesdk. 

Related to https://github.com/OCA/odoo-community.org/issues/204#issuecomment-2782433559 maybe ?

huguesdk:
@legalsylvain would you please try again?

PR #1259, opened 2024-10-30
[16.0][ADD] pos_auto_invoice_default_partner

huguesdk:
compatibility layer between `pos_auto_invoice` and `pos_default_partner`.

depends on #1258.


polchampion (changes requested):
unsuccessful functional test on rotor-v16-test

I still have the error "please select the customer" when trying to validate a payment without a customer

polchampion:
> unsuccessful functional test on rotor-v16-test
> 
> I still have the error "please select the customer" when trying to validate a payment without a customer

This accidental behaviour is preferable to the v12 way, because accounting-wise it should not be a default option to create invoices for a default customer when no customer is manually selected. It's also more logical, and there is no use case needing the old behaviour.

@huguesdk Does this module do anything then ? If we remove the part of the code that isn't used, is there something left in this module ?

huguesdk:
@polchampion yes it does, but it doesn’t work if `pos_loyalty` is also installed. i fixed that, but it makes more sense to implement it directly in `pos_default_partner`, so i did it there: #1276.

PR #1260, opened 2024-10-30
[16.0][FIX] pos_order_to_sale_order: rpc helper

danielduqma:
`rpc` attribute used comes from LegacyComponent, that is deprecated and throws an error in some specific cases. Changing it by `this.env.services` works properly in all cases.

This can be reproduced executing this from other component:
```javascript
const create_order = new CreateOrderPopup(props, this.env);
const {sale_order_id} = await create_order._createSaleOrder("draft");
```

FL-556-4346

### DIFF
--- a/pos_order_to_sale_order/static/src/js/CreateOrderPopup.js
+++ b/pos_order_to_sale_order/static/src/js/CreateOrderPopup.js
@@ -43,11 +43,12 @@ odoo.define("point_of_sale.CreateOrderPopup", function (require) {
         async _createSaleOrder(order_state) {
             const current_order = this.env.pos.get_order();
             framework.blockUI();
-            return await this.rpc({
-                model: "sale.order",
-                method: "create_order_from_pos",
-                args: [current_order.export_as_JSON(), order_state],
-            })
+            return await this.env.services
+                .rpc({
+                    model: "sale.order",
+                    method: "create_order_from_pos",
+                    args: [current_order.export_as_JSON(), order_state],
+                })
                 .catch(function (error) {
                     throw error;
                 })


legalsylvain:
Hi. Do you mean it should be changed in all modules ? 

```
rgrep "this\.rpc" ./ --include=*.js
./pos_order_to_sale_order/static/src/js/CreateOrderPopup.js:            return await this.rpc({
./pos_product_label/static/src/PrintLabelPopup.esm.js:            const data = await this.rpc({
./pos_product_label/static/src/PrintLabelPopup.esm.js:        const response = await this.rpc({
./pos_partner_firstname/static/src/js/PartnerDetailsEdit.js:                this.rpc({
./pos_financial_risk/static/src/js/PaymentScreen.esm.js:            this.rpc({
./pos_order_to_sale_order_delivery/static/src/js/Popups/CreateOrderPopup.esm.js:    const response = await this.rpc({
./pos_partner_location_google_map/static/src/js/PartnerMapGoogleEdit.esm.js:            return this.rpc({
./pos_lot_barcode/static/src/js/Screens/ProductScreen.js:                    foundLotIds = await this.rpc({
./pos_order_to_sale_order_sale_financial_risk/static/src/js/CreateOrderPopup.esm.js:            const [partnerRiskValues] = await this.rpc({
./pos_loyalty_redeem_payment/static/src/js/PaymentScreen.esm.js:                const payload = await this.rpc({

rgrep "this\.env\.services\.rpc" ./ --include=*.js
./pos_lot_selection/static/src/js/models.js:                    return await this.env.services.rpc(
./pos_loyalty_redeem_payment/static/src/js/PaymentScreen.esm.js:            return await this.env.services.rpc({
```

danielduqma:
Thanks for reviewing! I'm not sure enough and I haven't tested globally, but as I analyzed:
- `CreateOrderPopup` extends `AbstractAwaitablePopup`
- `AbstractAwaitablePopup` extends `PosComponent`
- `PosComponent` extends `LegacyComponent`

The last component is defined [here](https://github.com/OCA/OCB/blob/ae8af1aac1c7718db7e2fafe24c8a1504c72b0ce/addons/web/static/src/legacy/legacy_component.js#L49), and it's just a wrapper over `owl.Component` that mainly adds `rpc` method for legacy. I think direct use of this `rpc` is deprecated in some way, and in fact `LegacyComponent` no longer exists in 17.0 and newer, where those calls are managed by `orm` service.

So, in our case the code in the main comment leads to an error, but I haven't found any other incompatibilities with the other modules.

Regards

danielduqma:
Hi @OCA/pos-maintainers can this be merged please? Despite it is difficult to be checked (only in code), it doesn't break any behaviour. Thanks!

PR #1273, opened 2024-12-04
update pos_product_label to pass the pos settings on wizzard context

cyrilmanuel:
The objective is to pass into the context of the wizzard some pos settings (example if need to specify a preferred printer into the wizzard)

### DIFF
--- a/pos_product_label/models/pos_session.py
+++ b/pos_product_label/models/pos_session.py
@@ -32,6 +32,7 @@ class PosSession(models.Model):
         """
         vals = self._prepare_product_label_layout_data(data)
         wizard = self.env["product.label.layout"].create(vals)
+        wizard = wizard.with_context(pos_config_id=self.config_id)
         if data.get("pos_quantity") == "order":
             wizard = wizard.with_context(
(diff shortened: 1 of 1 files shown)


victorvermot (approved):

pos_product_label/models/pos_session.py
```diff
@@ -32,6 +32,7 @@ def print_product_labels(self, data):
         """
         vals = self._prepare_product_label_layout_data(data)
         wizard = self.env["product.label.layout"].create(vals)
+        wizard = wizard.with_context(pos_config_id=self.config_id)
```
santostelmo, line 35: Where this variable `pos_config_id` is going to be used ? I don't see any usage in module `pos_product_label`

ivantodorovich: Neither in Odoo core 🤔

ivantodorovich: If what you need is to add a custom context key used in an external module, you can always override `print_product_labels` like this, in your custom module:

```py
def print_product_labels(self):
    self = self.with_context(pos_config_id=self.config_id)
    return super().print_product_labels()
```

PR #1274, opened 2024-12-10
[16.0] pos_receipt_replace_user_by_trigram: Drop compute function

grindtildeath:
The compute function does not allow any customization on how the trigram had to be generated, and there a multiple ways to define how trigram must be generated.

Moreover, the compute did generate duplicates and there was no way to fix manually since the field was not stored. Also, making it stored but still computed did risk unwanted updates in case name was updated.

Having a basic stored field allows each company to customize how trigram can be generated using a server action. The logic previously provided is then moved to a cron that can be customized to suit trigram generation for each company.


grindtildeath:
ping @trisdoan @francesco-ooops @legalsylvain @ivantodorovich @cyrilmanuel

trisdoan (approved):
Sounds like a great idea to me! Thanks

Could you squash the fixup?

grindtildeath:
@OCA/pos-maintainers anyone to reopen please :pray:

ivantodorovich:
Github's not letting me :/ :

image

Could you re-create the PR, I'll review & merge asap, given that this one already had one approval

cyrilmanuel:
superseeded --> https://github.com/OCA/pos/pull/1493

PR #1276, opened 2024-12-13
[16.0][IMP] pos_default_partner: set default partner also in pos ui

huguesdk:


huguesdk:
@OCA/pos-maintainers could someone please review or merge this?

polchampion (approved):
Successful functional test
- [x] default partner auto selected in POS at opening and after each sale
- [x] partner can be changed or manually unselected
- [x] when no default partner is set in the config, no error and no partner is auto selected

danielduqma:
Hello @huguesdk,

Thanks for your contribution! In my opinion, there are processes in PoS interface that need a specific partner, and they would use by default the one in this module. For example, if cashier clicks on invoice button, an invoice for default partner would be generated, with no warning or redirection to partner selection screen. That's because there would be no distinction between a real customer and default partner ([as in invoice example](https://github.com/odoo/odoo/blob/4167fe687443ef825d4aa051f5913dd18371e4bd/addons/point_of_sale/static/src/js/Screens/PaymentScreen/PaymentScreen.js#L387)). In addition, default customer data would be shown on ticket.

Could you please take this in consideration and rebase to rebuild runboat and confirm those behaviours? Thanks in advance

Regards.

huguesdk:
@danielduqma thanks for your message! indeed, with a default partner configured, there would be no warning if an invoice is created and no partner has been selected, but there is a clear visual hint that a partner is already set (name and teal background).

bildo

after all, isn’t this the expected behavior of this module? should the default partner be removed in case an invoice is generated, or the warning still present if the default partner is selected?

a runboat build has been triggered.

danielduqma:
Thanks @huguesdk! As we see this point (Spain) it would not be right to create a complete invoice assigned to this generic partner. In fact, there are many processes apart from invoicing that require a real customer to be selected, like `pos_order_to_sale_order`, and right now there is no blocking alert that prevents cashier to place it with default partner.

Regards.

huguesdk:
@danielduqma i understand. what would you suggest as behavior, then? the roadmap contained “set the default customer also on the pos ui.”, which is what is done here.

danielduqma:
Yes, this is in the roadmap but it hasn't been done I guess because it's not as simple as assigning default partner at the beginning of the process. Besides, this assignment would include its data in ticket, so a real customer would be able to see it.

I don't really know what was pretended to be done, according to roadmap, but the other option I'd prefer is to change placeholder of set customer button ([here](https://github.com/odoo/odoo/blob/b51c927ee78fcc6afa3da3a1979ae1558833b963/addons/point_of_sale/static/src/xml/Screens/ProductScreen/ActionpadWidget.xml#L13) and [here](https://github.com/odoo/odoo/blob/b51c927ee78fcc6afa3da3a1979ae1558833b963/addons/point_of_sale/static/src/xml/Screens/PaymentScreen/PaymentScreen.xml#L71)), showing cashier that default partner will be assigned if no customer is selected but without actual modifications in order until it is saved at server side.

Regards.

remytms:
Merged via the PR #1468

PR #1279, opened 2024-12-20
[17.0][MIG]  pos_discount_all: Migration to 17.0 

peluko00:


peluko00:
Hi @legalsylvain can you help me with the computation of the discount amount with tax without any discount in the template? Thanks in advance!

PR #1282, opened 2025-01-02
18.0 mig pos order to sale order

fd-oerp:


legalsylvain:
Hi @fd-oerp. Thanks for porting this module. Could you fix red CI ? ``pre-commit run -a`` 

thanks !

PR #1287, opened 2025-01-17
[17.0][ADD] pos_loyalty_card_fixed_expiration_date: New module pos_loyalty_card_fixed_expiration_date

BernatObrador:
This module introduces the ability to set a fixed expiration date for gift cards.

Follow these steps to configure and use the module:

1. Navigate to **Point of Sale -> Products -> Gift Cards & eWallet**.
2. Create a new record and set the **Program Type** to **Gift Card**.
3. Specify the **Duration Days**.
4. Generate gift cards without setting an expiration date.
   - The generated gift cards will automatically have an expiration date set to **today's date + Duration Days**.
5. If you manually set an expiration date, the system will use this specified date instead.

cc https://github.com/APSL 165899

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @ppyczko please review


peluko00 (approved):
LGTM! Code review and tested in runboat

mpascuall (approved):
LGTM, tested in runboat and reviewed code

ppyczko (approved):
LGTM!

BernatObrador:
Hi @ivantodorovich, could you review please. Thanks!

pos_loyalty_card_fixed_expiration_date/models/loyalty_card.py
```diff
@@ -0,0 +1,24 @@
...
+class LoyaltyCard(models.Model):
+    _inherit = "loyalty.card"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        # If the expiration date is not set, we set it to
+        # the current date + the fixed duration
+        if vals_list:
+            if not vals_list[0].get("expiration_date", False):
```
ivantodorovich, line 18: This doesn't seem right.

`vals_list` here can be a list containing completely different values. You shouldn't rely only on the first one, for all the records.

Instead you can do something like this:

```py
for record, vals in zip(res, vals_list):
    ...
```

This works because create will return the records on the same order than the provided values that created them.

---

But moreover, I wonder if it wouldn't be a better idea to simply make the `expiration_date` a computed field with `store=True`, `readonly=False`, `precompute=True`, and `@api.depends('program_id')`, instead of overriding `create`.

ivantodorovich (changes requested):

ivantodorovich (changes requested):
I fail to see the real dependency on Point of Sale.
It looks like a generic loyalty module, that should be proposed in https://github.com/OCA/sale-promotion instead (without the `pos_sale` dependency)

BernatObrador:
@ivantodorovich Totally agree with your changes! Thanks for your time.

PR #1289, opened 2025-01-28
[16.0][FIX] pos_access_right: replace deleteOrder for _onDeleteOrder

paku1036:
Correction has been made to block the delete button.


flotho (approved):
make sense in the end
LGTM

PR #1293, opened 2025-02-10
[18.0][MIG] pos_edit_order_line: Migration to 18.0

BhaveshHeliconia:


pos_edit_order_line/static/src/xml/EditOrderPopup.xml
```diff
@@ -2,56 +2,75 @@
...
+        <!--                <footer class="footer">-->
+        <!--                    <div class="button confirm" t-on-click="confirm">-->
+        <!--                        <t t-esc="props.confirmText" />-->
+        <!--                    </div>-->
+        <!--                    <div class="button cancel" t-on-click="cancel">-->
+        <!--                        <t t-esc="props.cancelText" />-->
+        <!--                    </div>-->
+        <!--                </footer>-->
+        <!--            </div>-->
+        <!--        </div>-->
```
FrancoMaxime: If it's not needed, can you remove it?

FrancoMaxime (commented):
LGTM: code & functional review.

A minor comment

BhaveshHeliconia:
Hey @OCA/pos-maintainers would be great if someone could have a look at this module.

pos_edit_order_line/models/ir_module_module.py
```diff
@@ -0,0 +1,9 @@
+from odoo import api, models
+
+
+class IrModuleModule(models.Model):
+    _inherit = "ir.module.module"
+
+    @api.model
+    def _load_pos_data_fields(self, config_id=None):
+        return super()._load_pos_data_fields()
```
ivantodorovich, line 9: This override is not doing anything, you can remove it

pos_edit_order_line/static/src/js/EditOrderButton.esm.js
```diff
@@ -0,0 +1,37 @@
...
+import {AlertDialog} from "@web/core/confirmation_dialog/confirmation_dialog";
+import {EditOrderPopup} from "./EditOrderPopup.esm";
+
+patch(ControlButtons.prototype, {
+    async onClickEditOrder() {
+        if (
+            this.pos.get_order().get_orderlines() &&
+            this.pos.get_order().get_orderlines().length > 0
+        ) {
+            var order_lines = this.pos.get_order().get_orderlines();
```
ivantodorovich: Same thing, but with fewer characters and calls to `get_order()` and `get_orderlines()`

```suggestion
        const order_lines = this.pos.get_order().get_orderlines();
        if (order_lines?.length) {
```

pos_edit_order_line/static/src/js/EditOrderPopup.esm.js
```diff
@@ -0,0 +1,71 @@
...
+            Object.assign(
+                {},
+                {_id: this._nextId()},
+                typeof item === "object" ? item : {text: item}
+            )
+        );
+    }
+    handleChange({id, changes}) {
+        var self = this;
+        self.changes[id] = changes;
```
ivantodorovich:
```suggestion
        this.changes[id] = changes;
```

ivantodorovich (approved):
Thanks!

Small comments, pre-approving

pos_edit_order_line/models/ir_module_module.py
BhaveshHeliconia: This override is required for method signature compatibility: POS may call _load_pos_data_fields(config_id), but core doesn’t accept that argument. The override keeps behavior identical while preventing a TypeError in POS tests.

OCB Error :- 
During handling of the above exception, another exception occurred:

  File "/opt/odoo/odoo/tests/common.py", line 1499, in _websocket_request
    return f.result(timeout=timeout)
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 460, in result
    raise TimeoutError()
concurrent.futures._base.TimeoutError

During handling of the above exception, another exception occurred:

  File "/opt/odoo/odoo/tests/common.py", line 1221, in stop
    self._websocket_request('Runtime.evaluate', params={'expression': """
  File "/opt/odoo/odoo/tests/common.py", line 1501, in _websocket_request
    raise TimeoutError(f'{method}({params or ""})')
TimeoutError: Runtime.evaluate({'expression': "\n                ('serviceWorker' in navigator) &&\n                    navigator.serviceWorker.getRegistrations().then(\n                        registrations => Promise.all(registrations.map(r => r.unregister()))\n                    )\n                ", 'awaitPromise': True})
2026-01-21 11:51:08,519 305 ERROR odoo odoo.addons.pos_display_total_quantity.tests.test_frontend: FAIL: TestDisplayTotalQty.test_display_total_qty
  File "/__w/pos/pos/pos_display_total_quantity/tests/test_frontend.py", line 13, in test_display_total_qty
    self.start_tour(
  File "/opt/odoo/addons/account/tests/common.py", line 1496, in start_tour
    super().start_tour(url_path, tour_name, step_delay, **kwargs)
  File "/opt/odoo/odoo/tests/common.py", line 2305, in start_tour
    return self.browser_js(url_path=url_path, code=code, ready=ready, timeout=timeout, success_signal="tour succeeded", **kwargs)
  File "/opt/odoo/odoo/tests/common.py", line 2279, in browser_js
    self.fail('%s\n\n%s' % (message, error))
AssertionError: The test code "odoo.startTour('DisplayTotalQty', {"stepDelay": 0, "keepWatchBrowser": false, "debug": false, "startUrl": "/pos/ui?config_id=7", "delayToCheckUndeterminisms": 0})" failed

ivantodorovich: I'm sorry, I don't get it.

I still see no reason for this override in this module.

Can you elaborate?

BhaveshHeliconia: So the reason for this override is actually to handle a specific edge case in the standard Odoo POS code.

When a regular POS user (who isn't an admin) loads the session, ir.module.module triggers an AccessError. Odoo's pos_session.py catches this error and tries to recover, but the fallback code calls _load_pos_data_fields
 passing the config_id as an argument.

Since the base method in ir.module.module doesn't define that argument, it immediately crashes with a TypeError.

I added this override just to accept the config_id (so it doesn't crash) and then pass execution to super(). It's basically patching that core inconsistency so the tests don't fail for non-admin users.

I've removed the override in the latest commit so you can check the failed OCB logs yourself. That should verify the issue clearly!

Also, to be honest, I don't think this workaround is the right long-term solution. It only solves the crash for this specific module, but other POS PRs for 18.0 are likely going to fail with the exact same error.

If I am understanding this in the wrong direction, please correct me.

ivantodorovich: I find it weird that tests pass with official Odoo but not with OCA.
I will attempt a rebase

ivantodorovich: Well. It seems I don't have permissions to do so. Can you rebase?

pos_edit_order_line/models/__init__.py
```diff
@@ -1,2 +1,3 @@
 from . import pos_config
 from . import res_config_settings
+# from . import ir_module_module
```
ivantodorovich: So, it seems tests are green without this file. Can you remove it completely, then? 🙏🏻

ivantodorovich (changes requested):

pos_edit_order_line/models/__init__.py
BhaveshHeliconia: @ivantodorovich I've removed that. Thanks for the review!

ivantodorovich (approved):
/ocabot migration pos_edit_order_line

Thanks!

FrancoMaxime (approved):
LGTM: code review

BhaveshHeliconia:
@ivantodorovich Could you please merge this? Thanks!

PR #1299, opened 2025-02-10
[18.0][MIG] pos_ticket_extra_company_info_l10n_fr: Migration to 18.0

BhaveshHeliconia:


flotho (changes requested):
Hi,
Thanks for the work.

I've tested on runboat with SIRET and registry on company and customer, nothing appears : 
image

If it works only for france, I think this module should have the documentation for this because SIRET is always displayed on contact forms whatever the country.
That's tricky.

Thanks

BhaveshHeliconia:
Hii @flotho,

Thank you for your feedback.

The module now displays both SIRET and company registry fields on the POS ticket, but only for companies configured as French (country code FR or DOM-TOM).

The documentation has been updated to clarify this France-specific behavior and field visibility.

If your company is not French, these fields will not appear on the POS ticket.

PR #1308, opened 2025-02-10
[17.0][MIG] pos_ticket_extra_company_info_l10n_fr: Migration to 17.0

BhaveshHeliconia:


BhaveshHeliconia:
Hey @OCA/pos-maintainers would be great if someone could have a look at this module.

pos_ticket_extra_company_info_l10n_fr/static/src/js/models.esm.js
```diff
@@ -1,22 +0,0 @@
...
-License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-*/
-
-odoo.define("pos_ticket_extra_company_info_l10n_fr.models", function (require) {
-    const {Order} = require("point_of_sale.models");
-    const Registries = require("point_of_sale.Registries");
-
-    const OrderTicketExtraInfo = (OriginalOrder) =>
-        class extends OriginalOrder {
-            export_for_printing() {
```
ivantodorovich, line 13: Are you sure this method not needed anymore?

Other modules are still implementing it. See for example: https://github.com/OCA/pos/blob/ecfdb08c2b4da93a8b98a9506a1fc755795b5bea/pos_receipt_gift_card/static/src/app/store/models/order.esm.js#L15

BhaveshHeliconia: Hello @ivantodorovich Thanks for the review!

In this case, I don't think overriding `export_for_printing()` is needed anymore.

The only purpose of this override was to expose the company SIRET to the receipt. In Odoo 17, the value is already added on the Python side (`res.company`) and is available directly in the receipt data, so `ReceiptHeader` can access it without extending the order model.

The `export_for_printing()` override is still required in other modules (such as the gift card module) when they need to inject additional order-specific data into the receipt payload. Since this module doesn't add any extra data anymore, the override becomes redundant.

ivantodorovich: Ok thanks for the explanation!

ivantodorovich (approved):

PR #1320, opened 2025-02-21
[16.0][MIG] pos_cashback_warning: Migration to 16.0 (from 12.0)

legalsylvain:
### description

This modules extends the Point of sale to manage cashback.

**Note** : The cashback is an operation that consists, for a merchant to collect more money in check, credit card, etc. than the amount of the sale and give change in cash.

By default, if the payment lines of an Odoo PoS order generate change, the journal for the change will be the Cash journal.
That is ok if the original payment line is a Cash payment, but if it's a payment by credit card, check, meal voucher, etc. confirming the transaction will generate a cashback.

This module simply add a message in such case.

pos_cashback_warning

### Roadmap

* implement charging commission, for the cashback service.


legalsylvain:
trivial migration.
work in production since a year.
/ocabot merge nobump

PR #1264, opened 2024-11-20
[17.0][MIG] pos_payment_change: Migration to 17.0

sylvainvh:

PR #1331, opened 2025-03-13
[17.0][MIG] pos_payment_change

em230418:


em230418:
I am aware of https://github.com/OCA/pos/pull/1264

pos_payment_change/tests/test_module.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def test_04_payment_change_security(self):
+        self.pos_config.payment_change_policy = "refund"
+
+        order = self._sale(self.cash_payment_method, 35, self.bank_payment_method, 65)
+
+        # the demo user should be able to do this
+        user_demo = self.env.ref("base.user_demo")
+        wizard = (
+            self.PosPaymentChangeWizard.with_user(user_demo)
```
em230418: Hello, @legalsylvain!
Tests are failing because Marc Demo stopped being related to pos manager group.
I don't want to spend time finding it out, why it happens.

How about just creating user with pos manager rights?

pos_payment_change/models/pos_order.py
```diff
@@ -0,0 +1,101 @@
...
+                        "amount": -payment.amount,
+                        "payment_date": fields.Date.context_today(self),
+                    }
+                )
+
+            refund_order.action_pos_order_paid()
+
+            # Resale order and mark it as paid
+            # with the new payment
+            resale_order = self.copy(default={"pos_reference": self.pos_reference})
```
em230418: Hello, @etobella!
Your module pos_order_copy breaks every other module, that uses "copy" method.
In this case to make compatibility with pos_order_copy, I need to manually copy values from every field, that you marked as "copy=False". Could you please fix your pos_order_copy module?

Here is what happens in auto tests in github actions:
```
2025-03-14 12:17:12,018 262 ERROR odoo odoo.addons.pos_payment_change.tests.test_module: ERROR: TestModule.test_02_payment_change_policy_refund
  File "/__w/pos/pos/pos_payment_change/tests/test_module.py", line 147, in test_02_payment_change_policy_refund
    self._change_payment(
  File "/__w/pos/pos/pos_payment_change/tests/test_module.py", line 92, in _change_payment
    wizard.button_change_payment()
  File "/__w/pos/pos/pos_payment_change/wizards/pos_payment_change_wizard.py", line 98, in button_change_payment
    orders = order.change_payment(new_payments)
  File "/__w/pos/pos/pos_payment_change/models/pos_order.py", line 72, in change_payment
    resale_order = self.copy(default={"pos_reference": self.pos_reference})
  File "/opt/odoo/odoo/models.py", line 5597, in copy
    record_copy = self.create(vals)
  File "<decorator-gen-227>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 430, in _model_create_multi
    return create(self, [arg])
  File "/opt/odoo/addons/point_of_sale/models/pos_order.py", line 482, in create
    session = self.env['pos.session'].browse(vals['session_id'])
KeyError: 'session_id'
 ````

Comment (etobella) [reply pos_payment_change/models/pos_order.py]: Mmmh, I get your concerns, however, the idea of the module was allowing to make copy properly. It would have no sense to copy the session for example, as the session should be set as the current session. The same for the order_date and so on... Actually, this copy=False should have been defined by Odoo as a core functionality...

We have 2 options here: Just make a copier update and set the incompatibility between this modules or find a way to make them work together.

From my perspective, you can set the information directly on the copy that you are doing if you want them to work together as the copy False that we have settled have a lot of sense (you should set this information on an specific function if other `copy=False` fields appear over time)

**Comment by em230418** [comment]
> so it has been marked as stale and it will be closed automatically if no further activity occurs in the next 30 days.

Please no

**Comment by cvinh** [comment]
@em230418 maybe you can rebase to unlock runboat. Then your PR can be tested by reviewers

**Review by cvinh — CHANGES_REQUESTED** [review]
Tested on runboat LGTM
There's a small issue (it's not because of the migration) when you select a refunded order in the frontend : the 3 orders get selected at once and the original one opens.
Then we want an invoice, the invoice is linked to the original one.
Maybe you can add this in the roadmap part of the readme

**Comment by ilmir-k** [comment]
LGTM

**Review by ivantodorovich — APPROVED** [review]
/ocabot migration pos_payment_change
/ocabot merge nobump

**Comment by ivantodorovich** [comment]
@em230418 can you fix the issue reported above? 🙏🏻

**Comment by em230418** [comment]
> @em230418 can you fix the issue reported above? 🙏🏻

Done. The only thing I did that probably fixes the problem is recreating README.rst via pre-commit.
Result diff looks like this:

```diff
diff --git a/pos_payment_change/README.rst b/pos_payment_change/README.rst
index e6291f5..0be6e68 100644
--- a/pos_payment_change/README.rst
+++ b/pos_payment_change/README.rst
@@ -7,7 +7,7 @@ Point Of Sale - Change Payments
    !! This file is generated by oca-gen-addon-readme !!
    !! changes will be overwritten.                   !!
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   !! source digest: sha256:6881f0be1df613c788f593eed6eefeb05567b3991680bf98bcd8497749e53a1a
+   !! source digest: sha256:1c6e8c040d05dda8166904f53fc0ab8ba31c3e0fc1d9c158284f6dbd75c966c4
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
 .. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png

```

em230418:
rebased to upstream 17.0 in hope it will fix this

pedrobaeza:
Superseded by #1569

PR #1335, opened 2025-03-15
[16.0][IMP] pos_order_to_sale_order: Set default warehouse from session config

rianmz-genz:
The default warehouse for POS orders is now automatically set based on  
`session.config_id.picking_type_id.warehouse_id.id`, ensuring correct warehouse assignment.


legalsylvain (approved):
Make sense 

Could you fix the name of the module in the commit message and title ?

Lgtm. Cc @sebastienbeau

rianmz-genz:
> Make sense
> 
> Could you fix the name of the module in the commit message and title ?
> 
> Lgtm. Cc @sebastienbeau

Thanks for the feedback! I'll correct the module name in the commit message and title now.

PR #1338, opened 2025-03-19
[17.0][ADD] Add pos_partner_birthdate_warning module 

parvezqureshi:


cvinh (commented):
Why don't you simply load res_partner.age field that is already computed in partner_contact_birthdate instead of recompute it ? Is it because it's not stored ?

parvezqureshi:
> Why don't you simply load res_partner.age field that is already computed in partner_contact_birthdate instead of recompute it ? Is it because it's not stored ?

@cvinh @vehi-invitu They failed in some cases like the one below.

**Customer:**

- **Birthdate** = 20-Mar-2007
- **Current Date** = 19-Mar-2025
- In the POS screen, the age displays as **17y**

**Issue:**

If the session is closed on **21-Mar-2025** or later, the age should update to **18y** on **20-Mar-2025.**
However, if the POS screen is not refreshed, it will still display **17y**, even after the birthday has passed.

**Conclusion:**

**res_partner.age** does not correctly handle cases where the session is not closed daily, leading to outdated age calculations on the POS screen

cvinh:
@parvezqureshi can you please rebase (as runboat is stuck)

parvezqureshi:
> @parvezqureshi can you please rebase (as runboat is stuck)

Rebased

cvinh (changes requested):
Thanks for the PR.
Here's are steps that do not work as expected (warning is bypassed) :
- the partner does not have a birthdate
- we edit the partner and add the birthdate.
- we save the partner and it's immediatly selected in the pos order without the warning

cvinh:
I just realize that this module should be renamed to pos_customer_age_warning ... or pos_customer_birthdate_age_warning...

pos_customer_birthdate_warning is meant to warn if it's the customer's birthday ... 

WDYT ? @legalsylvain

parvezqureshi:
supersede https://github.com/OCA/pos/pull/1382

PR #1347, opened 2025-03-28
[17.0][ADD] Add pos_partner_birthdate_weight_mandatory module

vehi-invitu:
cc @cvinh


legalsylvain:
Same remark here. I think we should put the module in a dedicated repo.

cvinh:
@vehi-invitu sorry it's a wrong push, here in OCA please push pos_partner_weight module

legalsylvain:
hi @cvinh. I think that such modules (with weight on partner for exemple) are very specific. (pos or not). 
I think it will be used only (or very mainly) in an medical / hospital / pharmacy context. 

For this reason, I think we should put that modules in a dedicated medical repository.

I propose you :
- to use an existing repo, here : https://github.com/OCA/vertical-medical/ (It is totally empty ...)
- to propose a PR to add your github login in this list, so you'll have some write access on this repo. https://github.com/OCA/repo-maintainer-conf/blob/master/conf/psc/vertical.yml#L102

What do you think ?

PR #1355, opened 2025-04-02
[18.0][MIG] pos_remove_pos_category

thibaultrey:
Migration to 18.0


legalsylvain (changes requested):
Hi. Thanks for porting this module. 
I think that this module should be simplified. This hard change catég -> pos categ is quite intrusive. 

With some hook in the load of pos, we could load link product to categ.
What do you think ?

PR #1359, opened 2025-04-07
[16.0][ADD] pos_partner_load_new_data

flaenen:
This module extends the functionality of the point of sale module to load newly created or modified partner data.

The data is loaded in the background as soon as the point of sale screen listing the customers is opened.


polchampion (approved):
Successful functional test on the runboat

polchampion:
@OCA/pos-maintainers Could you take a look at this? This new module allows seamless loading of pos customers in the case of newly created partners (in our business case, new partners are continuously added in the backend during pos session openings).

pos_partner_load_new_data/static/src/js/PartnerListScreen.esm.js
```diff
@@ -0,0 +1,36 @@
...
+import {onWillStart} from "@odoo/owl";
+
+export const UpdatedPartnerListScreen = (OriginalPartnerListScreen) =>
+    class extends OriginalPartnerListScreen {
+        setup() {
+            super.setup();
+
+            onWillStart(async () => {
+                this.env.pos
+                    .loadPartnersBackground(
```
legalsylvain, line 16: not sure if it can help, but there is a more global function in the odoo core, here : 

https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L375

flaenen: This function takes into account the values of limited_partners_loading and partner_load_background and can potentially load only one new partner (the last one). The purpose of this module is to load all new partners in all cases. Using this function would therefore require checking the value of limited_partners_loading and possibly forcing the value of partner_load_background to true. This seems more constraining to me than using the loadPartnersBackground function.

polchampion:
@remytms could you review?

flaenen:
> @remytms could you review?

@remytms when you have a moment

huguesdk (approved):

polchampion:
@OCA/pos-maintainers hi, could you take a look and merge this? thank you!

PR #1360, opened 2025-04-07
[16.0][ADD]pos_order_to_sale_order_commitment_date

adasatorres:
This add-on allows you to add the delivery date to orders created with the pos_order_to_sale_order add-on.


pos_order_to_sale_order_commitment_date/models/pos_config.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 Binhex - Adasat Torres de León.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    iface_sale_order_allow_commitment_date = fields.Boolean()
```
legalsylvain: I guess that if we install the module, we want the feature to be enabled. Don't you think ? 
```suggestion
    iface_sale_order_allow_commitment_date = fields.Boolean(default=True)
```

pos_order_to_sale_order_commitment_date/models/sale_order.py
```diff
@@ -0,0 +1,39 @@
...
+from datetime import datetime
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    def create_order_from_pos(self, order_data, action):
```
legalsylvain: Please overload the function ``_prepare_from_pos``

pos_order_to_sale_order_commitment_date/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+After enabling the option mentioned in the configuration section, go to the Point of Sale.
+Once inside, create an order just like with the ``pos_order_to_sale_order`` add-on,
+and use the option to create a sale order.
+
+Once the popup appears, there is an option called **"Add Commitment Date"**.
+If you select it and click on any of the base options,
+a new popup will appear allowing you to add the delivery date.
```
legalsylvain, line 9: Could you add a screenshot ? 

thanks !

legalsylvain (changes requested):
Hi @adasatorres. thanks for sharing your work

main question : why not just update the main module pos_order_to_sale_order, adding an extra field in the popup that ask the state of the order to create ? 
I don't see the interest to create a dedicated module here.

Otherwise, some remarks inline.

thanks ! 

Note : I think you should handle timezone. when I put 5:00 in the PoS, then it is displayed 7:00 in the backoffice.

pos_order_to_sale_order_commitment_date/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,26 @@
...
+            <xpath
+                expr="//field[@name='pos_iface_create_invoiced_sale_order']/../.."
+                position="after"
+            >
+                <div class="col-12 col-lg-6 o_setting_box">
+                    <div class="o_setting_left_pane">
+                        <field
+                            name="iface_sale_order_allow_commitment_date"
+                            string="Allow Commitment Date"
+                        />
```
legalsylvain, line 20: the display is broken here : 

image

pos_order_to_sale_order_commitment_date/static/src/xml/Popups/CreateOrderPopup.xml
```diff
@@ -0,0 +1,26 @@
...
+                <td>
+                    <div
+                        class="button"
+                        t-att-class="{ highlight: state.addCommitmentDate }"
+                        t-on-click="() => this.onClickToAddCommitmentDate()"
+                    >
+                        <i class="fa fa-calendar" /> Add a Commitment Date
+                    </div>
+                </td>
+            </tr>
```
legalsylvain, line 23: hi. could you clean the display of the popup ? 

image

adasatorres:
Hello @legalsylvain, I believe that creating a dedicated module keeps the code clean and flexible, allowing it to be reused in other contexts and facilitating future customizations without affecting other processes in the system. Additionally, it prevents the main module from becoming complex and hard to maintain.

rrebollo (approved):
Great job! The code looks good to me (LGTM). Thanks a lot for your contribution! 🙌
Just one thought: I see quite a bit of code around timezone handling (if I'm not mistaken). Do you think context_timestamp could already cover your needs, at least on the Python side?

legalsylvain:
Hi.

just tested again. Thanks for the changes !

functionally, I don't understand why having two screens ? It makes the code quite complex. 

why not just having add an optional field in the main screen ? 

something, like : 

image

adasatorres:
"Hi @legalsylvain , I've finally decided to implement the functionality directly on pos_order_to_sale_order, so I'm closing this PR and opening a new one."

PR #1361, opened 2025-04-08
[16.0][IMP]pos_order_to_sale_order: Allow add the commitment date.

adasatorres:
This add-on allows you to add the commitment date to orders.  
I've included a reference to the previous PR to maintain continuity of the discussion #1360 .

### DIFF
--- a/pos_order_to_sale_order/README.rst
+++ b/pos_order_to_sale_order/README.rst
@@ -117,6 +117,10 @@ Contributors
 
 * Sylvain LE GAL (https://www.twitter.com/legalsylvain)
 
+* `Binhex <https://binhex.cloud>`_:
+
+  * Adasat Torres de León <a.torres@binhex.cloud>
+
 Maintainers
 ~~~~~~~~~~~
 
--- a/pos_order_to_sale_order/__manifest__.py
+++ b/pos_order_to_sale_order/__manifest__.py
@@ -15,11 +15,9 @@
     "data": ["views/view_res_config_settings.xml"],
     "assets": {
         "point_of_sale.assets": [
-            "pos_order_to_sale_order/static/src/css/pos.css",
-            "pos_order_to_sale_order/static/src/js/CreateOrderButton.js",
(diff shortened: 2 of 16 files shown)


rrebollo (approved):
Great job! The code looks good to me (LGTM). Thanks a lot for your contribution! 🙌
Just one thought: I see quite a bit of code around timezone handling (if I'm not mistaken). Do you think context_timestamp could already cover your needs, at least on the Python side?

legalsylvain (commented):
Thanks a lot ! Looks more simple.

One question : finally, don't see clearly the interest of the new setting "iface_sale_order_allow_commitment_date". 

i mean, this field is optional, so it's possible for the use to enter the date, or let empty. don't you think ?

christian-ramos-tecnativa:
The concern seems to be that the view could become too crowded in the future, so IMO keeping this as an option is more flexible for the users

legalsylvain (changes requested):
> The concern seems to be that the view could become too crowded in the future, so IMO keeping this as an option is more flexible for the users

OK.

I guess it's ready to be merged, as soon as the date problem is fixed.

edescalona (approved):

legalsylvain:
I think this comment is not handled : https://github.com/OCA/pos/pull/1361#pullrequestreview-2750974532

adasatorres:
> Great job! The code looks good to me (LGTM). Thanks a lot for your contribution! 🙌 Just one thought: I see quite a bit of code around timezone handling (if I'm not mistaken). Do you think context_timestamp could already cover your needs, at least on the Python side?

context_timestamp does not parse strings, it converts a UTC datetime to the user's timezone.
Looking at the code now, that part shouldn’t really be necessary. I’ll replace creating the datetime with the parser from dateutil.

adasatorres:
Good evening, @legalsylvain. I’ve already reviewed it and looked into an alternative. I also added some unit tests for the method to improve coverage. Sorry for the delay.

ivantodorovich (approved):
Seems good

PR #1363, opened 2025-04-14
[16.0][FIX] pos_payment_change: Change of payment should force the session

etobella:
Otherwise, something weird might happen. For example, if we have a rescue session and we try to do the change, the refund is assigned to another session, no the rescue one.

### DIFF
--- a/pos_payment_change/models/pos_order.py
+++ b/pos_payment_change/models/pos_order.py
@@ -53,7 +53,9 @@ class PosOrder(models.Model):
         elif self.config_id.payment_change_policy == "refund":
             # Refund order and mark it as paid
             # with same payment method as the original one
-            refund_result = self.refund()
+            refund_result = self.with_context(
+                force_session_id=self.session_id.id
+            ).refund()
             refund_order = self.browse(refund_result["res_id"])
             for payment in self.payment_ids:
                 refund_order.add_payment(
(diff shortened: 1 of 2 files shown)


legalsylvain (commented):
I'm more in favor to reuse core feature.

something like 


    def _prepare_refund_values(self, current_session):
        session = current_session
        if force_session_id in self.env.context:
            session = self.env["pos.session"].browse(self.env.context.get("force_session_id")
        return super()._prepare_refund_values(session)



Then : only replace ``refund_result = self.refund()`` by ``refund_result = self.with_context(force_session_id=self.session_id.id)).refund()``


Otherwise, LGTM. Thanks !

etobella:
Sure! Not a problem :smile:

etobella:
@legalsylvain can you check it again?

pos_payment_change/models/pos_order.py
```diff
@@ -99,3 +101,11 @@ def _check_payment_change_allowed(self):
                     session=", ".join(closed_orders.mapped("session_id.name")),
                 )
             )
+
+    def _prepare_refund_values(self, current_session):
+        session = current_session
```
legalsylvain, line 106: I'm not sure it will work if the session is closed. A test should be done on state. Dont you think ?

luisDIXMIT (approved):
LGTM!

pos_payment_change/models/pos_order.py
etobella: @legalsylvain Test added

pedrobaeza: @etobella what @legalsylvain means is to check in this method which is the state of the session.

PR #1364, opened 2025-04-15
[IMP] pos_partner_vat_required: Validate VAT

peluko00:


legalsylvain (changes requested):
Hi.

Thanks a lot for your contribution.

However, I think this new interesting feature you introduce is not related to `pos_partner_vat_required` module.

I think that a little light module named `pos_partner_vat_valid` with your code should do the job. (install `pos_partner_vat_valid` without `pos_partner_vat_required` looks legit IMO, do'nt you think ?)

thanks !

peluko00:
> Hi.
> 
> Thanks a lot for your contribution.
> 
> However, I think this new interesting feature you introduce is not related to `pos_partner_vat_required` module.
> 
> I think that a little light module named `pos_partner_vat_valid` with your code should do the job. (install `pos_partner_vat_valid` without `pos_partner_vat_required` looks legit IMO, do'nt you think ?)
> 
> thanks !

Yes, i'll do another module. Thanks!

PR #1365, opened 2025-04-15
[ADD] pos_partner_vat_valid: New module pos_partner_vat_valid

peluko00:
New module for validate vat in POS partner view.

cc https://github.com/APSL 9140
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall @BernatObrador @ppyczko  @legalsylvain please review


pos_partner_vat_valid/static/src/app/PartnerDetailsEdit.esm.js
```diff
@@ -0,0 +1,31 @@
...
+
+patch(PartnerDetailsEdit.prototype, {
+    setup() {
+        super.setup(...arguments);
+        this.popup = useService("popup");
+        Object.assign(this.props.imperativeHandle, {
+            save: async () => this.saveChanges(),
+        });
+    },
+    async saveChanges() {
```
legalsylvain: I'm not sure you can change a non async function to a async function. I guess you can break inheritance mechanism here. Don't you think ?

legalsylvain (commented):
hi @peluko00. Thanks for sharing your work. Just a technical question here. otherwise, LGTM. Thanks !

pos_partner_vat_valid/static/src/app/PartnerDetailsEdit.esm.js
peluko00: I just test other module such as pos_address_required, vat_required works good with no async inheritance function but i don't know how to call ORM without async call.

legalsylvain: @ivantodorovich : any point of view about that technical point ?

thanks !

ivantodorovich: The difference with async methods and non-async methods is that the first ones will return a `Promise` that resolves to the async method return value, while the later will just return the value as is.

An async method is almost the same as a regular method that returns a `Promise`, from an outer point of view.

Now, I'd avoid changing it if possible, but I don't expect any issues here because this method is not expected to return anything. So it doesn't matter if the caller calls `await` or not, it will never expect anything as a return value, so if it now returns a `Promise` it's just as fine.

How would I avoid it?

Well, you wanted to use an `async` method so that you can `await` other async methods calls inside. But you could do it with the old-fashioned way: handle the promises directly.

Instead of `await this.pos.orm.call(..)`, you'd do `this.pos.orm.call(..).then((result) => { ...}))`

⚠️ careful with the `super` keyword, though. It will probably not work inside the callback fn. So you may need to store it in a variable before hand.

legalsylvain: Thanks a lot for your complete answer !

peluko00: Hi @ivantodorovich, many thanks for your help.
I refactored the code with handle the promise.
Can you review again please?

legalsylvain (approved):

legalsylvain:
Thansk a lot !

BernatObrador (approved):
LGTM!

ppyczko (approved):
LGTM

peluko00:
It's ready for merge @legalsylvain ?

mpascuall (approved):
LGTM

ivantodorovich (approved):

ivantodorovich:
Thanks!

/ocabot merge nobump

PR #1366, opened 2025-04-16
[16.0][IMP]pos_access_right: compatibility with multi-employee option.

adasatorres:
This implementation aims to enable the option to use the permissions of the user linked to the employee when the option to use multiple employees per session is active. If the employee has no related user, the permissions of the user logged into the backend are used instead.

### DIFF
--- a/pos_access_right/README.rst
+++ b/pos_access_right/README.rst
@@ -36,6 +36,8 @@ This module extends Odoo Point Of Sale features, restricting possibility to cash
 * **PoS - Many Orders**: The cashier can many orders at the same time;
 * **PoS - Delete Order**: The cashier can not delete a full order;
 
+This addon is compatible with the multi-employee per session option.
+
 **Table of contents**
 
 .. contents::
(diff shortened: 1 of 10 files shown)


legalsylvain (changes requested):
With that change, you force to install hr module, that it is not acceptable. 

you have two choices : 
- a lazy dependency. make the module working with and without pos_hr installed. (not sure it's possible).
- create a dedicated pos_hr_access_right.

rrebollo:
@adasatorres I guess you’ll go with the dedicated addon from @legalsylvain’s options. I’m generally OK with the feature, but the fallback to the logged-in user’s rights feels a bit unclear. Could the feature make it more explicit when an employee is acting with those permissions? A small use case might help clarify the intent too.

adasatorres:
@legalsylvain Alright, I’ll create a dedicated addon.

adasatorres:
@rrebollo the use of the logged-in user's permissions as a fallback is intended for those employees who don’t have an associated user account. A use case would be, for example, when you have a point of sale and want all employees without user accounts to have a specific permission — in that case, you log in with a user that has that specific permission. 

I hope I explained myself well, best regards.

PR #1372, opened 2025-04-27
[16.0][FIX] pos_membership_extension: add dependency

flaenen:
Without the pos_membership module, it is possible to sell a membership without creating an invoice and without selecting a customer.


polchampion (approved):
functional LGTM

polchampion:
@legalsylvain what do you think?

legalsylvain:
Hi. I don't understand why changing this dependency.

Maybe better put pos_membership as auto_install=True, if we thing that this module should be allways installed (if pos and membership are installed).

What do you think ?

PR #1374, opened 2025-04-28
[16.0][ADD] pos_membership_delegated_partner

flaenen:
WIP

This module implements the features of the Membership Delegated Partner module in the point of sale UI. This allows a customer to purchase memberships for other partners.

### Dependencies

- pos_membership
- https://github.com/OCA/vertical-association/pull/207

### Usage

1. Open the point of sale
2. Add a membership product to the order
3. Select the corresponding order line
4. Click on the "Delegated Member" button that has just appeared
5. Select a partner from the screen that opens

Once the order has been validated, the designated partner becomes a member.


pos_membership_delegated_partner/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "website": "https://github.com/OCA/pos",
+    "author": "Coop IT Easy SC, Odoo Community Association (OCA)",
+    "maintainers": ["flaenen"],
+    "license": "AGPL-3",
+    "depends": ["pos_membership", "membership_delegated_partner"],
+    "assets": {
+        "point_of_sale.assets": [
+            "pos_membership_delegated_partner/static/src/css/*.css",
+            "pos_membership_delegated_partner/static/src/js/*.js",
+            "pos_membership_delegated_partner/static/src/xml/*.xml",
```
huguesdk: i think it’s best to have js and xml files in the corresponding directory of the files they extend. for example, `ProductScreen.esm.js` should go to `static/src/js/Screen/ProductScreen/ProductScreen.esm.js` and `OrderlineDetails.xml` should go to `static/src/xml/Screens/TicketScreen/OrderlineDetails.xml`.

pos_membership_delegated_partner/static/src/xml/Orderline.xml
```diff
@@ -0,0 +1,42 @@
...
+                        />
+                        <t t-out="props.line.order.get_partner().name" />
+                        <span> (<i
+                                class="fa fa-user"
+                                role="img"
+                                aria-label="Customer"
+                                title="Customer"
+                            />)</span>
+                    </t>
+                </t>
```
huguesdk: instead of having a big `t-if`/`t-else` with almost the same thing in both parts, the orderline should have a method returning the membership partner (with the logic there). the display of the customer icon could be in an `t-if="!props.line.delegated_member"`.

pos_membership_delegated_partner/static/src/xml/OrderlineDetails.xml
```diff
@@ -0,0 +1,42 @@
...
+                            title="Member"
+                        />
+                        <t t-out="props.line.order.get_partner().name" />
+                        <span> (<i
+                                class="fa fa-user"
+                                role="img"
+                                aria-label="Customer"
+                                title="Customer"
+                            />)</span>
+                    </t>
```
huguesdk: same remark as in the `Orderline` template. better: as these have exactly the same content, this should additionally be defined in a component reused in both places.

huguesdk (changes requested):
some remarks, mostly cosmetic. otherwise, lgtm.

flaenen:
@huguesdk Thanks for the review! I've made the adjustments.

huguesdk (approved):
thanks for the changes! lgtm! 🎉

PR #1376, opened 2025-05-06
[16.0][IMP] pos_membership: highlight membership status

flaenen:
The "Current Membership Status" field is highlighted :
*  It is displayed in green if it matches "Invoiced Member", "Free Member" or "Paid Member" and in red if it matches "Cancelled Member", "Old Member" or "Waiting Member".
*  It is added to the POS product screen.


harry-ngn (approved):
Functional test successful

pos_membership/static/src/xml/PartnerLine.xml
```diff
@@ -15,7 +15,7 @@
         <xpath expr="//button[hasclass('edit-partner-button')]" position="after">
             <span
                 t-if="props.partner.membership_state != 'none'"
-                class="membership-state-tag"
+                t-attf-class="membership-state-tag {{ props.partner.is_member ? 'membership-green' : 'membership-red' }}"
```
huguesdk, line 18: this could also be written as (see [owl’s doc about dynamic class attributes](https://github.com/odoo/owl/blob/master/doc/reference/templates.md#dynamic-class-attribute)):
```suggestion
                class="membership-state-tag"
                t-att-class="{'membership-green': props.partner.is_member, 'membership-red': !props.partner.is_member}"
```
but it is more verbose and doesn't allow for an “else” clause, which forces the condition to be evaluated twice.

huguesdk (approved):
lgtm

pos_membership/static/src/xml/PartnerLine.xml
flaenen: @huguesdk I've just made an update with another proposal. What do you think?

remytms: Juste to notice that there is a fix to do in [this module](https://github.com/coopiteasy/cie-custom/tree/16.0/lpcr_pos_membership), when this change will be validated.

remytms: @flaenen I find the proposal of huguesdk more readable, than the `and/or` solution.

pos_membership/static/src/xml/PartnerLine.xml
```diff
@@ -16,6 +16,12 @@
             <span
                 t-if="props.partner.membership_state != 'none'"
                 class="membership-state-tag"
+                t-attf-class="
+                {{
+                    props.partner.is_member
+                    and 'membership-green'
+                    or 'membership-red'
+                }}"
             >
```
remytms: It can be good to add a `name` attribute to this element, so that we can easily target it when creating extended modules.

pos_membership/static/src/xml/PartnerLine.xml
huguesdk: @flaenen i actually prefer how it was written originally (with the ternary operator), as it is more clear that it selects a value instead of using the side-effects of boolean operators. this is how it was approved and is now used in production on our side.

moreover, the owl documentation states that `class` and `t-att-class` can be used together, but it doesn’t state the same for `t-attf-class`.

would you please revert to how it was before (in commit 57887c283bb07f7127ee7ab833013e3f5c8c03de) (except for the changes in the documentation)?

flaenen: @huguesdk Done

huguesdk: many thanks!

polchampion:
@OCA/pos-maintainers hi, could you take a look and merge this? thanks a lot

PR #1384, opened 2025-05-16
[FIX] pos_display_*: refactor to show extra info only in Product screen

trisdoan:
### This change
- Before this commit, extra info were shown on Receipt,... which is prohibited in some countries
- Beside, template customization was not versatile (replace standard with a wrapper div), comparing to new approach:
   - Take advantage of `t-set-slot='details'` from standard to add extra info
   - Ensure the extra info is displayed on Product screen, not Receipt/Ticket/Customer display
   - Side-effect: don't need to add `usePos` service

#### Before
before_pos

#### After
after_pos

#### Receipt 
Odoo-POS


sebduch73:
Hello.
I have test the modules and it's ok.
@ivantodorovich can you validate the PR to OCA ? thx a lot

ivantodorovich (approved):
Looks good ! Thanks!

sebduch73 (approved):

etobella (approved):
LGTM Technically

/ocabot merge nobump

Issue #1325, opened 2025-03-03
pos_payment_change : payment_date


jguenat:
Hello,

In module pos_payment_change, the field payment_date (Datetime) is set as shown bellow when changing a payment. 

https://github.com/OCA/pos/blob/df170b59eae65558c05f6372bf381b9e7198c9e8/pos_payment_change/wizards/pos_payment_change_wizard.py#L93

I have some issues with this :
- the field payment_date is of type Datetime but is assigned a Date -> New payment will be at a strange hour (01:00 in the morning).
- If you change a payment a few days after the order, e.g. when closing the pos session, then the new payment_date will be the day you do the correction which does not reflect reality.

I think it would be best to use the payment_date of the original payment or better to let the user choose the date and time in the wizard (with a default at the order creation date/time).

Do you know why it was done that way originally?

Depending on feedback I could propose an improvement.

Thank you

legalsylvain:
> I think it would be best to use the payment_date of the original payment or better to let the user choose the date and time in the wizard (with a default at the order creation date/time).

Totally agree. @quentinDupont what do you think ?

pedrobaeza:
Was this tackled?

jguenat:
yes here #1387

pedrobaeza:
OK, closing the issue and continuing through the PR.

PR #1387, opened 2025-05-19
[14.0][IMP] pos_payment_change: Datetime management

jguenat:
I found that order/payment date had some strange behavior when modifying a payment, see #1325 

Also fix a bug where the resale order was not at the same time as the refund order in case of refund policy.

In case of update policy keep payment_date the same
In case of refund policy use current datetime for refund/resale order and payments


pedrobaeza (approved):

legalsylvain (approved):
Thanks !

/ocabot merge patch

PR #1388, opened 2025-05-20
[ADD]17.0: Added pos_product_info_location module.

parvezqureshi:

PR #1385, opened 2025-05-16
[17.0][MIG] pos_stock_available_online

edescalona:
@BinhexTeam

Standard migration to 17.

PR #1389, opened 2025-05-21
[ADÐ] 16.0: pos_product_stock_location_hide_no_qty_available

benwillig:
This modules allows to remove stockable products which are not available on configured warehouse.


cvinh:
Hello
A functional question : why didn't you consider the stock location in pos config operation type ?
AFAIK the stock moves of sold products will be taken from it

benwillig:
Thanks for your comment. To be honest, I didn't think enough. We thought that the user would select the corresponding warehouse if the feature is enabled. But after more thinking, I will change this to select stock locations, but using by default the default source location of the picking type if the feature is enabled.

cvinh:
cc @vehi-invitu what do you think ?

ThomasBinsfeld (approved):

cvinh:
Did you check this module https://github.com/OCA/pos/pull/1385 ?
And this one https://github.com/OCA/pos/pull/1388 ?
they have different approaches but I'm wondering if they can be merge into one module and pos config settings allow to choose one or another approch...
WDYT ?

benwillig:
Yes we saw the first one but it didn't meet our customer requirements. Also, the notifications are a bit intrusive and as we don't need it, I prefer to not install it. I think that both could be used in the same instance. But yes, the module we propose is a first try and can evolve if needed.

PR #1323, opened 2025-02-28
[16.0][MIG] pos_tare: migration to 16.0

huguesdk:
this is a follow-up on #1224. it contains all of its commits (as it’s the same branch), as well as several commits to restore features present in version 12.0 but that disappeared during the migration to 14.0, with also some fixes and improvements.

PR #1397, opened 2025-06-08
[16.0][ADD] self-service weighing station modules

huguesdk:
add a collection of modules to configure a pos as a self-service weighing station:

*   `base_pos_self_service_weighing`: base module providing the ui and the common functionalities; not useful on its own
*   `pos_self_service_weighing_product`: allows customers to weigh products and print labels with a barcode identifying the product and containing its weight or its price
*   `pos_self_service_weighing_tare`: allows customers to weigh containers and print labels with a barcode containing the weight of the container (tare weight) (labels usable with the `pos_tare` module)
*   `pos_self_service_weighing_product_tare`: glue module between `pos_self_service_weighing_product` and `pos_tare`, allowing customers to take the weight of containers into account to compute the net weight when weighing products
*   `pos_self_service_weighing_print_zpl`: allows to print barcode labels using zpl-compatible printers

this pull request depends on #1323.


legalsylvain:
base_pos_self_service_weighing could be named simply pos_self_service_weighing

huguesdk:
@legalsylvain i’ve followed the second item in [the modules section of the oca guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules). did these change?

legalsylvain:
@legalsylvain i’ve followed the second item in [the modules section of the oca guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules). did these change?

Hi. I didn't Know that rule. I dont like it cause that prefix breaks the order. 

But well.

mihien:
LGTM. Only thing I would change is the welcome screen, I think it would benefit from making it more obvious that it is a welcome screen. Just to make sure that a customer seeing it for the first time would know what he's seeing and what he needs to do, since it's self-service. Changing the button text closer to something like "Press here to start weighing products" would already be enough, or adding a welcome message above the button.

huguesdk:
@OCA/pos-maintainers as this is quite a big amount of code, could the “stale” label please be removed and the “no stale” label set?

mihien (approved):
LGTM !

ivs-cetmix:
Hey @huguesdk thank you for your contribution!  Shouldn't we migrate the modules one by one? Such huge piece of code is not so easy for review.

huguesdk:
hello @ivs-cetmix. indeed, that’s a lot to review. the thing here is that it is a mix of porting, refactoring and new features. in 12, there were only 2 (non-oca) modules:

1.  `pos_self_service_base`: allows to turn a pos into a self-service tare weighing station and handles barcode generation.
2.  `pos_self_service_print_zpl`: allows to print the barcode with zpl-compatible printers through ipp

here, the base module was stripped down to be more generic, so that it can be used to weigh products, containers (tare) or both. the other modules implement this (with some code coming from the 1st module in 12). `pos_self_service_weighing_print_zpl` is almost equivalent to the 2nd module in 12, but has been improved.

the self-service weighing of products (with optional tare removal) is a completely new feature.

so i think we can consider these as new modules, and as they depend very much on each other, i think that having them bundled all together makes sense. it makes it possible to test them more easily and changes would be easier too.

base_pos_self_service_weighing/README.rst
```diff
@@ -0,0 +1,141 @@
...
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/pos&target_branch=16.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+Base module to configure a PoS as a self-service weighing station.
+
+This module allows to configure a PoS to be a self-service weighing station, where customers can weigh things by themselves.
```
ferdymercury, line 33:
```suggestion
This module allows to configure a PoS to be a self-service weighing station, where customers can see the weight of the products (or the tare of containers) they place.
See optional combination with pos_self_service_weighing_print_zpl to print a barcode sticker encoding that weight for later reading it in the actual PoS.
```

base_pos_self_service_weighing/README.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Setup the self-service weighing station
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+1. Create a new PoS
+2. Enable "Is a Self-Service Weighing Station"
+3. Setup hardware proxy
+
+   * Enable electronic scale
+4. Enable barcode reader and its nomenclature
+
```
ferdymercury, line 58: It's not clear why one would need here a barcode reader? shouldn't it say here rather the optional barcode printer of the pos_self_service_weighing_print_zpl?

base_pos_self_service_weighing/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+Setup the self-service weighing station
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+1. Create a new PoS
+2. Enable "Is a Self-Service Weighing Station"
+3. Setup hardware proxy
+
+   * Enable electronic scale
+4. Enable barcode reader and its nomenclature
```
ferdymercury, line 8: printer?

base_pos_self_service_weighing/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+Base module to configure a PoS as a self-service weighing station.
+
+This module allows to configure a PoS to be a self-service weighing station, where customers can weigh things by themselves.
+
```
ferdymercury, line 4: same comment / suggestion of changes as above

pos_self_service_weighing_tare/static/src/js/models.esm.js
```diff
@@ -0,0 +1,81 @@
...
+            );
+            this._tare_uom = this.units_by_id[this.tare_uom_id];
+            this._tare_decimal_places = this.convert_uom_decimal_places(
+                this.tare_barcode_decimal_places,
+                this.tare_barcode_uom,
+                this._tare_uom
+            );
+        }
+
+        generate_tare_barcode(weight) {
```
ferdymercury, line 30: not sure if it's easy, but it would nice-to-have if, when generating the barcode with the bars and the EAN13 number below in small as text, the part of the EAN13 that represents the weight appeared in bold face. Makes it easy to more human-readable in case the barcode reader is broken and the PoS needs to enter the weight of the tare manually.

ferdymercury (approved):
Awesome development! Thanks a lot.

I have some minor questions / suggestions related to the documentation

base_pos_self_service_weighing/README.rst
huguesdk: while potentially useful, that is a use-case that wasn’t thought about. this module on its own doesn’t provide a logical feature set: it will always display a “print” button, but clicking it would have no effect, which would be confusing. most weighing scales have a display displaying the weight, so setting up a computer just for this doesn’t seem likely.

about the second line, the text below already references the other useful modules. installing these directly will install this one.

base_pos_self_service_weighing/README.rst
huguesdk: indeed, this seems like a leftover from the modules in 12.0, where the barcode reader needed to be enabled for the barcode nomenclature to be recognized. in 16.0, it seems to be always enabled.

base_pos_self_service_weighing/readme/CONFIGURE.rst
huguesdk: indeed, but with this module alone it is not possible. this should be covered by the readme of `pos_self_service_weighing_print_zpl`.

base_pos_self_service_weighing/readme/DESCRIPTION.rst
huguesdk: please note that `README.rst` and `static/description/index.html` are both automatically generated from the files in the `readme` folder, so it is easier to review these only in one place.

pos_self_service_weighing_tare/static/src/js/models.esm.js
huguesdk: the `pos_self_service_weighing_print_zpl` module prints the barcode with a title above it (which is the description, for example the product name or “Tare”), and a value beneath it (the weight or the price), which corresponds to what is encoded in the barcode. this way it is already readable by humans. ☺️

huguesdk (commented):
@ferdymercury hey, thanks for the review!

ferdymercury:
gentle ping @ivantodorovich
(thanks in advance)

PR #1400, opened 2025-06-19
[14.0][FIX] pos_payment_terminal: read payment method ids fron config

sbidoul:
Otherwise, it gets all payment methods even those that are not configured for the pos session, which leads to trying to use the proxy even if no payment mode that needs it is configured.


legalsylvain (approved):
Make sense.

Still draft ?

sbidoul:
@xavier-bouquiaux is going to test it.

legalsylvain:
Hi @sbidoul and @xavier-bouquiaux. Any new on that PR ? 

thanks !

PR #1403, opened 2025-06-30
[18.0][MIG] pos_product_multi_barcode

thibaultrey:
Migration of module pos_product_multi_barcode to version 18.0


pos_product_multi_barcode/static/src/js/overrides/screens/product_screen/product_screen.esm.js
```diff
@@ -0,0 +1,13 @@
+import {patch} from "@web/core/utils/patch";
+import {ProductScreen} from "@point_of_sale/app/screens/product_screen/product_screen";
+
+patch(ProductScreen.prototype, {
+    loadProductFromDBDomain(searchProductWord) {
+        const domain = super.loadProductFromDBDomain(searchProductWord);
+        /* Add a condition to search by barcodes_json */
+        if (searchProductWord) {
+            domain.push(["barcode_ids.name", "in", [searchProductWord]]);
+        }
```
Sibert-Aerts, line 10: Hi, thank you for volunteering to upgrade this module. When I tried using it this override however caused some incorrect behavior, as the extra condition it appends to the domain makes it so ONLY products with a barcode matching the search term are shown.

I fixed this by modifying the domain in the following way:

```js
// Turn every ("barcode", "ilike", term) leaf into
// ["|", ("barcode", "ilike", term), ("barcode_ids.name", "ilike", term)]
for (let i = 0; i < domain.length; i++) {
    const leaf = domain[i];
    if (leaf instanceof Array && leaf[0] === "barcode" && leaf[1] === "ilike") {
        domain.splice(i, 0, ["barcode_ids.name", "ilike", leaf[2]]);
        domain.splice(i, 0, "|");
        i += 2;
    }
}
```

PR #1404, opened 2025-07-01
[16.0][ADD] pos_hide_receipt_line

adasatorres:
Good morning, this add-on allows you to hide product lines on the point of sale receipt, as long as the line's price is 0.

To do this, go to the product in the backend and enable the Hide Receipt Line field, which is only visible if the product is available in the point of sale. After that, simply go to the point of sale and create a line with the product at a price of 0.


antoniodavid (approved):
LGTM

edescalona (approved):
LGTM

FrancoMaxime (approved):

pos_hide_receipt_line/tests/test_pos_hide_receipt_line.py
```diff
@@ -0,0 +1,31 @@
...
+        cls.product = cls.ProductTemplate.create(
+            {
+                "name": "Test Product",
+                "available_in_pos": True,
+                "pos_hide_receipt_line": True,
+            }
+        )
+
+    def test_hide_receipt_line(self):
+        self.assertTrue(self.product.pos_hide_receipt_line)
```
ivantodorovich, line 27: This test doesn't add any value. It's just testing the ORM, and we don't have to do that.

What would be useful is to test this module with a js tour that makes sure the receipt line is not actually displayed.

ivantodorovich (commented):
Thanks for your contribution!

I just have a comment regarding the unit tests

PR #1405, opened 2025-07-06
[18.0][ADD] pos_financial_surcharge: module financial surcharge in pos for cards

fsuleWitdata:
This module enables custom financial surcharges based on the installment plan selected by the customer when paying by card in the Odoo 18 Point of Sale (POS). It is especially useful for credit card transactions where the merchant absorbs a bank interest and wants to pass it on to the consumer.


legalsylvain (commented):
Files are missing

PR #1413, opened 2025-07-14
[18.0][MIG] pos_customer_required: Migration to 18.0

P-H-Phuc:
This module was written to extend the functionality of odoo pos and allows you to require a customer for each pos order. In the pos session configuration, you can choose to require the customer for pos orders.

If a customer is not selected, the pos ui will display an message.
In the backend the customer field is required when needed.

Two new options are available:

- Customer 'Required before starting the order';
- Customer 'Required before paying';

## 'Required before starting the order' Option

In the frontend PoS, the default screen is the screen to select customers.

- Users are not allowed to start selling before having selected a customer;

pos_customer_required_before_order

## 'Required before paying' Option

In the frontend PoS, the user can start selling, but if the user tries to make payment and if a customer is not selected, the pos ui will display an message.

pos_customer_before_paying


ivantodorovich (approved):
/ocabot migration pos_customer_required

flotho:
Hi, 
Thanks for your PR @P-H-Phuc 
could you please rebase to trigger the CI/CD

PR #1415, opened 2025-07-16
[MIG] pos disable pricelist selection 17.0

kopeyev:


legalsylvain (changes requested):
Hi. Thanks for migrating the module.

- could you take a look on other implémentation https://github.com/OCA/pos/tree/16.0/pos_pricelist_technical#readme
Thanks !

ilmir-k:
> Hi. Thanks for migrating the module.
> 
>     * could you take a look on other implémentation https://github.com/OCA/pos/tree/16.0/pos_pricelist_technical#readme
>       Thanks !

Hi @legalsylvain If you mean here that the module you mentioned replaces `pos_disable_pricelist_selection`, it's actually not. For example, at least it doesn't have **Hide pricelist button** option.

legalsylvain:
Sorry. I didnt understood your use case. Could you elaborate why you want to hide the button, but why do you load the pricelists in the PoS ?

Thanks !

ilmir-k:
> Sorry. I didnt understood your use case. Could you elaborate why you want to hide the button, but why do you load the pricelists in the PoS ?
> 
> Thanks !

The pricelist itself should be applied in POS e.g. if we use it for some of the customer but the cashiers must not be able to change it.

PR #1416, opened 2025-07-21
[16.0][IMP] pos_default_partner: Set default partner on pos generated moves

sergiobstoj:
Assigns the default partner to the Entries

### DIFF
--- a/pos_default_partner/models/pos_payment.py
+++ b/pos_default_partner/models/pos_payment.py
@@ -1,0 +1,20 @@
+from odoo import models
+
+
+class PosPayment(models.Model):
+    _inherit = "pos.payment"
+
+    def _generate_payment_move(self, is_reverse, change_payment=None):
+        order = self.pos_order_id
+        pos_session = order.session_id
+        default_partner_id = pos_session.config_id.default_partner_id
+        return super(
+            PosPayment, self.with_context(pos_default_partner_id=default_partner_id)
+        )._generate_payment_move(is_reverse, change_payment=change_payment)
+
+    def _prepare_debit_line_payment(self, payment_move, is_reverse):
+        vals = super()._prepare_debit_line_payment(payment_move, is_reverse)
+        if self.env.context.get("pos_default_partner_id", False):
+            default_partner_id = self.env.context["pos_default_partner_id"].id
+            vals["partner_id"] = default_partner_id
+        return vals
--- a/pos_default_partner/README.rst
+++ b/pos_default_partner/README.rst
@@ -74,6 +74,7 @@ Contributors
 * FactorLibre:
 
   * Alvaro Rollan
+  * Sergio Bustamante
 * `Coop IT Easy SC <https://coopiteasy.be>`_:
 
   * hugues de keyzer
--- a/pos_default_partner/models/__init__.py
+++ b/pos_default_partner/models/__init__.py
@@ -1,3 +1,6 @@
 from . import pos_order
 from . import pos_config
 from . import res_config_settings
+from . import account_move
+from . import pos_payment
+from . import pos_session
--- a/pos_default_partner/models/account_move.py
+++ b/pos_default_partner/models/account_move.py
@@ -1,0 +1,14 @@
+from odoo import api, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self.env.context.get("pos_default_partner_id", False):
+            default_partner_id = self.env.context["pos_default_partner_id"].id
+            for vals in vals_list:
+                if not vals.get("partner_id", False):
+                    vals["partner_id"] = default_partner_id
+        return super().create(vals_list)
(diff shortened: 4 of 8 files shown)


aritzolea (approved):

sergiobstoj:
@OCA/pos-maintainers can this be reviewed?

pos_default_partner/models/pos_payment.py
```diff
@@ -0,0 +1,20 @@
+from odoo import models
+
+
+class PosPayment(models.Model):
+    _inherit = "pos.payment"
+
+    def _generate_payment_move(self, is_reverse, change_payment=None):
```
ivantodorovich, line 7: I'm not able to find this method in core's `16.0` branch 🤔 
Are you sure it's correct?

pos_default_partner/models/pos_payment.py
```diff
@@ -0,0 +1,20 @@
...
+        order = self.pos_order_id
+        pos_session = order.session_id
+        default_partner_id = pos_session.config_id.default_partner_id
+        return super(
+            PosPayment, self.with_context(pos_default_partner_id=default_partner_id)
+        )._generate_payment_move(is_reverse, change_payment=change_payment)
+
+    def _prepare_debit_line_payment(self, payment_move, is_reverse):
```
ivantodorovich, line 15: Same here

ivantodorovich (changes requested):
Hello 👋🏻 

It seems this module's unit tests are not being executed in the CI.
It's failing silently with this message:

```
2025-07-21 08:55:44,408 470 INFO odoo odoo.tests.suite: skipped setUpClass (odoo.addons.pos_default_partner.tests.test_pos_default_partner.TestPosDefaultPartner) : Accounting Tests skipped because the user's company has no chart of accounts. 
```

What triggered me is the fact that I couldn't find in core the methods you're overriding.

Any chance you can fix that here?



----


Moreover, from a functional point of view, I don't think I agree with this change.
The `pos_default_partner` module does exactly what it says: sets a default partner for orders. However, it's only a default value.. other partners can be chosen per different orders.

Then, having this passed onto the single account.move wouldn't be correct IMO.

I would, however, ask for input from the module's author: @bealdav

sergiobstoj:
Hi, @ivantodorovich 
First of all, thanks for your review.

Having read your comments, ¿it could be an idea to make it configurable what I am proposing in here? ¿Or even including it in a new addon?
Overall I believe it's a good improvement.

About the tests, I added the fix for loading a CoA.
Greetings.

pos_default_partner/models/pos_payment.py
sergiobstoj: You cand find that here: https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/models/pos_payment.py#L99

pos_default_partner/models/pos_payment.py
sergiobstoj: And this one here: https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/models/pos_payment.py#L133

pos_default_partner/models/pos_payment.py
ivantodorovich: Apologies, it seems I was looking into a really old version of the `16.0` branch, before https://github.com/odoo/odoo/pull/140605 landed

ivantodorovich:
> Having read your comments, ¿it could be an idea to make it configurable what I am proposing in here? ¿Or even including it in a new addon?
> Overall I believe it's a good improvement.

I think both options are good ideas.
Personally I wouldn't oppose to it being configurable. Module authors' have the last word about such topics, so we'll give them some time to catch up and review.

> About the tests, I added the fix for loading a CoA.

Hey, thanks for giving it a try! It seems they're still not being executed, though.

We have an issue with our CI, that gives a false positive ✅  in such cases (tracked here: https://github.com/OCA/oca-ci/issues/105)

🧠 How to tell if they're being executed, then?

1️⃣ Well, the first and most obvious one is the coverage report:
- It's red
- Access the codecov report and shows no tests were executed: https://app.codecov.io/gh/OCA/pos/pull/1416?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=OCA

2️⃣ Then, there's also this log line in the build logs:
```
2025-08-22 05:43:43,275 470 INFO odoo odoo.tests.suite: skipped setUpClass (odoo.addons.pos_default_partner.tests.test_pos_default_partner.TestPosDefaultPartner) : Accounting Tests skipped because the user's company has no chart of accounts. 
```

bealdav:
Hi @ivantodorovich  thanks to ping me. 

FYI information I didn't use this module. I prefer to be away from POS projects.
Particularly point_of_sale module which duplicate code, and workflows to make simple things complex and confusing for users i.e. pos.order vs sale.order and then all implications with update/cancellation. In 95% of the use of pos in real life, you don't need to be disconnected, then why this complexity ?

Then I'll let main contributors of this module replace me as author.

I'm sure your options will be more consistent than mine.

Thanks a lot.

PR #1419, opened 2025-07-24
[17.0][MIG] pos_disable_pricelist_selection: Migration to 17.0

kopeyev:
Migration done according to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0


legalsylvain:
Same remark as in https://github.com/OCA/pos/pull/1415#issuecomment-3104416892

PR #1423, opened 2025-08-04
[16.0][MIG] pos customer required: Migration to 16.0

mihien:


pos_customer_required/__manifest__.py
```diff
@@ -17,9 +18,20 @@
...
+    "assets": {
+        "point_of_sale.assets": [
+            "pos_customer_required/static/src/js/"
+            "Screens/PartnerListScreen/PartnerListScreen.esm.js",
+            "pos_customer_required/static/src/js/"
+            "Screens/PaymentScreen/PaymentScreen.esm.js",
+            "pos_customer_required/static/src/js/"
+            "Screens/ProductScreen/ProductScreen.esm.js",
+            "pos_customer_required/static/src/xml/"
+            "Screens/PartnerListScreen/PartnerListScreen.xml",
```
huguesdk: this is enough:

```
"pos_customer_required/static/src/js/**/*.js",
"pos_customer_required/static/src/xml/**/*.xml",
```

pos_customer_required/views/pos_config_view.xml
huguesdk, line 1: as the setting is now available from the general configuration, is it still needed here?

pos_customer_required/static/src/xml/Screens/PartnerListScreen/PartnerListScreen.xml
```diff
@@ -0,0 +1,15 @@
...
+<templates id="template" xml:space="preserve">
+    <t
+        t-name="PartnerListScreenNoCreate"
+        t-inherit="point_of_sale.PartnerListScreen"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//div[hasclass('back')]" position="attributes">
+            <attribute name="class">oe_hidden</attribute>
+        </xpath>
```
huguesdk, line 17: why is this needed? a comment would help.

pos_customer_required/static/src/js/Screens/ProductScreen/ProductScreen.esm.js
```diff
@@ -0,0 +1,28 @@
...
+import Registries from "point_of_sale.Registries";
+
+const PosRequiredCustomerProductScreen = (OriginalProductScreen) =>
+    class extends OriginalProductScreen {
+        setup() {
+            super.setup(...arguments);
+        }
+
+        async onMounted() {
+            super.onMounted(...arguments);
```
huguesdk: `await` missing.

pos_customer_required/static/src/js/Screens/PartnerListScreen/PartnerListScreen.esm.js
```diff
@@ -0,0 +1,27 @@
...
+/* Copyright CoopITEasy - Simon Hick <sim@coopiteasy.be>
+   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl) */
+
+import PartnerListScreen from "point_of_sale.PartnerListScreen";
+import Registries from "point_of_sale.Registries";
+
+const PosCustomerRequiredPartnerListScreen = (OriginalPartnerListScreen) =>
+    class extends OriginalPartnerListScreen {
+        clickPartner(partner) {
+            if( this.env.pos.config.require_customer === "order" ) {
```
huguesdk: spacing: should be `if (condition) {` instead of `if( condition ) }`. `if` is not a function, and there should never be spaces after an opening parenthesis or before a closing one.

pos_customer_required/static/src/js/Screens/PartnerListScreen/PartnerListScreen.esm.js
```diff
@@ -0,0 +1,27 @@
...
+
+/* Copyright CoopITEasy - Simon Hick <sim@coopiteasy.be>
+   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl) */
+
+import PartnerListScreen from "point_of_sale.PartnerListScreen";
+import Registries from "point_of_sale.Registries";
+
+const PosCustomerRequiredPartnerListScreen = (OriginalPartnerListScreen) =>
+    class extends OriginalPartnerListScreen {
+        clickPartner(partner) {
```
huguesdk: a comment would help to understand why this is needed. i looked at the original function and if i understand it correctly, the goal is to prevent to call `confirm()` if the partner is unselected, right? would it work by overriding `confirm()` instead to avoid re-implementing the logic?

pos_customer_required/static/src/js/Screens/PartnerListScreen/PartnerListScreen.esm.js
```diff
@@ -0,0 +1,27 @@
...
+            if( this.env.pos.config.require_customer === "order" ) {
+                if (this.state.selectedPartner && this.state.selectedPartner.id === partner.id) {
+                    this.state.selectedPartner = null;
+                    this.render(true);
```
huguesdk: this should not be needed, should it?

pos_customer_required/static/src/js/Screens/PaymentScreen/PaymentScreen.esm.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        async _isOrderValid(isForceValidate) {
+            if (
+                this.env.pos.config.require_customer !== "no" &&
+                !this.env.pos.get_order().get_partner()
+            ) {
+                const result = await this.showPopup("ConfirmPopup", {
+                    title: _t("An anonymous order cannot be confirmed"),
+                    body: _t("Please select a customer for this order."),
```
huguesdk: use `this.env._t()` instead of `_t()`.

huguesdk (changes requested):
nice conversion to owl, @mihien! a few remarks, but mostly lgtm.

if i understand correctly, you added the improvement of preventing the user to close or confirm the partner selection screen if no partner has been selected, right? nice addition! does it also work from the payment screen when `require_customer` is set to `payment` (or should it not work at that point)?

pos_customer_required/static/src/js/Screens/PartnerListScreen/PartnerListScreen.esm.js
mihien: Both your assumptions are correct, I've moved the change to `confirm()` (and added a comment)

pos_customer_required/static/src/js/Screens/PartnerListScreen/PartnerListScreen.esm.js
mihien: `this.render(true);` was needed to reload the screen to show the user had been unselected. But now that the logic is moved to confirm() `this.state.selectedPartner = null;` will be handled in clickPartner already, so I've removed it.

pos_customer_required/views/pos_config_view.xml
mihien: Good question, it's probably not but I'll ask an analyst what he thinks

mihien:
@huguesdk I've applied all you suggestions and added a test to increase coverage. The only thing left is this remark:

> if i understand correctly, you added the improvement of preventing the user to close or confirm the partner selection screen if no partner has been selected, right? nice addition! does it also work from the payment screen when `require_customer` is set to `payment` (or should it not work at that point)?

It doesn't work when `require_customer` is set to `payment` and the changes to make it so aren't that obvious to me. So I've left it as is, since it's also how it works in 12.0

huguesdk (commented):
i’ve just functionally tested the module, and the fact that it is not possible to close the partner selection screen is actually annoying and confusing (in my opinion), especially when the mode is not `order`: if one clicks on the customer selection button and doesn’t want to select a customer, the only way to go back is to select one, then re-opening the same screen and unselect the customer. i think that the button should stay, and a pop-up should appear in case a customer is mandatory.

an opinion from the users of the module in previous versions would be useful.

mihien:
> i’ve just functionally tested the module, and the fact that it is not possible to close the partner selection screen is actually annoying and confusing (in my opinion), especially when the mode is not `order`: if one clicks on the customer selection button and doesn’t want to select a customer, the only way to go back is to select one, then re-opening the same screen and unselect the customer. i think that the button should stay, and a pop-up should appear in case a customer is mandatory.
> 
> an opinion from the users of the module in previous versions would be useful.

Good points. I've changed it so that the discard button is only removed when the setting is set to `order`.

pos_customer_required/static/src/xml/Screens/PartnerListScreen/PartnerListScreen.xml
```diff
@@ -0,0 +1,19 @@
...
+        t-inherit="point_of_sale.PartnerListScreen"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <!-- Removes the ability to go back, forcing the user to select a partner
+             to exit the screen when the config has been set to 'order'.  -->
+        <xpath expr="//div[hasclass('back')]" position="attributes">
+            <attribute name="t-attf-class">
+                back {{env.pos.config.require_customer=='order' ? 'oe_hidden' : ''}}
+            </attribute>
```
huguesdk: please use the cleaner `t-att-class` attribute (as explained [here](https://github.com/odoo/owl/blob/master/doc/reference/templates.md#dynamic-class-attribute)):


```suggestion
            <attribute name="t-att-class">
                {"oe_hidden": env.pos.config.require_customer == "order"}
            </attribute>
```

huguesdk (changes requested):
just a remark about the attribute syntax, otherwise lgtm.

huguesdk (approved):
lgtm, thanks!

huguesdk:
@OCA/pos-maintainers could someone please merge this?

PR #1425, opened 2025-08-07
[16.0][IMP] pos_session_pay_invoice: Restrict to journals of the current POS session

carlos-lopez-tecnativa:
To prevent mistakes, users must select only the journals available in the pos.payment.method of the POS session.

TT57575

@Tecnativa @pedrobaeza could you please review this?

### DIFF
--- a/pos_session_pay_invoice/models/account_journal.py
+++ b/pos_session_pay_invoice/models/account_journal.py
@@ -1,0 +1,13 @@
+from odoo import api, models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
+        if self.env.context.get("pos_session_id"):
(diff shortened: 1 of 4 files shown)


pos_session_pay_invoice/models/account_journal.py
```diff
@@ -0,0 +1,13 @@
+from odoo import api, models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
```
pedrobaeza, line 8: I think this shouldn't be done at this level, but modifying the domain of the field `journal_id` in the wizard.

pedrobaeza: Seen that this is working even if no typing.

pedrobaeza (approved):
OK, let's apply the hotfix for the changes done yesterday.

/ocabot merge patch

PR #1426, opened 2025-08-07
[16.0][IMP] pos_session_pay_invoice: Generate pos.order and pos.payment for customer transactions

carlos-lopez-tecnativa:
Before this commit, the wizard generated an `account.bank.statement.line`, but the problem was that when the POS had cash control enabled, all transactions generated by this wizard were displayed under Cash, because Odoo considered such statements as Cash In/Out operations, see https://github.com/odoo/odoo/blob/c1d3153ed564c0f47a201e5751740857a15d7db2/addons/point_of_sale/models/pos_session.py#L557-L567.

Now, the transactions are shown under Cash or Bank depending on the payment method and journal type.

For payments of vendor bills, a statement line will still be created, because the POS only works with receivable accounts and cannot reconcile with payable accounts. See https://github.com/odoo/odoo/blob/c1d3153ed564c0f47a201e5751740857a15d7db2/addons/point_of_sale/models/pos_order.py#L890-L898

TT57575
@Tecnativa @pedrobaeza could you review please?


pedrobaeza (approved):

victoralmau (approved):

chienandalu (approved):
/ocabot merge patch

PR #1432, opened 2025-09-06
[18.0][ADD] pos_order_datepicker: add datepicker filter in POS order list

rahulpatel333:
Summary: Introduces datepicker filter in POS order list.
Type: [ADD]
Tested on: Odoo 18 CE.
Screenshot: 
screenshot


adasatorres (approved):
LGTM.

Code review and functional review in the runboat, the only thing missing is for the tests not to fail.

jelenapoblet (approved):
I tested it through runboat and it works as spected, simple yet usefull. Works as expected.

jelenapoblet:
I noticed that the README file doesn’t fully follow the recommended structure. You can check the guide here: https://odoo-community.org/readme-structure
As a note, if you follow the instructions in that guide, there’s a built-in pre-commit script in the repository that will automatically generate the README in the correct format.

adasatorres:
Good evening @rahulpatel333 , I recommend squashing all commits. This is generally required to maintain a clean and consistent commit history. Please refer to the following link, which provides a clear explanation of the process:
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

adasatorres (changes requested):
It is not necessary to remove the other addons.

rahulpatel333:
> It is not necessary to remove the other addons.


My module is completely separate from the other modules in the repository. Since their test cases were failing, my module’s tests were also being marked as failed. That’s why I removed the other modules to ensure my module’s tests could pass independently.

adasatorres:
> My module is completely separate from the other modules in the repository. Since their test cases were failing, my module’s tests were also being marked as failed. That’s why I removed the other modules to ensure my module’s tests could pass independently.

The solution should not be to remove the addons, as they would be deleted from the repository if the PR is merged. The correct approach would be to open a PR to fix the tests for each addon. Once the fixes are merged into the 18.0 branch in this case, you should perform a rebase and bring the changes from the repository into the branch you are using for the PR.

PR #1437, opened 2025-09-22
[17.0][ADD] pos_early_receipt_printing

angelgarciadelachica:
This module allows you to display the option to print bills from the POS even if it is not a restaurant type.

[T-8910]


Jaimermaccione (approved):
@angelgarciadelachica LGTM! I tested the module on local environment and it seems to work correctly.

I tried enabling the option to print tickets in the POS configuration and the option was properly activated.

- I also tested updating the POS settings and even upgrading the POS module itself, and in both cases the configuration was preserved.
- Everything looks fine from a functional point of view.

One minor suggestion: consider renaming the module to something like pos_early_receipt_printing, so that the purpose is clearer at first glance. Maybe @HaraldPanten can also review this and share his opinion.

HaraldPanten:
> @angelgarciadelachica LGTM! I tested the module on local environment and it seems to work correctly.
> 
> I tried enabling the option to print tickets in the POS configuration and the option was properly activated.
> 
> * I also tested updating the POS settings and even upgrading the POS module itself, and in both cases the configuration was preserved.
> * Everything looks fine from a functional point of view.
> 
> One minor suggestion: consider renaming the module to something like pos_early_receipt_printing, so that the purpose is clearer at first glance. Maybe @HaraldPanten can also review this and share his opinion.

Hi,

I completely agree with @Jaimermaccione . We should think in a better module name.

angelgarciadelachica:
> @angelgarciadelachica LGTM! I tested the module on local environment and it seems to work correctly.
> 
> I tried enabling the option to print tickets in the POS configuration and the option was properly activated.
> 
> * I also tested updating the POS settings and even upgrading the POS module itself, and in both cases the configuration was preserved.
> * Everything looks fine from a functional point of view.
> 
> One minor suggestion: consider renaming the module to something like pos_early_receipt_printing, so that the purpose is clearer at first glance. Maybe @HaraldPanten can also review this and share his opinion.

THX!

@Jaimermaccione @HaraldPanten What do you think of pos_shop_prepayment_receipt? It makes it clear that it is not for restaurants and that the receipt is generated before payment is made.

HaraldPanten:
> pos_early_receipt_printing

Sounds better, THX

@angelgarciadelachica

Jaimermaccione:
> @Jaimermaccione @HaraldPanten What do you think of pos_shop_prepayment_receipt? It makes it clear that it is not for restaurants and that the receipt is generated before payment is made.

@angelgarciadelachica ¡Great! Looks good to me. Ty!

ValentinVinagre (approved):

angelgarciadelachica:
@ivantodorovich Hi! Could you take a look at this? 
THX!

PR #1438, opened 2025-09-27
[16.0][MIG] pos_require_product_quantity: Migration to 16.0

mihien:
[Internal task T14562](https://gestion.coopiteasy.be/web#id=14554&action=475&active_id=492&model=project.task&view_type=form&menu_id=536)


polchampion (changes requested):
@mihien failed test on runboat : there are pos lines with quantity zero but no warning is shown

mihien:
> @mihien failed test on runboat : there are pos lines with quantity zero but no warning is shown

@polchampion You first have to check the option "Require Product Quantity" in the pos settings to activate it. The option was already in the settings in 12.0 but wasn't used in the module's logic. Whether you checked it or not, quantities of 0 used to result in a popup either way. If that's better, I could have it turned on by default when installing the module or if we want the same behaviour as in 12.0, I can remove the option from the settings entirely

polchampion (approved):
@mihien sucessfully tested, can you add a "configuration" section in the readme so that it's clear how it needs to be set up ?

also, why put this in the session properties and not in the general pos settings? it seems not to respect the general logic

mihien:
> @mihien sucessfully tested, can you add a "configuration" section in the readme so that it's clear how it needs to be set up ?
> 
> also, why put this in the session properties and not in the general pos settings? it seems not to respect the general logic

@polchampion I've set the setting to true by default, so that there is no need for a configuration anymore and took the opportunity to move it to the general pos settings as well.

polchampion:
> I've set the setting to true by default, so that there is no need for a configuration anymore and took the opportunity to move it to the general pos settings as well.

perfect thank you!

pos_require_product_quantity/models/res_config_settings.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 CoopITEasy - Simon Hick <sim@coopiteasy.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    require_product_quantity = fields.Boolean(
```
huguesdk: the `res.config.settings` field should be prefixed with `pos_`.

pos_require_product_quantity/static/src/js/Screens/ProductScreen/ProductScreen.esm.js
```diff
@@ -0,0 +1,41 @@
...
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+*/
+
+import ProductScreen from "point_of_sale.ProductScreen";
+import Registries from "point_of_sale.Registries";
+
+const PosRequireProductQuantityProductScreen = (OriginalProductScreen) =>
+    class extends OriginalProductScreen {
+        async _onClickPay() {
+            const result = super._onClickPay(...arguments);
```
huguesdk: i think that this should be called after the rest of the checks, because it will try to switch to the payment screen. if a check fails, it is better to not run this code at all. this way, there is even no need to switch back to the `ProductScreen` in case of error.

huguesdk (changes requested):

pos_require_product_quantity/static/src/js/Screens/ProductScreen/ProductScreen.esm.js
mihien: Would it be okay to do an empty return when a case fails and return the call to super otherwise?

huguesdk: actually, the parent method doesn’t return anything, so yes, you can do an empty return (to exit the function) after showing the popup, and otherwise simply call `super._onClickPay(...arguments);` without returning its value.

huguesdk (approved):
lgtm, thanks!

ivs-cetmix:
Hey @mihien , thank you for you contribution!
/ocabot merge nobump

PR #1442, opened 2025-10-08
[16.0][ADD] pos_margin_stored

quentinDupont:
Before v16.0, margin were handle by OCA module and stored, > v16, it's in Odoo core code
but it's not stored.
For searching bad sales, for statistics, it can be useful to have margin stored

See screenshot of pos_margin OCA module
image


pos_margin_stored/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "summary": "Margin Stored on PoS Order",
+    "version": "16.0.1.0.0",
+    "category": "Point Of Sale",
+    "author": "GRAP, Odoo Community Association (OCA)",
+    "maintainers": ["quentinDupont"],
+    "website": "https://github.com/OCA/pos",
+    "license": "AGPL-3",
+    "depends": [
+        "pos_margin",
+    ],
```
legalsylvain:
```suggestion
    "depends": ["point_of_sale"],
```

I think the dependency to the OCA module is not necessary. Don't you think ?

pos_margin_stored/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2025 - Today: GRAP (http://www.grap.coop)
+# @author: Quentin DUPONT
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "PoS Order Margin",
```
legalsylvain:
```suggestion
    "name": "PoS Order Margin Stored",
```

pos_margin_stored/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "summary": "Margin Stored on PoS Order",
```
legalsylvain:
```suggestion
    "summary": "Margin Stored on PoS Order and Pos Order Line",
```

legalsylvain (approved):
Minor remarks inline. Otherwise, LGTM.

Thanks !

pos_margin_stored/__manifest__.py
quentinDupont: you're god damn right !

PR #1443, opened 2025-10-08
17.0 imp pos order attachment scan

kamladodiyaserpentcs:


pos_order_attachment_scan/static/src/js/screens.esm.js
```diff
@@ -0,0 +1,86 @@
...
+                    _t("Failed to initiate scan: ") + errorText,
+                    { type: "danger", title: _t("Scan Error") }
+                );
+                return;
+            }
+
+            const scanResult = await scanResponse;
+
+            let attachmentId = false;
+            if (scanResult && scanResult.url) {
```
vehi-invitu, line 48: Hello @kamladodiyaserpentcs,
We tested it several times, and it turns out that when we scan the document, the scanned image is visible at "`${scanServerURL}/api/scan/${scannerIdForURL}`" but it isn't being added to the order's attachments. We also tried to retrieve it through the browser's debugger but couldn't find it (either as a jpg file or in base64 format). Could you please take a look at this and explain how this part works ? Thanks

kamladodiyaserpentcs: Hello @vehi-invitu 
Thank you for the update and for testing.

I checked how the scan process works. When the “Scan” button is clicked, the POS calls the Scan Server API at ${scanServerURL}/api/scanners to get the available scanner list. It then picks the first scanner and calls ${scanServerURL}/api/scan/{scannerId} to initiate the scan.

Once the scanning is complete, the server returns the scanned file details — either as a file URL or as base64 image data.
That data is then passed to the Odoo backend using attach_document_from_url() or attach_document() on the related POS order record, which automatically creates an attachment and links it to that order.

If the Scan Server URL is missing or the API doesn’t return a valid file, a user notification appears in the POS showing the specific error (for example, “Scan Server URL is not configured” or “No valid file URL or image data received”).

Thank You !!

cvinh: Hi @kamladodiyaserpentcs 

We have a Brother DS-840 scanner...

${scanServerURL}/api/scan/{scannerId} is perfectly launched... BUT
The response is the image itself, it's not an url. We are sure about that because if we launch ${scanServerURL}/api/scan/{scannerId}  in a browser, we have the image shown in the browser (image/jpeg).
The problem is that the response contains an 'url' aswell so it goes in the 'url clause' and the server tries to call this url (of course, it's not reachable because the server cannot reach the client)

So I think that the behaviour is depending on the scanner... WDYT ?

We tried to debug the js but we cannot find the base64 encoded in the response with the browser debugger... 

On the other side, when we call the URL directly from a browser with developer tools activated, the scanner starts and we can see the base64 in the network/response tab (the image is shown in the preview tab) but it does not help to know how we can send the base64 to the orm...

Do you have some hints to help ?

Thanks

PR #1444, opened 2025-10-09
[FIX] pos_access_right: enforce payment button restriction in tablet mode

sistecem:
Previously, users without the required access rights could still access the payment button when switching to tablet mode in the POS interface. This bypassed the intended security restriction.

This fix ensures that the payment button remains disabled for unauthorized users regardless of the POS display mode.

Issue: Payment button enabled in tablet mode despite access restriction

### DIFF
--- a/pos_access_right/static/src/css/pos.css
+++ b/pos_access_right/static/src/css/pos.css
@@ -5,6 +5,11 @@
     cursor: not-allowed;
 }
 
+.pos .switchpane .btn-switchpane.disabled-mode,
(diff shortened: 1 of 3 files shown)


LudwingRopz:
Hi, this PR has already passed automatic validation. Could someone please review it when they have a chance?
@pedrobaeza

pedrobaeza:
This may be checked by the maintainers of the module/PSC of this repository, but first of all, the PR is incorrect with a merge commit that shouldn't be there. Rebase over main branch.

PR #1448, opened 2025-10-24

gfcapalbo:
When visualizing paid orders, the container_products associated to a product  would show duplicated.

this is because after loading an existing pos.order from the backend, the code would still add another container_deposit_product, therefore duplicating them all , in visualization.

the problem was seen by users when accessing the "orders" icon, and visualizing paid orders, they would appear with duplicates. also the receipts, generated from that visualization were also incorrect.

This would only be a POS visualization problem, because the pos.order had been payed and closed and was not editable, but the visualization was broken, and also the receipts were printed wrong on every step of the way.

I initially tried to put a check on  init_from_Json, to stop the orders coming from json to have duplicates re-added, but thiswould not stop the duplicates from appearing on the currently active order, wich was also, at times loaded form json (e.g. page reload of the same session).

a simple way to stop the addition of the lines was to check the state, an open active order needs the container_deposit_product  , all the orders that have a state do not, they already have the container products added.

i prefer to check for` order.state == undefined`  to determine currently active order instead of    `  !['done', 'invoiced', 'paid'].includes(order.state) `  to cover future state extensions.

after few temporary solutions, i found no need to add flags in constructor of the order or  in init_from_Json. 

The option of sanitizing json source from the container_products and leaving the auto-add active for all orders is a bad idea, because it de-syncs odoo and pos.  if someone were to manually removes the container_deposit_product line from the backend , this will show the correct lines, without re-adding.

tested 25 minutes in various settings. undergoing more testing from others.

### DIFF
--- a/pos_container_deposit/static/src/js/models.js
+++ b/pos_container_deposit/static/src/js/models.js
@@ -40,7 +40,7 @@ odoo.define("pos_container_deposit.models", function (require) {
                  * When adding a product with container deposit, add its container deposit product
                  **/
                 super.add_orderline(...arguments);
-                // Get unpayed orders orders from db.pos
+                // Get unpayed orders from pos.orders wich comes from pos.db.orders
                 const all_unpaid = this.pos.get_order_list().map((x) => x.cid);
                 const is_unpaid = all_unpaid.includes(line.order.cid);
                 if (
--- a/pos_container_deposit/static/src/js/models.js
+++ b/pos_container_deposit/static/src/js/models.js
@@ -40,7 +40,14 @@ odoo.define("pos_container_deposit.models", function (require) {
                  * When adding a product with container deposit, add its container deposit product
                  **/
                 super.add_orderline(...arguments);
-                if (line.container_deposit_product && !line.container_deposit_line) {
+                // Get unpayed orders orders from db.pos
+                const all_unpaid = this.pos.get_order_list().map((x) => x.cid);
+                const is_unpaid = all_unpaid.includes(line.order.cid);
+                if (
+                    line.container_deposit_product &&
+                    !line.container_deposit_line &&
+                    is_unpaid
+                ) {
                     this.add_product(line.container_deposit_product, {
                         quantity: line.get_quantity(),
                     });
--- a/pos_container_deposit/__manifest__.py
+++ b/pos_container_deposit/__manifest__.py
@@ -4,7 +4,7 @@
     "category": "Point of Sale",
     "summary": "This module is used to manage container deposits for products"
     " in Point of Sale.",
-    "author": "Sunflower IT, Open2bizz, Odoo Community Association (OCA)",
+    "author": "Therp B.V., Sunflower IT, Open2bizz, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/pos",
     "license": "AGPL-3",
     "depends": ["point_of_sale"],
--- a/pos_container_deposit/readme/CONTRIBUTORS.rst
+++ b/pos_container_deposit/readme/CONTRIBUTORS.rst
@@ -1,3 +1,4 @@
 Tom Blauwendraat <tom@sunflowerweb.nl>
 Holger Brunn <mail@hunki-enterprises.com> (https://hunki-enterprises.com)
 Stefan Rijnhart <stefan@opener.am>
+Giovanni Francesco Capalbo <giovanni@therp.nl>


pos_container_deposit/static/src/js/models.js
```diff
@@ -40,7 +40,11 @@ odoo.define("pos_container_deposit.models", function (require) {
                  * When adding a product with container deposit, add its container deposit product
                  **/
                 super.add_orderline(...arguments);
-                if (line.container_deposit_product && !line.container_deposit_line) {
+                if (
+                    line.container_deposit_product &&
+                    !line.container_deposit_line &&
+                    this.state === undefined
+                ) {
```
thomaspaulb, line 50: It looks like the real problem is that when a previous order is re-opened and loaded using `init_from_JSON`, which loops over the JSON and calls [add_orderline](https://github.com/OCA/OCB/blob/b8b7d14d1f89e9e3548320e6ffa928efe2431325/addons/point_of_sale/static/src/js/models.js#L2648) for each line to build up the order again; ... then when it comes to the "normal" product, `container_deposit_line` attribute is not set on the line (logical, because that deposit line does not yet exist), and so it adds a new one, but it also loads the line from scratch.

I agree that the solution should be that whenever add_orderline is called from init_from_JSON, then it should not auto-add container products, but should rather assume that these are already added when the order was first created. However is checking for `this.state === undefined` the best way of assuring that? I don't know if this is 100% accurate?

Can we instead perhaps set a flag on the order `is_from_json = True` in init_from_Json that is below?

Or perhaps we correct after the fact - after completing init_from_json, we remove any extra container deposit products that were added by add_order_line?

gfcapalbo: My first solution was to override init_from_json.
My understanding was that : if a cart is loaded from json, it has already the products added. fantastic. clean.
If the cart comes from json, do not auto-add.  unfortunately this is not how pos works.

init_from_json is called when loading pos.orders, yes,  but also when loading pos.session, live data, wich is not recorded in the backend but stored in the session .. in the pos.db object:  https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2647

  
So there is a case where init_from_json is called on a cart that is not coming from the backend.

I have verified this this morning again to be sure.  i create a new session , add product. In the backend a pos.session is created . the session in the backend has no products added. it seems that all the info is stored in the pos.db, with no autoproducts added, but still loaded with init_from_json.

Here you can see how does the session load unpayed orders,  and you can see they are stored in pos.db, not in backend model, and do not have the auto-add.
https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L466

Only after a form of payment , is the pos.order created and are the orderlines re-created in the backend. in other words:  !['done', 'invoiced', 'paid'].includes(order.state)  ensures  that the pos.order was created in the backend and that odoo is actually fetching from backend lines.

the only way i could find   to distinguish a pos.session (live cart) and a pos.order (a pos.session that has been payed) is the fact they don't have a state.  There is a "finalized" variable but it does not switch to true after payement.   
Pos uses this in code too... https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2664     to determine if order is "locked" or not.

I am thinking of a better way to distinguish the two. but both are loaded from.

ALTERNATIVE SOLUTION:  
I could add a flag in https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L481     createReactiveOrder.    and then check for this flag... if the order is reactive..... do your addition. other wise not.

I think it is equivalent to the state check. but if you like more.... please confirm

thomaspaulb: Okay I took a look at the POS code and I indeed see what you say: `Order.init_from_json` is called from `Order()`, which in turn is called from `createReactiveOrder()` and that is called from multiple places: `_load_orders` and `import_orders`, for example. I don't know exactly what these cases are, but it is always an order loaded from a piece of JSON. I think the two cases are 1. a backend order that is reloaded, and 2. an order imported from another POS terminal where it had been created, and then the session has been taken over by another terminal.

However I think that both of these cases should not add the deposit product. The only time when the deposit product should be added is when `add_order_line` is being called not from the Order() constructor but from `add_product`, which is called whenever a barcode is scanned or a product clicked on the touch screen.

Now I still fail to see how `state` will reliably determine if it's an order that's loaded from JSON, or a live order that a product is being added to. Also you say that it helps to distinguish the session import case from the backend import case, but in my opinion, neither of them should trigger the deposit product - no Order import from JSON should. Even if it does work, it seems a cryptic and fragile way to rely on the state being null.

So I still would rather see either of:
- An `Order.init_from_json` override that sets a flag on the Order object that will prevents the extra add_orderline of the deposit product to be triggered
- Moving the trigger from `add_orderline` to `add_product`, which is more guaranteed to always come from a user action on the POS front end

Unless of course you disagree. But I have a feeling that code as it is now will still duplicate deposit products when being loaded from JSON under certain circumstances.

gfcapalbo: @thomaspaulb  I would like to send a long message concerning all i have learned with various aspects. this message is inspired instead to new style of communication.

Point 1:
State is a reliable way to distinguish payed/history orders (wich are recorded in pos.order)  from current open carts from  this session (and other sessions) ( wich are not recorded in backend pos.order).

we see it in this line of code:  https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2664

I cannot use "locked" because add orderline is done before. but this line of code is proof that that is the distinguishing factor between orders that are frozen, and written to backend, and orders that are still stored in pos.db.  But the criteria for a "locked" order is based on the state present in the json.

init_from_json is called from order constructor : https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2564

We could plug in   pos.db.get_unpaid_orders and set flag there (it returns the Jsons of unpayed carts) https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L473   But i feel this is equivalent. we are still filtering the behaviour based on what the "state" will be.  do you like that more?


Point 2:
cannot put the flag in add product, because the orderline is already created at that point. we want to stop the creation of the extra orderline.

thomaspaulb: @gfcapalbo Looked at this a bit more and what I see is that:

- Any order that gets loaded into the current session via `load_orders` is loaded correctly, because it gets loaded from a JSON in `pos.db`, which got saved there usually by `export_to_JSON` which is overridden in `pos_container_deposit` to include `container_deposit_line` to link the two lines together, so based on that, the extra `add_orderline` does not triggger.
- Any order that gets loaded into the current session via direct RPC, like is being done in `_fetchSyncedOrders`, does not have the `container_deposit_line` link set on each orderline JSON and so `add_orderline` is wrongly triggered for a second time there.

So your solution works because in the above two cases, `state` is indeed a discerning factor because `_fetchSyncedOrders` always returns orders that have a state, and the currently open order does not have a state. But it's more of a coincidence.

gfcapalbo:
**summary of technical findings TL;DR:**

Unfortunately, when loading history, the "deposit product" orderline , is both loaded from the backend and then re-added with the same logic, resulting in a double deposit product line. We need to distinguish orders loaded from backend from live orders loaded from pos.db.

We have found that the order state is a reliable criteria to stop re-adding the deposit product , and determine it is already coming from the backend. https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2664
Locked orders cannot be saved to DB (wich in pos language means saved to pos.db) https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2578
the order is saved as pos.order here:

https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L1032

We can see it is removed from the pos.db, and not synced anymore.
The automatically added deposit products are saved to the backend, from this moment on, it will be loaded with a state , and with the auto added deposit products present in the pos.order, and should not be re-adedded.

In conclusion: Orders loaded from backend will have "state" in there Json and be set as "locked" (unmodifiable). These orders are written as pos.order records in the backend at the moment of payment. All other orders are saved in pos.db, do not have state.

Implementing the connection between the product and it's added deposit in the backend too, would also be a solution and stop the addition , but if the user changes the backend in any way (quantity, removing or adding deposit products) this type of sync would force additions customer may not want. We do not want this module to force modifications the user may have overridden.

thomaspaulb (approved):
So yes I feel I can approve this because even though I think the "proper" solution would have been that we also save "container_deposit_line" on the sale.order.line id in the backend, to hard-link the product line with its deposit line, so that when reloading it from the backend the both are linked:

1. The "state === undefined" criterion also works because at least in 16.0 `point_of_sale`, the only function that really loads orders from the backend is `_fetchSyncedOrders`, and that one always loads finished orders anyway, so state !== undefined and it does solve the bug of the historic orders that are loaded, having the duplicate lines. 

2. Looking at state is actually better, because in fact you never want a deposit product to be (re)added if it is not already present in a paid order: the order is finalized and should not be changed. People could have manually removed the deposit product as a correction and then that manual correction should be respected.

So I can live with this

gfcapalbo:
@thomaspaulb   In certain circumstances we find our order.state is still undefined.  new solution is better.

To distinguish payed orders and not payed orders , all we need to do is fetch unpayed orders from posGlobalState object:
https://github.com/OCA/OCB/blob/16.0/addons/point_of_sale/static/src/js/models.js#L698-L701

If our line.order is not in the unpayed orders, we do not apply the logic.
Only payed orders are recorded in backend , with the container_deposit permanently added, and are removed from pos.db once payment happens and pos.order is created:
https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L1043

pos_container_deposit/static/src/js/models.js
```diff
@@ -40,7 +40,14 @@ odoo.define("pos_container_deposit.models", function (require) {
                  * When adding a product with container deposit, add its container deposit product
                  **/
                 super.add_orderline(...arguments);
-                if (line.container_deposit_product && !line.container_deposit_line) {
+                // Get unpayed orders orders from db.pos
```
thomaspaulb: If anything this would be `pos.db`, but `get_order_list` does not get from `pos.db` (=localStorage), but directly from `this.orders`, which as far as I can see is just a list of all orders that have been touched since this particular browser tab was opened. So the claim that this comment does is false.

pos_container_deposit/static/src/js/models.js
```diff
@@ -40,7 +40,14 @@ odoo.define("pos_container_deposit.models", function (require) {
...
+                const all_unpaid = this.pos.get_order_list().map((x) => x.cid);
+                const is_unpaid = all_unpaid.includes(line.order.cid);
```
thomaspaulb, line 45: This looks like it is a good criterium, I tested on Runboat and it seems like this is indeed a measure of which orders are open in the current session and not paid yet.

thomaspaulb (commented):
nonblocking but the comment is technically false

pos_container_deposit/static/src/js/models.js
gfcapalbo: @thomaspaulb 
pos.db.. i will correct my typo now.  but get_orders is connected to pos.db   

You can see that the orders attribute of the pos object is taken from pos.db:  https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L246

get_orders fetches pos.orders, pos.orders is taken on session load from   pos.db.orders

the comment is true.

gfcapalbo:
@legalsylvain @thomaspaulb  comment was fixed when MR was reviewed. ready for merge.

@ntsirintanis  @KKamaa  if you can , an extra review might increase merge chances.

thomaspaulb (approved):

pos_container_deposit/static/src/js/models.js
KKamaa: @thomaspaulb @gfcapalbo maybe we can use [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) like below, it avoids creating an intermediate array and looks cleaner:
```javascript
const is_unpaid = this.pos
    .get_order_list()
    .some((order) => order.cid === this.cid);
```

KKamaa:
> @legalsylvain @thomaspaulb comment was fixed when MR was reviewed. ready for merge.
> 
> @ntsirintanis @KKamaa if you can , an extra review might increase merge chances.

@thomaspaulb @gfcapalbo not sure if tours tests are needed but looks like the pipeline is all green maybe its not needed for such a small fix?

gfcapalbo:
> @thomaspaulb @gfcapalbo not sure if tours tests are needed but looks like the pipeline is all green maybe its not needed for such a small fix?

@KKamaa  it is very very small. i think adding tours for such a small feature is just overkill. 
Please approve if the rest looks good.

@legalsylvain @thomaspaulb @OCA/pos-maintainers    do i need 3 approvals for merge?
@NL66278  could you check it out as well?

KKamaa (approved):

mihien (approved):
Reviewed and tested, LGTM !

thomaspaulb:
@OCA/pos-maintainers Merge?

legalsylvain:
no review. 

/ocabot merge patch

thanks for your contribution !

PR #1450, opened 2025-10-30
[16.0][ADD] pos_hide_refund_button

mihien:
Hide the refund button in the TicketScreen.


legalsylvain:
Why prevent refunding ?

mihien:
Now that you mention it, this might not be a module suited for the OCA. It's a request by a client, whose worflow doesn't work with refund. 
I'm moving this to another repo.

PR #1451, opened 2025-11-03

gfcapalbo:

Unfortunately, when loading history, the deposit product orderline , is both loaded from the backend and then re-added with the same logic, resulting in a double deposit product line. We need to distinguish orders loaded from backend from live orders loaded from pos.db.

We have found that the order state is a reliable criteria to stop re-adding the  deposit product , and determine it is already coming from the backend.  https://github.com/odoo/odoo/blob/16.0/addons/point_of_sale/static/src/js/models.js#L2664

Keeping the connection between the product, and it's added deposit would also be a solution, and stop the addition , but if the user changes the backend in any way (quantity, removing or adding deposit products) this type of sync would force additions customer may not want. We do not want this module to force modifications the user may have overridden.

### DIFF
--- a/pos_container_deposit/static/src/js/models.js
+++ b/pos_container_deposit/static/src/js/models.js
@@ -40,7 +40,11 @@ odoo.define("pos_container_deposit.models", function (require) {
                  * When adding a product with container deposit, add its container deposit product
                  **/
                 super.add_orderline(...arguments);
-                if (line.container_deposit_product && !line.container_deposit_line) {
+                if (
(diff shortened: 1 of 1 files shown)


legalsylvain:
I know that order lines looks duplicated, but avoid to duplicate issues also ! 
Ref : https://github.com/OCA/pos/pull/1448

thomaspaulb:
Ah sorry, we felt the original PR got a bit convoluted with all our discussions happening, so we wanted to open a new clean one.

PR #1453, opened 2025-11-09
[18.0][MIG] pos_payment_change: Migration to 18.0

Abranes:
Migration to 18.0 from #1331


legalsylvain (approved):
Diff review. LGTM. Thanks !

pedrobaeza (approved):
Only OCB is failing due to a tour... let's try again

/ocabot merge nobump

pedrobaeza:
There's no way...

flotho:
looks like the same as https://github.com/OCA/pos/pull/1455#issuecomment-3521499431

HaraldPanten:
@pedrobaeza could you retry the merge command? Or do you think it will fail again?

THX.

pedrobaeza:
I think this will fail while #241122 is not merged.

HaraldPanten:
THX @pedrobaeza .

By the way... are you sure that this module is needed in >17.0 ? I see that Odoo has an option for this. Isn't this functionallity included in core? (see screensot).

Captura de pantalla 2026-02-04 a las 14 13 50

And it can be changed using backend options by entering the changes directly in the pos order, as well.

@Abranes @legalsylvain

HaraldPanten:
@Abranes @legalsylvain Could you check my comments?

THX!

legalsylvain:
> By the way... are you sure that this module is needed in >17.0 ? I see that Odoo has an option for this. Isn't this functionallity included in core? (see screensot).

Nice catch @HaraldPanten. Well, in fact this module should be refactored. 
-> indeed the "change and replace previous payment lines" is now native and could be removed. (payment_change_policy = update)
-> However, the "refund the order and create a new one with correct payment lines" doesn't look to exist natively. (Or did I missed something ? ) So this feature should be ported in V18. (payment_change_policy = refund)

legalsylvain:
@Abranes : could you rebase and take a look on red CI ? 
thanks, and sorry for the noice.

victoralmau:
Related to https://github.com/OCA/pos/pull/1453#issuecomment-3847414401 I agree that the module is not necessary **EXCEPT** for https://github.com/OCA/pos/blob/bd4dcc239a8ebb4b4296626b485bf15b830ecc57/pos_payment_change/views/view_pos_order.xml#L19

In v16, it is possible to change the payment method after an invoice has been issued, and that is something the core does not allow; therefore, perhaps this module (with a simpler approach) should be kept just for that specific case.

ejemplo

Ping @pedrobaeza TT62460

pedrobaeza:
Isn't possible to edit the `account.payment` records?

victoralmau:
> Isn't possible to edit the `account.payment` records?

That's not possible, but I also think there are several important points to clarify.

**v16**
A user with POS permissions but NO invoicing permissions could see the button to open the wizard (https://github.com/OCA/pos/blob/16.0/pos_payment_change/security/ir. model.access.csv#L2 the permissions listed are incorrect IMO; they should be POS permissions) and upon completion, the `pos.payment` record had been “modified” via https://github.com/OCA/pos/blob/16.0/pos_payment_change/models/pos_order.py#L16

v16-test1
v16-test-2


**v18**
A POS user does not have access to invoicing, so they cannot modify `account.payment` records. However, the user wants to modify the `pos.payment` record in some way so that the corresponding information on the invoice is updated.
In v18, it is NOT possible to modify that data
v18-example
and therefore, I believe it is also not possible to propagate the change to the invoice (reconcile the `account.payment` record, delete it, or modify it to define the appropriate account according to the new payment method), etc.

pedrobaeza:
OK, let's create then a new module `pos_order_invoiced_payment_change` for that case.

pedrobaeza:
Or are you saying you need invoicing permission for changing any payment, being invoiced or not?

legalsylvain:
> Related to https://github.com/OCA/pos/pull/1453#issuecomment-3847414401 I agree that the module is not necessary EXCEPT for


Not only see : https://github.com/OCA/pos/pull/1453#issuecomment-3947962602

In france, the core feature is not usable, as it is not allowed to edit a paid order. (see l10n_fr_cert_pos ...) As a result, only the current implemented function "refund and create a new one order" is the valid approch.

regards.

pedrobaeza:
@legalsylvain but you don't modify the invoice, but the payment. Why do you need to refund in that case?

legalsylvain:
> @legalsylvain but you don't modify the invoice, but the payment. Why do you need to refund in that case?

sorry I did'nt explained correctly. 

- "Edit payment" allow to change "pos.payment", right ? (in classic orders without invoicing & without customers).
- But the l10n_fr_pos_cert prevent to change payment_ids of pos.order. In fact, if there is a change (a payment has been replaced by two payments for exemple), the hash will change and the inalterability will be broken.
see : payment_ids here : https://github.com/odoo/odoo/blob/18.0/addons/l10n_fr_pos_cert/models/pos.py#L45

So that's the reason I think we should continue to implement the "refund and create a new order" feature.
don't you think ?

pedrobaeza:
Yes, indeed, but I would say that should be another different module.

legalsylvain:
Something like pos_payment_change_via_refund ?

Yes. Module can be renamed if it's more clear.

pedrobaeza:
Uhm, if everything goes around the payment change, we may keep the name and global module. No strong opinion on this, so what you prefer.

legalsylvain (approved):
diff review. LGTM.

victoralmau (changes requested):
It is not working properly

- In pos.config, we set Payment Change Policy=Update Payments
- In POS, we create an order linked to a customer with an invoice and pay with cash
- In the backend, in pos.order, we use the wizard to select bank, but an error appears

Image

Image

victoralmau:
Superseed by https://github.com/OCA/pos/pull/1573

PR #1459, opened 2025-11-25
[18.0][MIG] pos_transfer_account: Migration to 18.0

P-H-Phuc:
This module was initially part of the [Foodcoop set of addons](https://github.com/AwesomeFoodCoops/odoo-production/tree/12.0). We're now migrating it to 18.0 and taking this occasion to contribute it to OCA.

- This module was written to extend the functionality of odoo Point Of Sale about transfer account.
- This module is usefull only in a multi Point of Sale context of multi company context.
- With this module, it is now allowed to define for each PoS config a transfer account.
- If this account is set, it will be used when user realizes "Cash In/Out" operation.

image

- Remove `pos_transfer_account/models/cash_box_in.py` and  `pos_transfer_account/models/cash_box_out.py`: its model was removed from 14.0.


ivantodorovich (approved):
/ocabot rebase

PR #1465, opened 2025-12-08
[MIG] pos_lot_barcode: Migration to 19.0

mkoeck:


mkoeck:
I also still have an open PR here https://github.com/OCA/pos/pull/1430 for the same module. Since this is a migration PR I did not include the changes from https://github.com/OCA/pos/pull/1430 but the issue still persists. Would be nice to get some feedback on both.

mkoeck:
Fixed all the existing tests which now pass and added a test for the issue described in https://github.com/OCA/pos/pull/1430 which now fails. Please advise on how to proceed with this.

ivantodorovich (approved):
Thanks!

Can you clean up your migration commits? pehaps squash all or most of them?

/ocabot migration pos_lot_barcode

mkoeck:
@ivantodorovich squashed the commits into two separate parts: 1. the normal migration, 2. the fix of the issue described in https://github.com/OCA/pos/pull/1430

I hope its fine to include that one here as well. 

The CI pipeline seems to fail, even though the tests are passing. I am not sure what is going on here. Seems to be a false positive due to the Chrome shutdown warning?

PR #1469, opened 2025-12-30
[18.0][ADD] pos_receipt_hide_payment_info

nihelgabsi-acsone:
This module extends Odoo payment method configuration to allow hiding terminal payment information from customer receipts. It adds a simple option on payment methods that controls whether payment details (such as card or terminal transaction information) should be displayed on printed or digital receipts.


flotho:
Could you elaborate the difference with 
- pos_ticket_without_price
- pos-receipt-hide-price

Thanks

PR #1470, opened 2026-01-04
[18.0][ADD] pos_sale_picking_keep: Keep sale pickings from PoS

pedrobaeza:
This module inhibits the manipulation that the point of sale mades over the sales orders pickings, and the creation of new pickings under the PoS picking type.

Use Cases / Context
===================

In some scenarios, you may not want that the point of sale (PoS) handles the pickings of the products you are paying:

- Using the PoS as a pure payment terminal.
- When complex stock flows are not supplied by the PoS.

In that cases, it's better to keep the original sales pickings.

TODO: Tests

@Tecnativa TT60048


legalsylvain (approved):
Makes sense. 

thanks !

victoralmau (approved):
Code review OK

victoralmau:
Test error will be fixed with https://github.com/odoo/odoo/pull/241122

victoralmau:
The CI is now fixed, it could be merged.

rousseldenis:
@pedrobaeza Is it possible this empeaches the use of mixed pos orders (some normal lines selected in POS and other ones coming from exisiting sale orders) ?

As I tried mixed pos orders (with sales and additional products), normal products (not in sales) are not transferred at all...

Many thanks

pedrobaeza:
Hi, Denis, indeed, this disables whole picking generation without checking which are added manually and which comes from SO. We can add it to the known issues. We don't plan to handle that, as our case needs to disable everything.

PR #1472, opened 2026-01-12
[18.0] [MIG] pos_session_pay_invoice

Tisho99:
Standard migration 
- The pos order is now created in draft state because we can no longer create pos payments in posted pos orders. The 'invoiced' state is assigned after that
- The refund of the setUpClass is now generated from a new invoice because in v18 the original and refund moves were being automatically conciliated

Depends of https://github.com/OCA/account-payment/pull/820

T-9200


javierizaca (approved):
Functional review completed. The module works as expected, allowing payments to be made from the backend of an active session. LGTM.

pos_session_pay_invoice/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "POS Session Pay invoice",
-    "version": "16.0.1.0.3",
+    "version": "18.0.1.0.3",
```
ValentinVinagre: version

pos_session_pay_invoice/wizard/cash_pay_invoice.py
```diff
@@ -89,13 +89,14 @@ def action_pay_invoice(self):
...
         pos_order._apply_invoice_payments()
+        pos_order.write({"state": "invoiced"})
 
     def _prepare_pos_order_vals(self):
         return {
             "amount_total": self.amount,
             "partner_id": self.invoice_id.partner_id.id,
             "account_move": self.invoice_id.id,
-            "state": "invoiced",
+            "state": "draft",
```
ValentinVinagre, line 99: Is it validated at some point?

pos_session_pay_invoice/tests/test_pay_invoice.py
```diff
@@ -68,14 +68,22 @@ def setUpClass(cls, chart_template_ref=None):
             }
         )
         cls.invoice_in.action_post()
+        cls.invoice_out2 = cls.invoice_out.copy()
```
ValentinVinagre, line 71: explain why copy please

ValentinVinagre (changes requested):

pos_session_pay_invoice/wizard/cash_pay_invoice.py
Tisho99: It is validated (or set as 'invoiced') here: https://github.com/OCA/pos/pull/1472/files#diff-f228ebf7313229f41451c1c8f72b80d566b40b9075e7fefd8dfac7313711e159R92

I searched for highet level function in pos.order to assign the state, but found anything useful for this case

pos_session_pay_invoice/tests/test_pay_invoice.py
Tisho99: We need to generate a invoice with the same values as invoice_out, and the copy is the methot that does it with less code

pos_session_pay_invoice/wizard/cash_pay_invoice.py
ValentinVinagre: Have you tried the following function 'action_pos_order_invoice' which calls '_generate_pos_order_invoice' which generates an invoice and marks the 'pos order' as invoiced?

https://github.com/odoo/odoo/blob/bcc1397c7d694dbe61ecbd44d0320b9518df84cb/addons/point_of_sale/models/pos_order.py#L984C9-L984C33

https://github.com/odoo/odoo/blob/bcc1397c7d694dbe61ecbd44d0320b9518df84cb/addons/point_of_sale/models/pos_order.py#L1010

Tisho99: We don't have to generate a new invoice, the invoice is already generated. We are associating a new pos order with an already created invoice. So using that function does not make sense for me

ValentinVinagre (approved):
LGTM 👍🏻

cristina-hidalgo-tecnativa (approved):
I've tested this locally, both at code level and functionally. 

Regarding the CI/CD failures, they don't seem related to this module at all — for example, the `IrModuleModule._load_pos_data_fields()` error (tracked in odoo/odoo#241122) was already fixed in the OCB via commit OCA/OCB@692620ab6. 

Also, the module's own tests pass cleanly on my end, so LGTM.

Tisho99:
Rebased to see if the tests pass now

pedrobaeza:
Waiting https://github.com/OCA/oca-ci/pull/123 for the CI to greenify.

Tisho99:
Hello @pedrobaeza 

This pull request is ready, in case you want to review it or merge it.

Thanks!

PR #1478, opened 2026-01-26
[16.0][IMP] pos_price_to_weight: price incl./excl. in the barcode

mihien:
## Description

Add an option in the point of sale settings to indicate whether, when we scan a product barcode, the taxes are included or not.


pos_price_to_weight/models/res_company.py
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: 2026 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _name = "res.company"
+    _inherit = ["res.company"]
```
legalsylvain:
```suggestion
    _inherit = "res.company"
```
is enough

legalsylvain (commented):
Hi @mihien. Thanks for your contribution. could you update the CONFIGURE section. A I think this option is a little obscure, so, I could be great to explain the use case when it's not the list_price field.

thanks !

pos_price_to_weight/models/res_company.py
```diff
@@ -0,0 +1,22 @@
...
+
+    price_to_weight_price_field_id = fields.Many2one(
+        string="Price To Weight Field",
+        comodel_name="ir.model.fields",
+        domain=[("model", "=", "product.product"), ("ttype", "=", "float")],
```
legalsylvain: nitpicking : is it possible to filter monetary field only ?

mihien:
@legalsylvain Hi Sylvain! I've rewritten my changes, as the goal was to be able to chose whether the taxes are included in the product's barcode or not. But I missed the fact that the prices with and without taxes are not fields of the product model in 16.0 anymore.

pos_price_to_weight/static/src/js/ProductScreen.esm.js
```diff
@@ -0,0 +1,65 @@
+/** @odoo-module **/
+
+/*
+    Copyright (C) 2024 - Today: GRAP (http://www.grap.coop)
+    @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
legalsylvain, line 6: Please git mv to make easy the review of the diff !

pos_price_to_weight/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _name = "res.company"
+    _inherit = "res.company"
+
+    is_price_to_weight_with_VAT = fields.Boolean(
```
legalsylvain: Default =True

pos_price_to_weight/static/src/js/ProductScreen.esm.js
mihien: There are too many changes to the file when converting to es and even with git mv when we make too many changes + rename the file there is no way (or at least none that I've found) to have git detect it as changes to one file. So I've separated the renaming and the conversion into two different commits

polchampion:
@mihien v12 test KO on bees-test: price and weight is zero : 
Capture d’écran du 2026-02-16 14-33-00

how do I test this in v16 if I can't make it work on v12?

polchampion (approved):
successful functional test

polchampion (approved):
LGTM, successful functional test on the staging env

huguesdk (approved):
lgtm! thanks @mihien!

@legalsylvain: if you want to see only the relevant changes, select the last commit only.

legalsylvain:
/ocabot merge minor

Thanks !

PR #1485, opened 2026-02-09
[18.0][ADD] pos_automatic_cashdrawer

P-H-Phuc:
This module was initially part of the [Foodcoop set of addons](https://github.com/AwesomeFoodCoops/odoo-production/tree/12.0). We're now migrating it to 18.0 and taking this occasion to contribute it to OCA.


pos_automatic_cashdrawer/static/description/icon.png
ivantodorovich, line 1: Please remove to let the standard OCA module take over

ivantodorovich (changes requested):
Thanks!

IMO the module name is too generic for the OCA, though.
There are lots of cashdrawers out there, and AFAIR this module only works for one of them: cashlogy

If we just rename the module to `pos_automatic_cashdrawer_cashlogy` it would be OK, IMO.

In the future, if people is interested in implementing other brands, a common base module could be factorized

pedrobaeza:
FYI, check #1557 without the need of an IoTBox.

PR #1489, opened 2026-02-27
[16.0][ADD] pos_partner_restriction

ntsirintanis:
pos_partner_restriction: restrict partners loaded in POS

This module restricts which partners are loaded into the POS frontend.
Only partners with the boolean field “Allowed in POS” enabled are available for selection.

Useful to:

- Limit pay-on-account customers
- Avoid exposing unnecessary partners in POS
- Reduce frontend partner payload


BernatObrador:
Hello @ntsirintanis , thanks for the contribution.

The first load of partners is correct. However, the issue is that the “Search More” action does not preserve the same domain, so it ends up displaying partners that are not available in the POS.

Right now, when clicking on Search More, all partners are shown, bypassing the POS restriction.

In my case, I only have “Acme Corporation” available in the POS, and the initial load works as expected.
image

However, if I search for example by “a” and then click on Search More, the system returns all partners, ignoring the pos_pay_on_account restriction.
image

To fix this, you have two possible approaches:

1. Inherit PartnerListScreen (JS) and override the get partners() function to ensure that pos_pay_on_account is included in the query domain.

2. Alternatively, inherit point_of_sale.DB and override the search_partner function to add pos_pay_on_account to the search domain.

Either approach should ensure that the restriction is consistently applied, including when using Search More.

ntsirintanis:
> Hello @ntsirintanis , thanks for the contribution.
> 
> The first load of partners is correct. However, the issue is that the “Search More” action does not preserve the same domain, so it ends up displaying partners that are not available in the POS.
> 
> Right now, when clicking on Search More, all partners are shown, bypassing the POS restriction.
> 
> In my case, I only have “Acme Corporation” available in the POS, and the initial load works as expected. image
> 
> However, if I search for example by “a” and then click on Search More, the system returns all partners, ignoring the pos_pay_on_account restriction. image
> 
> To fix this, you have two possible approaches:
> 
>     1. Inherit PartnerListScreen (JS) and override the get partners() function to ensure that pos_pay_on_account is included in the query domain.
> 
>     2. Alternatively, inherit point_of_sale.DB and override the search_partner function to add pos_pay_on_account to the search domain.
> 
> 
> Either approach should ensure that the restriction is consistently applied, including when using Search More.



> 2\. Alternatively, inherit point_of_sale.DB and override the search_partner function to add pos_pay_on_account to the search domain.

Thanks for pointing this out

PR #1493, opened 2026-03-04
[16.0] pos_receipt_replace_user_by_trigram: Drop compute function

cyrilmanuel:
superseed of : https://github.com/OCA/pos/pull/1274

The compute function does not allow any customization on how the trigram had to be generated, and there a multiple ways to define how trigram must be generated.

Moreover, the compute did generate duplicates and there was no way to fix manually since the field was not stored. Also, making it stored but still computed did risk unwanted updates in case name was updated.

Having a basic stored field allows each company to customize how trigram can be generated using a server action. The logic previously provided is then moved to a cron that can be customized to suit trigram generation for each company.

@ivantodorovich @grindtildeath


wouitmil (approved):

PaoloYam (approved):

cvinh (commented):
Removing partner_firstname could lead to unexpected effects (lastname before firstname, 2 words firstnames...)
BTW you might depend on pos_partner_firstname which would be better
my 2 cents

pos_receipt_replace_user_by_trigram/models/res_partner.py
```diff
@@ -1,17 +1,10 @@
...
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
-    pos_trigram = fields.Char(compute="_compute_pos_trigram")
-
-    @api.depends("firstname", "lastname")
-    def _compute_pos_trigram(self):
-        for partner in self:
-            partner.pos_trigram = utils.get_trigram(partner.firstname, partner.lastname)
+    pos_trigram = fields.Char(tracking=True)
```
cvinh, line 10: maybe you might add a constraint to make trigrams unique as it was said https://github.com/OCA/pos/pull/1274#issue-2730842639

PR #1494, opened 2026-03-09
[ADD] pos_product_fixed_price

mihien:
Products can be set as having a fixed price in the PoS, meaning users can't change its price from the PoS interface.


pos_product_fixed_price/models/product.py
```diff
@@ -0,0 +1,11 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    is_pos_price_fix = fields.Boolean()
```
huguesdk, line 11:
```suggestion
    is_pos_price_fixed = fields.Boolean()
```

pos_product_fixed_price/static/src/js/Screens/NumpadWidget.esm.js
```diff
@@ -0,0 +1,33 @@
...
+
+const LPCRPosFixedPriceNumpadWidget = (OriginalNumpadWidget) =>
+    class extends OriginalNumpadWidget {
+        changeMode(mode) {
+            const orderLine = this.env.pos.get_order().get_selected_orderline();
+            if (mode === "price" && orderLine) {
+                const product = orderLine.product;
+                if (product.is_pos_price_fix) {
+                    this.showPopup("ErrorPopup", {
+                        title: this.env._t("Restricted access to product price."),
```
huguesdk, line 17:
```suggestion
                        title: this.env._t("Restricted access to product price"),
```

pos_product_fixed_price/static/src/js/Screens/NumpadWidget.esm.js
```diff
@@ -0,0 +1,33 @@
+/** @odoo-module **/
+// SPDX-FileCopyrightText: 2026 Coop IT Easy SC
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import {Component} from "point_of_sale.Registries";
+import NumpadWidget from "point_of_sale.NumpadWidget";
+
+const LPCRPosFixedPriceNumpadWidget = (OriginalNumpadWidget) =>
```
huguesdk, line 9:
```suggestion
const PosFixedPriceNumpadWidget = (OriginalNumpadWidget) =>
```

pos_product_fixed_price/static/src/js/Screens/ProductScreen/ProductScreen.esm.js
```diff
@@ -0,0 +1,21 @@
...
+import ProductScreen from "point_of_sale.ProductScreen";
+
+const PosProductFixedPriceProductScreen = (OriginalProductScreen) =>
+    class extends OriginalProductScreen {
+        async _addProduct(product, options) {
+            await super._addProduct(product, options);
+            if (product.is_pos_price_fix) {
+                this.env.pos.numpadMode = "quantity";
+            }
+        }
```
huguesdk, line 16: it’s good to have thought of this case, but it is the default behavior. `Order.add_product()` calls `Order.select_order_line()`, which sets `numpadMode` to `"quantity"`.

edit: i think that it is because of a custom module that sets the mode to `"price"` by default. but then this case should be handled in that module instead.

pos_product_fixed_price/static/src/js/Screens/NumpadWidget.esm.js
```diff
@@ -0,0 +1,33 @@
...
+        changeMode(mode) {
+            const orderLine = this.env.pos.get_order().get_selected_orderline();
+            if (mode === "price" && orderLine) {
+                const product = orderLine.product;
+                if (product.is_pos_price_fix) {
+                    this.showPopup("ErrorPopup", {
+                        title: this.env._t("Restricted access to product price."),
+                        body: this.env
+                            ._t(
+                                "The following product price can't be changed from the PoS: %(name)s"
```
huguesdk, line 20: i think this is slightly more clear:
```suggestion
                                "It is not allowed to change the price of this product: %(name)s"
```

pos_product_fixed_price/static/src/js/models.esm.js
```diff
@@ -0,0 +1,45 @@
...
+const PosProductFixedPriceOrderline = (OriginalOrderline) =>
+    class extends OriginalOrderline {
+        set_unit_price(price) {
+            // Price is not assigned on line creation and empty on line deletion
+            if (this.price && price !== "") {
+                if (this.product.is_pos_price_fix && this.price !== price) {
+                    Gui.showPopup("ErrorPopup", {
+                        title: _t("Restricted access to product price."),
+                        body: _t(
+                            "The following product price can't be changed from the PoS: %(name)s"
```
huguesdk, line 21: `env._t()` should be used when available.
```suggestion
                        title: this.pos.env._t("Restricted access to product price"),
                        body: this.pos.env._t(
                            "It is not allowed to change the price of this product: %(name)s"
```

pos_product_fixed_price/static/src/js/models.esm.js
```diff
@@ -0,0 +1,45 @@
...
+import {Gui} from "point_of_sale.Gui";
+import {Model} from "point_of_sale.Registries";
+import {_t} from "web.core";
+
+const PosProductFixedPriceOrderline = (OriginalOrderline) =>
+    class extends OriginalOrderline {
+        set_unit_price(price) {
+            // Price is not assigned on line creation and empty on line deletion
+            if (this.price && price !== "") {
+                if (this.product.is_pos_price_fix && this.price !== price) {
```
huguesdk, line 17: in which case is this true? it doesn’t seem possible to happen, as the price can only be modified when the numpad is in price mode, which cannot happen on this product. is this whole method actually useful?

pos_product_fixed_price/static/src/js/models.esm.js
```diff
@@ -0,0 +1,45 @@
...
+import {Order, Orderline} from "point_of_sale.models";
+import {Gui} from "point_of_sale.Gui";
+import {Model} from "point_of_sale.Registries";
+import {_t} from "web.core";
+
+const PosProductFixedPriceOrderline = (OriginalOrderline) =>
+    class extends OriginalOrderline {
+        set_unit_price(price) {
+            // Price is not assigned on line creation and empty on line deletion
+            if (this.price && price !== "") {
```
huguesdk, line 16: price is equal to `""` only when the numpad is in price mode and the backspace key is used. in quantity mode it is never set to this value.
```suggestion
            if (this.price) {
```

pos_product_fixed_price/static/src/js/models.esm.js
```diff
@@ -0,0 +1,45 @@
...
+
+const PosProductFixedPriceOrder = (OriginalOrder) =>
+    class extends OriginalOrder {
+        select_orderline(line) {
+            super.select_orderline(line);
+            // Ensures the mode is quantity, so that when deleting the orderline
+            // the quantity is being set to 0 and not the price
+            if (line && line.product.is_pos_price_fix) {
+                this.pos.numpadMode = "quantity";
+            }
```
huguesdk, line 41: again, this is the default behavior.

huguesdk (changes requested):
useful module, thanks @mihien. the code can be simplified, as explained below.

PR #1496, opened 2026-03-11
[16.0][FIX] pos_tare: support other decimal separators

huguesdk:
ensure that the locale-defined decimal separator can be used to input tare values, and not only the dot (on the scale screen and on the product screen).


legalsylvain (approved):
thanks ! no tests.

ivantodorovich (approved):
Makes sense

/ocabot merge minor

PR #1497, opened 2026-03-12
[18.0][ADD] pos_report_product_tag module : This module adds product tags in pos order report

cvinh:
This module adds a product tags in Pos Order Report to allow grouping
and filtering

Known issues / Roadmap
======================

Since product tags are a Many2many field, we had to choose to take the
first tag in the products. This avoids any double-counting of sales
figures that would occur with a direct Many2many field. This can be
improved if we want real figures for every tag in the products


cvinh:
Depends on https://github.com/odoo/odoo/pull/253001

cvinh:
Can you reopen please

PR #1511, opened 2026-03-26
[16.0][mig] pos_hide_empty_category

legalsylvain:
Port pos_hide_empty_category from V12.

This module extends Odoo Point Of Sale features, hidding empty PoS categories.

this module can be usefull to avoid to display categories, if they contains only non saleable products
that can occurs frequently if your shop has seasonal products.

image

image

CC : original reviewers : @ivantodorovich, @wytzejan, @chienandalu


ivantodorovich (approved):

legalsylvain:
nice ! it works today !

PR #1513, opened 2026-03-27
[FIX] pos_order_to_sale_order_delivery: avoid to block settings, in demo instance.

legalsylvain:
Rational, when lauching runboat with that module installed, it is not possible to configure settings. (write on res.config.settings model) because the field pos_iface_sale_order_delivery_carrier_ids is required when pos_iface_sale_order_allow_delivery is True. This trivial patch fixes this problem


ivantodorovich (approved):
Makes sense

PR #1519, opened 2026-04-03
Remove Akretion from authors

bealdav:
Removed Akretion and its contributor from the list.

I do no more use this module and I'm not an expert on POS.

Thanks

cc @legalsylvain could you merge please ?


ivantodorovich (approved):
fast-tracking per author request

/ocabot merge patch

PR #1249, opened 2024-10-01
[16.0][MIG] pos_customer_required_fields: Migration to 16.0

LaetitiaElabore:
@petrus-v 
migration of pos_customer_required_fields from 12.0 to 16.0 (with commits history)

PR #1520, opened 2026-04-03
[16.0][MIG] pos_customer_required_fields: migration to 16.0

huguesdk:
follow-up of #1249.


pos_customer_required_fields/static/src/js/Screens/PartnerListScreen/PartnerDetailsEdit.esm.js
```diff
@@ -0,0 +1,42 @@
...
+                        "section.partner-details"
+                    );
+                    required_fields.forEach((field_name) => {
+                        const inputField = baseElement.querySelector(
+                            `input[name="${field_name}"]`
+                        );
+                        if (inputField) {
+                            inputField.setAttribute("required", true);
+                        }
+                    });
```
huguesdk, line 34: it seems strange to me to do something like this while using owl. does someone knows a better way?

legalsylvain:
Hi. Thanks for porting this module.

I don't understand how it can work in reality, as there is a single fields with the list of fields required.

In my opinion, we should have 2 fields :

    res_partner_individual_required_fields_ids
    res_partner_company_required_fields_ids

IE : In the screenshot, there is the "firstname". It doesn't make sense for company.
I guess that you could also force user to enter a VAT number for company customer (in a e-invoicing context), but it will prevent creating individual.

Don't you think ?

Thanks !

huguesdk:
that’s an interesting point, indeed! one of our customers uses this with only the `email` field, so it works with both. @petrus-v what do you think about @legalsylvain's remark? what was your use-case when designing this module?

petrus-v:
The suggestion make sense legit to me, I guess we didn't needs it at that time.

@gregory-moka, @Rom10811 do you still use this module, do you have any opinion ?

PR #1523, opened 2026-04-09
[18.0][IMP] pos_order_to_sale_order: Added option to print sale pdf

christian-ramos-tecnativa:
Right now, once the sales order is created, the PoS order disappears, so there’s nothing you can hand to the customer as a receipt. This PR adds an option  to print the Sales Order PDF directly from the PoS.
@Tecnativa TT61831


ivantodorovich (approved):

legalsylvain (approved):
Code review. No test. Thanks !

PR #1528, opened 2026-04-18
[18.0][FIX] pos_product_display_default_code: skip opening control in test tour

cristina-hidalgo-tecnativa:
The tour clicks "Open Register" to confirm the opening cash, but in CI the `set_opening_control` RPC takes ~10s to respond. While waiting, the popup remains open and blocks the next tour step (product search), causing a timeout.

```
FAILED: [4/6] Tour SearchProductByDefaultCode → Step Search for a product using the search bar (trigger: .pos-rightheader .form-control > input).
Element has been found.
BUT: It is not allowed to do action on an element that's below a modal.
TIMEOUT step failed to complete within 10000 ms.
```

To fix this, `set_opening_control(0, "")` is called in test_frontend.py before starting the tour so the session is already opened and the popup never appears, removing the need for the `Dialog.confirm("Open Register")` step in the tour.

@Tecnativa
@pedrobaeza @christian-ramos-tecnativa


pedrobaeza (approved):
Let's fast track it for solving the integration tests.

/ocabot merge patch

PR #1536, opened 2026-04-28
[18.0][FIX] pos_order_to_sale_order: fix delivery_status not updated on invoiced order

cristina-hidalgo-tecnativa:
In Odoo 18, `is_storable` defaults to `False`, so no pickings are created when confirming the sale order. Additionally, the stored field `delivery_status` is not automatically recomputed after `button_validate()` in this RPC context, causing the following failure:

```
  File "/opt/odoo/auto/addons/pos_order_to_sale_order/tests/test_module.py", line 47, in test_pos_order_to_sale_order
    self.assertEqual(
AssertionError: False != 'full' : Delivery status must be equal to 'full'
```

To fix this, `setUpClass` is overridden to set `is_storable = True` on the test products, and `create_order_from_pos` is updated to explicitly write `delivery_status = 'full'` after all pickings are validated.

@Tecnativa
@pedrobaeza  @christian-ramos-tecnativa

### DIFF
--- a/pos_order_to_sale_order/models/sale_order.py
+++ b/pos_order_to_sale_order/models/sale_order.py
@@ -54,6 +54,9 @@ class SaleOrder(models.Model):
             for move in sale_order.mapped("picking_ids.move_ids_without_package"):
                 move.quantity = move.product_uom_qty
             sale_order.mapped("picking_ids").button_validate()
+            pickings = sale_order.picking_ids
+            if pickings and all(p.state in ("done", "cancel") for p in pickings):
+                sale_order.write({"delivery_status": "full"})
 
         if action in ["invoiced"]:
             # Create and confirm invoices
--- a/pos_order_to_sale_order/tests/test_module.py
+++ b/pos_order_to_sale_order/tests/test_module.py
@@ -9,6 +9,12 @@ from odoo.addons.point_of_sale.tests.test_frontend import TestPointOfSaleHttpCom
 
 @tagged("post_install", "-at_install")
 class TestUi(TestPointOfSaleHttpCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.whiteboard_pen.is_storable = True
+        cls.wall_shelf.is_storable = True
+
     def test_pos_order_to_sale_order(self):
         self.main_pos_config.with_user(self.pos_user).open_ui()
 


pos_order_to_sale_order/models/sale_order.py
```diff
@@ -54,6 +54,9 @@ def create_order_from_pos(self, order_data, action):
             for move in sale_order.mapped("picking_ids.move_ids_without_package"):
                 move.quantity = move.product_uom_qty
             sale_order.mapped("picking_ids").button_validate()
+            pickings = sale_order.picking_ids
```
pedrobaeza, line 57: I don't think this code is adding value, as the sales order has just been created, so pickings are not going ever to be in done or cancel state.

pos_order_to_sale_order/tests/test_module.py
```diff
@@ -9,9 +9,14 @@
 
 @tagged("post_install", "-at_install")
 class TestUi(TestPointOfSaleHttpCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.whiteboard_pen.product_tmpl_id.write({"is_storable": True})
+        cls.wall_shelf.product_tmpl_id.write({"is_storable": True})
```
pedrobaeza: It's clearer to use dot notation in this case.

```suggestion
        cls.whiteboard_pen.is_storable = True
        cls.wall_shelf.is_storable = True
```

pos_order_to_sale_order/models/sale_order.py
cristina-hidalgo-tecnativa: Sorry, but I don't get your point. `button_validate()` is called two lines above, so pickings are in done state at that point.

The block is needed precisely because `delivery_status` is a stored computed field that does not recompute automatically after `button_validate()` in this RPC context — a known Odoo 18 behaviour, [also acknowledged in OCA's own sale_delivery_state](https://github.com/OCA/sale-workflow/blob/18.0/sale_delivery_state/models/sale_order.py#L104).

pos_order_to_sale_order/tests/test_module.py
cristina-hidalgo-tecnativa: Modified now.

pos_order_to_sale_order/models/sale_order.py
pedrobaeza: Ah, OK, true, then what I see weird is to auto-validate the picking, as you may want to just transfer the PoS order to a sales order, not to "complete" the cycle.

@legalsylvain @christian-ramos-tecnativa does this makes sense to be something parametric?

pedrobaeza (commented):
Anyway, as it is, the patch is correct, so let's fast-track it to fix the problem in integration tests.

/ocabot merge patch

pedrobaeza:
Uhm, maybe I have been very quick in changing this, as I'm seeing that this affects not only test code, but the module itself. Let me know if this should be reverted.

PR #1545, opened 2026-05-05
[18.0][IMP] pos_sale_picking_keep: Add a strategy to keep sale pickings and/or POS pickings too

rousseldenis:
@victoralmau @pedrobaeza


pedrobaeza (commented):
If you mix directly items from PoS from other loaded from sales, I think there's no discussion: both should be preserved, the ones from PoS in the PoS picking, and the sales one in the sales pickings, so I think there shouldn't be an option, just to proceed that way.

We were previously discarding everything from PoS, because we don't have that mixed option (for now), but if having it in the future, I don't conceive another option except that.

rousseldenis:
> If you mix directly items from PoS from other loaded from sales, I think there's no discussion: both should be preserved, the ones from PoS in the PoS picking, and the sales one in the sales pickings, so I think there shouldn't be an option, just to proceed that way.
> 
> We were previously discarding everything from PoS, because we don't have that mixed option (for now), but if having it in the future, I don't conceive another option except that.

I wanted to keep your behaviour as I didn't know if you have particular processes on your side.

So, I keep the option activation or not through a Boolean in place of Selection.

pos_sale_picking_keep/models/sale_order_line.py
```diff
@@ -1,19 +1,44 @@
...
-    @api.model
-    def _convert_qty(self, sale_line, qty, direction):
-        if self.env.context.get("from_qty_delivered"):
-            return 0
-        return super()._convert_qty(sale_line=sale_line, qty=qty, direction=direction)
+        # Mimic what is done at pos_sale level but lowering the quantity
+        def update_qty_delivered_from_pickings(sale_line, pos_lines):
+            if all(
+                picking.state == "done" for picking in pos_lines.order_id.picking_ids
+            ):
```
petrus-v, line 17: not sure if we should prevent empty list (`all([]) == True`) here ?

```suggestion
            if pos_lines.order_id.picking_ids and all(
                picking.state == "done" for picking in pos_lines.order_id.picking_ids
            ):
```

petrus-v (approved):

pedrobaeza:
@rousseldenis you can go directly with no possible choice :)

rousseldenis:
> @rousseldenis you can go directly with no possible choice :)

I'd prefer not as using stuffs for tests in code is not my cup of tea ;-)

pedrobaeza:
I don't get you. What do you mean by `using stuffs for tests`?

pos_sale_picking_keep/models/sale_order_line.py
rousseldenis: @petrus-v Oh yes, wrong copy paste :sweat:

rousseldenis:
> I don't get you. What do you mean by `using stuffs for tests`?

You want to say, install this module and the behaviour changes. Correct ?

pedrobaeza:
Well, yes, so I get your point. Let's keep the activation by pos config.

pos_sale_picking_keep/models/pos_config.py
```diff
@@ -0,0 +1,16 @@
...
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    picking_keep_strategy = fields.Selection(
+        selection=[
+            ("keep_sale_pickings", "Keep Sale Pickings Only"),
+            ("keep_sale_pos_pickings", "Keep Sale and POS Pickings"),
+        ],
+        help="When loading sale orders in POS, Odoo cancels the sale pickings."
```
pedrobaeza, line 10: Put a default value

rousseldenis: You'd prefer a lambda to override if necessary ?

pedrobaeza: No need, as I can change that default in any custom redefining the field with that attribute, and in general, I will prefer the new behavior.

Anyway, now I'm thinking that none of these options is like the standard without the module, as Odoo by default nullifies the sales order picking, isn't it?

rousseldenis: I would just replace this by an option:

- [] Keep pickings for POS and Sale Orders.

pedrobaeza: OK

pos_sale_picking_keep/models/sale_order_line.py
rousseldenis: @petrus-v In fact, no. We should wait : https://github.com/odoo/odoo/pull/253333

pos_sale_picking_keep/models/pos_order_line.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def _launch_stock_rule_from_pos_order_lines(self):
+        """
+        Launch stock rules for pos order lines that are not linked to a sale order
+        line and when the strategy is to keep both pos and sale order pickings
+        """
+        lines_to_launch = self.filtered(
+            lambda line: not line.order_id.pos_config.keep_picking
+            or (
+                line.order_id.pos_config.picking_keep_strategy
```
victoralmau, line 17: The `picking_keep_strategy` field does not exist; you should use `keep_picking` instead.

pos_sale_picking_keep/models/sale_order_line.py
```diff
@@ -1,19 +1,42 @@
...
 
-    # TODO: Delete if merged https://github.com/odoo/odoo/pull/253333
-    @api.model
-    def _convert_qty(self, sale_line, qty, direction):
-        if self.env.context.get("from_qty_delivered"):
-            return 0
-        return super()._convert_qty(sale_line=sale_line, qty=qty, direction=direction)
+        # Mimic what is done at pos_sale level but lowering the quantity
+        # TODO: Delete if merged https://github.com/odoo/odoo/pull/253333
+        def update_qty_delivered_from_pickings(sale_line, pos_lines):
```
victoralmau, line 14: Personally, I don't like all this extra logic; the way it works now — https://github.com/OCA/pos/blob/18.0/pos_sale_picking_keep/models/sale_order_line.py#L9-L19 — is enough, right? In that case, I think it’s best to leave it as is and try to merge it with https://github.com/odoo/odoo/pull/253333

pos_sale_picking_keep/models/pos_order.py
```diff
@@ -27,20 +26,16 @@ def sync_from_ui(self, orders):
...
+        # Save picking state values
+        for picking in pickings:
+            picking_values[picking.id] = picking.state
+        # Don't mark the concerned pickings state as dirty to avoid
+        # unwanted recomputations
+        with self.env.protecting([state_field], pickings):
+            pickings.state = "draft"
+            res = super().sync_from_ui(orders)
+            for picking in pickings:
+                picking.state = picking_values[picking.id]
```
victoralmau, line 40: Why are these changes necessary? IMO, the current behavior at https://github.com/OCA/pos/blob/18.0/pos_sale_picking_keep/models/pos_order.py#L30-L33 should be enough, right?

victoralmau (changes requested):

carlos-lopez-tecnativa:
Superseded by https://github.com/OCA/pos/pull/1583

PR #1547, opened 2026-05-06
[CI] bump GitHub actions to Node 24 compatible versions

MiguelMachadoM:
This PR updates GitHub Actions versions to avoid Node.js 20 deprecation warnings and align workflows with the Node 24 runtime migration.
### Changes
- `actions/checkout` bumped from `v4` to `v5` in:
  - `.github/workflows/test.yml`
  - `.github/workflows/pre-commit.yml`
- `actions/stale` bumped from `v9` to `v10` in:
  - `.github/workflows/stale.yml`
- `codecov/codecov-action` bumped from `v4` to `v6` in:
  - `.github/workflows/test.yml`
No functional addon code changes are included.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,7 +13,7 @@ jobs:
   pre-commit:
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please don't do this. We work this in a central way through https://github.com/OCA/oca-addons-repo-template. There, the templates for all the repositories are agreed, and then spread through each of them.

If these updates are valid with no drawbacks (and should be seen if for all Odoo versions), then they should be included there.

PR #1551, opened 2026-05-11
[IMP] pos_place: avoid to display the button 'Select Place' in the Po…S front UI

legalsylvain:
if there is no places set in the back-office


quentinDupont (approved):
Code review ✔️

BhaveshHeliconia (approved):
LGTM!

PR #1562, opened 2026-05-29
[16.0][FIX] pos_tare: fix weight formatting on receipt

huguesdk:
*   ensure that the gross and tare weights are formatted correctly on the receipt, according to the uom rounding, the decimal precision and the decimal separator (in the same way as on the order screen).
*   rename `format_tare()` to `format_quantity()`.
*   separate the quantity and the unit name using a non-breaking space to avoid a line wrapping between them.
*   add `has_tare` property to `Orderline`.
*   update and improve fr translation.


legalsylvain (approved):
LGTM. Thanks.

PR #1510, opened 2026-03-26
[REF] pos_payment_change: move pos config in res.config.settings

legalsylvain:
Trivial PR.
use "new" res.config.settings model.

PR #1569, opened 2026-06-17
[17.0][MIG] pos_payment_change: Migration to 17.0

victoralmau:
Superseed https://github.com/OCA/pos/pull/1331

Extra changes:
- [x] Fix pre-commit changes
- [x] Squash administrative commits
- [x] Purge translations
- [x] Squash commits in the migration commit
- [x] Improve tests
- [x] Improve wizard form visualization

Please @pedrobaeza and @eduezerouali-tecnativa can you review it?

@Tecnativa


pedrobaeza (commented):
Shouldn't the readme be adapted and maybe change the module name (or directly to not be a migration, but an addition)?

victoralmau:
> Shouldn't the readme be adapted and maybe change the module name (or directly to not be a migration, but an addition)?

I don't quite understand what you mean by changing the README or that this isn't a migration; in v17, this module is required, and the functionality is as described.

pedrobaeza:
Wasn't the payment change already covered by core and this is only for the case where the invoice was issued?

victoralmau:
This module is used to change the payment method when the order has been paid or invoiced.

pedrobaeza:
So the option in core is in 18, not in 17?

victoralmau:
> So the option in core is in 18, not in 17?

Not exactly—if you don't mind, let's leave the v18-related issues in the v18 migration PR (once it's merged).

pos_payment_change/wizards/pos_payment_change_wizard_new_line.py
```diff
@@ -14,14 +14,18 @@ class PosPaymentChangeWizardLine(models.TransientModel):
...
     )
-
     new_payment_method_id = fields.Many2one(
         comodel_name="pos.payment.method",
         string="Payment Method",
         required=True,
-        domain=lambda s: s._domain_new_payment_method_id(),
+        domain="[('id', 'in', available_payment_method_ids)]",
+    )
+    available_payment_method_ids = fields.Many2many(
```
pedrobaeza, line 24: Why not being a related field?

pedrobaeza (approved):

legalsylvain:
@victoralmau : 

thanks for porting this module.

could you review pending V16 PR https://github.com/OCA/pos/pull/1510 that improve configuration and ideally cherry-pick the commit in this migration PR ? 

thanks !

pos_payment_change/models/pos_order.py
```diff
@@ -23,17 +23,14 @@ def change_payment(self, payment_lines):
         """
         self.ensure_one()
         orders = self
-
```
legalsylvain: Could you avoid to remove empty lines. It generate a lot of diff that are quite useless. 

Thanks !

pedrobaeza: We prefer that style when possible. It's true that it can be done in a `[REF]` commit instead.

legalsylvain: Indeed. It's just preference, not oca rules. Please avoid introducing changes based on preferences.

If all people are doing the same in each migration (removing empty lines, adding new empty lines, ) it will générate a lot of useless diff and conflict during cherry picks. 

Thanks a lot !

victoralmau: I'll make the change you mentioned soon, but personally, I'll avoid doing it that way next time. 
Right now, I've done something (removing empty lines) and now I have to undo it, redo it in a separate commit, and then apply the migration changes.

legalsylvain: > I'll make the change you mentioned soon, but personally, I'll avoid doing it that way next time.

I'm sorry if this creates extra work.

For the next time, the more simple is to avoid to do changes that doesn't bring nothing and are pure preferences. In the list you mention in the first messages of your PR, I think you can remove "Remove empty lines".

> and now I have to undo it, redo it in a separate commit,

I think that amending the last commit that introduced the empty lines removal should do the job (a06ce1ffe08ec3928cd14bd24961abb7eed29f86)

That being said, thanks a lot for your work in this migration.

regards.

victoralmau: Split into a previous commit.

legalsylvain: weird. there are still some removals...

victoralmau: Sorry, but in the migration commit https://github.com/OCA/pos/pull/1569/changes/ac6bd576faa99afc4c64b8f514cc5bab31c4c888, I don't see any empty lines that have been removed.

legalsylvain: I think you checked the wrong commit.

when I take a look on the three commits you introduced, we see that there are some removals.

https://github.com/OCA/pos/pull/1569/changes/2589ef127c2a1733ca8f24484e506ec0287a95d3..ac6bd576faa99afc4c64b8f514cc5bab31c4c888#diff-805b555138cd4a5f5dd53d991f4ae8d2dc41ee9dd4a7398d14245e9d77329939L34

legalsylvain (changes requested):

pos_payment_change/models/pos_order.py
victoralmau: According to https://github.com/OCA/pos/pull/1569#discussion_r3437022202, there is ONE commit prior to the migration that is ONLY intended to remove empty lines.

ejemplo

legalsylvain:
> According to https://github.com/OCA/pos/pull/1569#discussion_r3437022202, there is ONE commit prior to the migration that is ONLY intended to remove empty lines.

Yes. could you remove it ? I can do it if you have no time.

thanks !

legalsylvain (approved):
Thanks for porting this module.

Diff review. No test.

/ocabot merge nobump

pos_payment_change/models/pos_order.py
pedrobaeza: @legalsylvain you can't impose to remove the refactoring of the empty lines. It has been done in a separate commit, for not contaminating the migration commit, but we are now maintaining this module for the next versions, so you should allow that we adapt it to our style for not having so much vertical scroll with a lot of lines meaningless empty lines.

PR #1570, opened 2026-06-18
[ADD] pos_ship_later_default: Set ship later by default

alan196:
Creates the pos_ship_later_default module to configure shipping delay by default.


alan196:
ℹ️ Tests pass (`tour SUCCEEDED`, `0 failed, 0 error(s)`). CI is red only because of the Chrome teardown WARNING (`Killing chrome descendants-or-self ... chrome (zombie)`) that `checklog-odoo` flags, which affects every PR with a tour in this repo. The infra fix is in #1577; once merged, this PR will go green when re-run on the updated base.

PR #1571, opened 2026-06-19
[ADD] pos_pywebdriver: support pywebdriver printing, even in Enterprise installs

dreispt:


pos_pywebdriver/models/ir_module_module.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def write(self, vals):
+        result = super().write(vals)
+        if "state" in vals:
+            conflicting = (
+                self.env["ir.module.module"]
+                .sudo()
+                .search_count(
+                    [
+                        ("name", "in", ["pos_iot", "pos_pywebdriver"]),
```
legalsylvain: naive question : why not use excludes key in manifest file.
something like `'excludes' = ['pos_iot']` ?

legalsylvain (commented):
Thanks for providing this module !
 
I faced similar issue some monthes ago on v18, and loose time before understanding `pywebdriver` and `pos_iot` was incompatible... A lot of ugly code in Odoo on that topic ! 

FYI, a new project exists that provide similar features as pywebdriver. it is used in GRAP & Coop It easy for the time being. _Maybe_ a more generic name could be great. (`pos_without_iot`  / `pos_community_driver` ... not a blocking point, though).

You can take a look here : 
https://gitlab.com/grap-rhone-alpes/odoo-driver

(In a few word : it replace pywebdriver in a more modern way, with plugin, etc. BUT, there is no installer for windows).

dreispt:
Thank you @legalsylvain .
This started as a `pos_escpos_status` replacement, but because it was an Enterprise installation I quickly went into the `pos_iot` rabbit hole.

The point for `pos_escpos_status`was to support `pywebdriver`.
And since this is completely new code, I took the opportunity to make the module name clearly express that.

I don't know if your `odoo-driver`is a replacement for `pywebdriver`. A quick look feels it has less features.
If it is, naming this module as `pos_odoodriver` isn't very informative either.

In the end of the day, I wish we had a project like `pywebdriver` or `odoo-driver` under the OCA, and that teh project has a title that is more meaningful about what it is about...

pos_pywebdriver/models/ir_module_module.py
dreispt: You're right, I'll make that change.

legalsylvain:
Thanks for your answer.

> A quick look feels it has less features.

Interesting. Could you elaborate ? What are the features that are missing for you ?

> In the end of the day, I wish we had a project like pywebdriver or odoo-driver under the OCA.

I thought the same thing at first, when I created pywebdriver.
But there's a problem with the OCA: its ability to maintain a "product" vision. 
When I saw the reactions to this issue (https://github.com/pywebdriver/pywebdriver/issues/57) for exemple, I thought it was best for the project to remain outside the OCA and under our control.

pos_pywebdriver/models/pos_config.py
```diff
@@ -0,0 +1,52 @@
...
+
+class PosConfig(models.Model):
+    _inherit = "pos.config"
+
+    use_pywebdriver = fields.Boolean(
+        string="Use PyWebDriver",
+        help="Connect receipt printer via a PyWebDriver proxy instead of IoT Box.",
+        default=False,
+    )
+    pywebdriver_ip = fields.Char(
```
legalsylvain, line 16: Sorry. I dont understand why not use native field. Proxy_ip

https://github.com/odoo/odoo/blob/19.0/addons/point_of_sale/models/pos_config.py#L125

dreispt:
@legalsylvain 
If I recall correctly, it seems Linux centric and missing Windows support, like Pywebdriver has.

We do have the maintainer role under the OCA, that can own a roadmap.
It is just that it is rarely owned, as people dedicate time to whatever their current project required, and there isn't a permanent commitment with the project (that's what happens with me at least).

I think this ends up being OK as long as there are multiple maintainers, that make the project move forward in bursts, at different point in time.
Otherwise you risk getting an unmaintained project, frustrating contributors. That is the case currently for Pywebdriver.

So moving it into the OCA could allow for different people in different moment make it move forward.

legalsylvain:
@dreispt : I propose to move this discussion here https://github.com/pywebdriver/pywebdriver/issues/57

thanks !

PR #1573, opened 2026-06-22
[18.0][MIG] pos_payment_change: Migration to 18.0

victoralmau:
Migration to 18.0

Superseed https://github.com/OCA/pos/pull/1453

Changes done:
- [x] Invoiced order domain compatibility (https://github.com/OCA/pos/pull/1453#pullrequestreview-4417087082).
- [x] Add Payment Date column to wizard
- [x] Show the “Change Payments” button only when the status is “invoiced” (if payment_change_policy=update).
- [x] Improve tests

@Tecnativa TT62460


pedrobaeza:
/ocabot migration pos_payment_change

If the main goal of the module is already covered, shouldn't this be renamed or to start a new one with the missing feature?

victoralmau:
> If the main goal of the module is already covered, shouldn't this be renamed or to start a new one with the missing feature?

I actually think the module's name is appropriate because its purpose is to modify pos payments when Odoo Core doesn't allow it.

pedrobaeza:
Yes, but it's too generic. Now that feature is covered by core, except the case where the order has been invoiced, so the module name should reflect it.

victoralmau:
> Yes, but it's too generic. Now that feature is covered by core, except the case where the order has been invoiced, so the module name should reflect it.

It is important to remember that this module also includes the “Refund and Resale” option (at the `pos.config` level), which is allowed when the order is in the ‘paid’ or “invoiced” status.

pedrobaeza:
Yes, and that's why I'm saying that the module is now other thing. I would say `pos_after_invoice_mgmt` for example.

victoralmau:
> Yes, and that's why I'm saying that the module is now other thing. I would say `pos_after_invoice_mgmt` for example.

Okay, so I'll rename the module to `pos_after_invoice_mgmt`, and we'll keep all the functionality there? (including the “Refund and Resale” functionality for the “paid” status)

pedrobaeza:
Let's see what @legalsylvain thinks. For us, AFAIK, we only need the other part.

PR #1576, opened 2026-06-27
[19.0][MIG] pos_sale_picking_keep

alan196:
Migration of `pos_sale_picking_keep` from 18.0 to 19.0.

### Migration changes
- Bump manifest version to `19.0.1.0.0`.
- Update the PoS test tour imports to the new 19.0 paths (`tests/pos/tours/utils/*` and `tests/generic_helpers/dialog_util`).
- Grant `sales_team.group_sale_salesman` in the test setup: in 19.0 `point_of_sale.group_pos_manager` no longer implies sales rights and the web `onchange` now checks `create` access, so creating the sale order in the test raised `AccessError`.
- Use f-strings for the tour URLs (`ruff` UP031).
- Add `readme/CONTRIBUTORS.md` and regenerate the README.

Functional behavior is unchanged; both PoS tours pass locally.

cc @pedrobaeza


alan196:
ℹ️ Tests pass (`tour SUCCEEDED`, `0 failed, 0 error(s)`). CI is red only because of the Chrome teardown WARNING (`Killing chrome descendants-or-self ... chrome (zombie)`) that `checklog-odoo` flags, which affects every PR with a tour in this repo. The infra fix is in #1577; once merged, this PR will go green when re-run on the updated base.

alan196:
Added a second commit with an improvement found while using the module in production: when settling a sale order that was already delivered, the PoS loaded the lines with zero quantity (or even dropped the lines completely for products whose UoM is not groupable in the PoS, due to the line splitting logic of `pos_sale`). Since with this module the PoS is only used to charge the order and the sale order pickings are kept, the ordered quantity is now loaded, deducting only what was already invoiced. Covered with a new tour-based test.

PR #1579, opened 2026-07-08
[19.0][MIG] pos_product_multi_barcode: Migration to 19.0

arielbarreiros96:
Standard migration to v19

Depends on:

- [ ] [stock-logistics-barcode #723](https://github.com/OCA/stock-logistics-barcode/pull/723)


DaanMarchal:
@arielbarreiros96 Thanks for this migration, we're running it on a 19.0 project.

One thing I've added on our side: a tour test that scans a product's main barcode and then its additional ones, asserting they all land on the same order line. The existing tests only cover barcodes_json server-side, so nothing currently demonstrates the module's actual purpose, and with the JS patch removed the tour fails at the second scan, so it does catch regressions.

Happy to push it to your branch, or to open a follow-up PR once this merges, whichever you prefer.

Note this is blocked on stock-logistics-barcode#723, which we've also offered to take over.

arielbarreiros96:
> @arielbarreiros96 Thanks for this migration, we're running it on a 19.0 project.
> 
> One thing I've added on our side: a tour test that scans a product's main barcode and then its additional ones, asserting they all land on the same order line. The existing tests only cover barcodes_json server-side, so nothing currently demonstrates the module's actual purpose, and with the JS patch removed the tour fails at the second scan, so it does catch regressions.
> 
> Happy to push it to your branch, or to open a follow-up PR once this merges, whichever you prefer.
> 
> Note this is blocked on stock-logistics-barcode#723, which we've also offered to take over.

Hello @DaanMarchal, I think the extra tests can be added as an extra [IMP] PR, since to allow the migration itself to be independent a little more independent. In relation to [#723](https://github.com/OCA/stock-logistics-barcode/pull/723) it seems basically approved, and waiting just for merge

PR #1583, opened 2026-07-13
[18.0][IMP] pos_sale_picking_keep: Keep sale pickings and/or POS pickings too

carlos-lopez-tecnativa:
Remove the configuration and related files. With this module installed, the expected behavior should apply by default, without any special configuration. Therefore, we removed these settings.
 
Supersedes https://github.com/OCA/pos/pull/1545
TT63602
@Tecnativa @victoralmau @pedrobaeza @rousseldenis could you please review this?


rousseldenis:
IMHO, the behaviour change introduced by this module should be configurable.

pedrobaeza:
Adding extra steps for me is an entry barrier that is not adding value in this case. If you ever have a case where you want different behaviors across PoS configs or companies, then you can complicate it, but is it a current real case?

pedrobaeza:
@carlos-lopez-tecnativa can you squash a bit the commit history and include TT?

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa can you squash a bit the commit history and include TT?

Done

pedrobaeza (approved):

victoralmau (approved):

PR #1586, opened 2026-07-17
[17.0][ADD] pos_customer_display_monitor: New module

BernatObrador:
This module allows Point of Sale users to select the monitor where the
local customer display should be opened.

The module will:

- Detect the screens currently connected to the device.
- Find the screen that best matches the saved selection.
- Open the Odoo customer display on that screen.
- Move and resize the window to occupy the available screen area.

If the previously selected monitor cannot be found, the module falls
back to a secondary monitor when available, or to the primary monitor.

This module only works with chromium based browsers that have access to getScreenDetails()

cc https://github.com/APSL 51344

@miquelalzanillas @miquelpascual @ppyczko @cubells please review


miquelpascual (approved):
LGTM, code review

ppyczko (approved):
LGTM, tested in runboat!

cubells (approved):

PR #1591, opened 2026-07-31
[IMP] pos_access_right: Add permission to delete single order line

ivilata:
Its operation and defaults are very similar to the "Negative Quantity" right: when the user does not belong to the "Delete Order Line" group, the backspace button in the PoS' numeric pad is grayed out and disabled.

Just as with the rest of the module, keyboard access still works, though.


ivilata:
@legalsylvain: Does this new feature look worth merging to you? Thanks! :slightly_smiling_face:

ivilata:
Hi again @legalsylvain (or anyone in @OCA/pos-maintainers interested in this PR). I was wondering if it may make more sense to create a separate module to add this "delete order line" permission, instead of modifying the `pos_access_right` module with this PR. I still think that it may add more overhead than just applying these changes though, but I may be wrong.

Thanks in advance for any comments/reviews! :slightly_smiling_face:

legalsylvain (approved):
code review. LGTM . 

/ocabot merge minor

ivilata:
Happy to hear that, thanks @legalsylvain! :smile:

PR #1598, opened 2026-08-18
[18.0][ADD] pos_order_line_change_variant: change the variant of an existin…

cvinh:
…g order line

In standard Point of Sale, once a product with variants has been added to the cart, there is no way to switch it to a different variant: the line has to be removed and the new variant re-added as a separate line. Clicking a cart line only lets you edit quantity, discount or price through the numpad; double-clicking a line currently does nothing beyond cancelling the line's pending auto-deselection timeout (`OrderSummary.clickLine`, `ev.detail === 2` branch), so this event was free to repurpose.

This module lets the user double-click a cart line whose product is configurable (`product.isConfigurable()`) to reopen the variant configurator, pre-filled with the line's current attribute values, and apply the newly picked variant on that same line instead of creating a new one.

`PosStore.changeOrderlineVariant(line)` reuses the same variant resolution logic `addLineToOrder` already uses when adding a configurable product (matching `product_template_variant_value_ids` against the confirmed `attribute_value_ids`), then updates the line's `product_id`, `attribute_value_ids`, `custom_attribute_value_ids`, `price_extra`, unit price and full product name in place.

The change is refused, with an explanatory dialog, when:
- the line already has lot/serial numbers assigned, since changing the product could invalidate them;
- another line in the order already uses the target variant, to avoid having to silently merge two lines.

A Python HttpCase + tour test covers the full flow: add a variant, add it to the cart, double-click the line, switch to another variant, check the line was updated in place rather than duplicated.

Assisted-by: Claude (Anthropic)


cvinh:
tests fail because of pos_user_restriction

PR #1611, opened 2026-09-18
[18.0][ADD] pos_restriction_per_user: New module

victoralmau:
New module

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT64552


pos_restriction_per_user/security/ir_rule.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="pos_config_user_rule" model="ir.rule">
+        <field name="name">POS configs for user</field>
+        <field name="model_id" ref="point_of_sale.model_pos_config" />
+        <field name="domain_force">[('restrict_user_ids', '=', user.id)]</field>
```
pedrobaeza, line 6: For showing PoS with no users configured:

```suggestion
        <field name="domain_force">[('restrict_user_ids', 'in', [user.id, False))]</field>
```

and the same for the rest.

pedrobaeza (commented):
And I think this will put CI green.

carlos-lopez-tecnativa:
Isn't this feature already covered by the [pos_user_restriction](https://github.com/OCA/pos/tree/18.0/pos_user_restriction) module? Or why do you need a new module?

pedrobaeza:
Ouch, I didn't know that one... then I think indeed it can serve.

carlos-lopez-tecnativa:
The main difference I noticed is that the other module creates a new group below the POS users.

Additionally, I noticed that the ir.rules mix the intercompany rules with the business rules. However, if you think this module can be reused, a few adaptations could be made, or it could be used as it is.

pedrobaeza:
Well, the extra sublevel is not a problem IMO: you assign that one when you want to restrict and that's it. If there's a technical problem, that's indeed should be fixed in the module itself.

victoralmau:
We will use `pos_user_restriction` and modify the `ir.rule` settings to remove the company-related information, which is already included in `point_of_sale`: https://github.com/odoo/odoo/blob/bba17c823b4509616126dc9d3cd4f24a3aee5474/addons/point_of_sale/security/point_of_sale_security.xml#L57

PR #1612, opened 2026-09-21
[18.0][IMP] pos_user_restriction: Remove the company_id condition from the rules' domain

victoralmau:
Remove the `company_id` condition from the rules' domain

The `point_of_sale` module already adds the company rules: https://github.com/odoo/odoo/blob/bba17c823b4509616126dc9d3cd4f24a3aee5474/addons/point_of_sale/security/point_of_sale_security.xml#L57

Related to: https://github.com/OCA/pos/pull/1611#issuecomment-5732779574

@Tecnativa TT64552


pedrobaeza (approved):

pos_user_restriction/security/pos_security.xml
```diff
@@ -21,7 +21,7 @@
             <field ref="point_of_sale.model_pos_config" name="model_id" />
             <field
                 name="domain_force"
-            >['|',('assigned_user_ids','in',user.id),('assigned_user_ids','=',False),('company_id', 'in', company_ids)]</field>
+            >['|',('assigned_user_ids','in',user.id),('assigned_user_ids','=',False)]</field>
```
carlos-lopez-tecnativa:
```suggestion
            >[('assigned_user_ids','in',[user.id, False])]</field>
```

carlos-lopez-tecnativa (commented):
I think this needs a migration script so that the changes are taken into account in existing databases where the module is already installed. Is that correct?

Remember, this code is inside a `noupdate` block.

pos_user_restriction/security/pos_security.xml
victoralmau: That domain doesn't work (sorry for not clarifying that earlier—I had already tried it before); it seems that it doesn't work properly with m2m fields.

victoralmau:
> I think this needs a migration script so that the changes are taken into account in existing databases where the module is already installed. Is that correct?
> 
> Remember, this code is inside a `noupdate` block.

IMO, it's incorrect for the rules to be defined with `noupdate=1`; in other words, I don't see any reason why they should be (I see that they were already defined that way in **12.0**), so I think the right thing to do would be to leave them without `noupdate=1`.

carlos-lopez-tecnativa:
> > I think this needs a migration script so that the changes are taken into account in existing databases where the module is already installed. Is that correct?
> > Remember, this code is inside a `noupdate` block.
> 
> IMO, it's incorrect for the rules to be defined with `noupdate=1`; in other words, I don't see any reason why they should be (I see that they were already defined that way in **12.0**), so I think the right thing to do would be to leave them without `noupdate=1`.

If you plan to remove `noupdate="1"` or set it to `noupdate="0"`, I think (I haven't tested this) you would still need a migration script to update the value in `ir.model.data`. You can use the [set_xml_ids_noupdate_value](https://github.com/OCA/openupgradelib/blob/59170a8fbfd69f5d5e5204c02a10667d76e15807/openupgradelib/openupgrade.py#L3735) function from `openupgradelib`.

Odoo added `noupdate="1"` in some places, mainly for `res.groups` or module data, to prevent future module updates from overwriting changes made through the UI. For security rules, this makes sense in some cases, but not necessarily in others. It all depends on whether it makes sense for users to be able to edit the record and have those changes preserved across future module updates.

pos_user_restriction/security/pos_security.xml
```diff
@@ -14,77 +14,60 @@
...
-            <field name="groups" eval="[(4, ref('point_of_sale.group_pos_user'))]" />
-        </record>
-    </data>
+    <record id="pos_config_assigned_users" model="ir.rule">
+        <field name="name">Assigned points of sale</field>
+        <field ref="point_of_sale.model_pos_config" name="model_id" />
+        <field
+            name="domain_force"
+        >['|',('assigned_user_ids','in',user.id),('assigned_user_ids','=',False)]</field>
+        <field name="groups" eval="[(4, ref('group_assigned_points_of_sale_user'))]" />
```
carlos-lopez-tecnativa: I noticed you updated the old syntax to use `Command` in the test. If you could update it the same way in the XML data, that would be great.
```suggestion
        <field name="groups" eval="[Command.link(ref('group_assigned_points_of_sale_user'))]" />
```

pos_user_restriction/migrations/18.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+]
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    # Workaround to execute the migration script without errors
+    # see https://github.com/odoo/odoo/blob/2a839ef1ed09c36f27ce7536ca3052d9f65ceed9/odoo/modules/migration.py#L252-L256
+    env = cr
+    openupgrade.set_xml_ids_noupdate_value(
+        env, "pos_user_restriction", _noupdate_xmlids, True
```
carlos-lopez-tecnativa: I tested this locally, and it’s not working. The rules are not being updated. Please set this to `False` so that the rules are updated correctly:
```suggestion
        env, "pos_user_restriction", _noupdate_xmlids, False
```

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge nobump

